Issue #3, opened 2024-01-02
Convert colors to Tailwind


Shpigford:
We need to convert the color palette to Tailwind color syntax (for use in tailwind.config.js).

Here's the Figma file with the colors: https://www.figma.com/file/rAn2k5U86nXZmBixAEsjdN/Shell?type=design&node-id=8%3A3495&mode=design&t=0q11JqWuYDnEcFDH-1

Tailwind color customization docs: https://tailwindcss.com/docs/customizing-colors

Tailwind config file: https://github.com/maybe-finance/maybe/blob/main/config/tailwind.config.js

<img width="2332" alt="Colors" src="https://github.com/maybe-finance/maybe/assets/35243/077148ba-2adb-413a-8cd3-e86afd5ae3b3">

faqndo97:
@Shpigford The list of colors from tailwind should be overwritten and reduced to only those colors? That means no other colors will be used, am I right?

Shpigford:
> @Shpigford The list of colors from tailwind should be overwritten and reduced to only those colors? That means no other colors will be used, am I right?

That's correct. This is the exclusive color palette.

PR #4, opened 2024-01-02
Add Maybe colorscheme to tailwind.config.js

jespr:
@Shpigford is this what you had in mind for #3?

I've kept the naming from the graphic you added, so we're using `primary-50` instead of `cyan-50`, `primary-300` instead of `cyan-300` and so on - but the others are still named using the color names: blue, orange, pink, grape, indigo and green.

I've also nulled out the standard tailwindcss colors - but we can change that of course.

### DIFF
--- a/config/tailwind.config.js
+++ b/config/tailwind.config.js
@@ -23,25 +23,25 @@ module.exports = {
         700: '#232428',
         800: '#1c1c20',
       },
-      primary: {
+      cyan: {
         50: '#d7f6fa',
         300: '#d7f6fa',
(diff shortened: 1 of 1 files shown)


jespr:
I just noticed that Primary Cyan 50 and Primary Cyan 300 has the same hex code in the graphic?

jespr:
@Shpigford changes pushed

Shpigford (approved):
Perfect! Thank you so much!

Shpigford:
Resolves #3

Issue #5, opened 2024-01-02
Build sidebar user menu


Shpigford:
![CleanShot 2024-01-02 at 13 40 33@2x](https://github.com/maybe-finance/maybe/assets/35243/d6b7bfb2-0bf3-4fe2-b965-6f59d6676a89)

In the sidebar we show the email address of the user. We need to add a basic menu pop-up.

Scope:

1. Add stacked "dots" menu icon
2. Use Stimulus to create the pop-up
3. "Settings" can use a blank/empty path for now
4. Log out link should point to `destroy_user_session_path`

Figma file: https://www.figma.com/file/rAn2k5U86nXZmBixAEsjdN/Shell?type=design&node-id=0%3A1&mode=design&t=0q11JqWuYDnEcFDH-1

Use Heroicon SVG files: https://heroicons.com

PR #8, opened 2024-01-03
Build sidebar user menu

claudey:
- Added a toggle-controller to handle the toggle action
- Using stimulus to create the pop-up
- Icons are from heroicons

This PR is for #5 

https://github.com/maybe-finance/maybe/assets/1866816/aa48764e-2cdb-421c-a6df-0795340910a1


app/views/layouts/application.html.erb
Shpigford, line 1: We should be using the raw SVG code instead of SVG images as it gives us more flexibility on things like icon color.

Shpigford: I'd suggest removing the SVG files and just using the SVG code directly in its place.

Otherwise, looks good!

Shpigford (changes requested):

app/views/layouts/application.html.erb
claudey: Updated in https://github.com/maybe-finance/maybe/pull/8/commits/a4b9a0507d06b013aba66bbc56be98b887f518af

Issue #21, opened 2024-01-10
Remove "Advisor' functionality


Shpigford:
Part of Maybe's original product offering included being able to work directly with a CFP/CFA. We built an entire UI for the Advisor side as well as a whole section of the user-facing app for asking questions and conversing with your advisor.

That's not necessary at this point, so we can remove those elements.

Things that likely need to be removed...

1. `/apps/advisor` — This is the advisor admin panel
2. Database tables: advisor, conversation, conversation_advisor, conversation_note, message
3. Configuration elements (such as ones that launch the advisor service)

tmyracle:
Working on this, should have a PR up soon. Lots of stuff to remove 😅

PR #25, opened 2024-01-11
Remove all advisor stuff and fix migrations

tmyracle:
Addresses #21 
- Removes all advisor functionality
- Removed relevant models and enums from database:
  - Advisor
  - Conversation
  - ConversationAdvisor
  - ConversationNote
  - MessageType
  - Message
  - ConversationStatus
  - ata fields from notifications on User

The notifications preferences UI will likely need an update (or total removal) later since the only thing left after removing the advisor notifications is a ConverKit function or 2. Will revisit once we get auth sorted and can log in and really test the UI.

I also tried to clean up some unused imports throughout files I touched, hence some of the deletions there.
 
This change also uses introspection to populate the dbgenerated parameters for columns that are a custom type (non native data type). There's a [known issue](https://github.com/prisma/prisma/issues/15654) in Prisma where dbgenerated() with nothing passed in will continue to generate migrations that are not compatible when it should noop. This should fix the migration hiccups discussed in the Discord.

I don't love having those big database functions in the schema.prisma file but seems to be the best workaround to get migrations working smoothly again. Might revisit later and refactor using triggers. 

I also bumped prisma dependency to see if they had patched the issue, but no luck. 

Lots of stuff going on here to remove all the advisor functionality so apologies for the massive PR.


Shpigford:
Thanks @tmyracle! Definitely would prefer smaller PRs in the future, but fine at the moment given we're doing some major cleanup and trying to reduce the overall complexity of the app.

Have merged to main.

PR #26, opened 2024-01-11
Remove Intercom

tamirfyi:
Removed all hooks/references, and replaced any interactive buttons with regular text.


Shpigford:
Thanks @tamirfyi. Ultimately decided to go with #27. Your PR had a couple of bits that were outside the scope of the Issue (DB migration + removal of `useAccounts`).

PR #27, opened 2024-01-11
feat: remove Intercom usage

### LINKED ISSUE #22: Remove Intercom
As you won't need Intercom to support yourself, we don't need those bits in the app.

MichaelDeBoey:
Closes #22

---

I already started working on this before I saw @tamirfyi's #26 🙈

Still noticed some differences:
- No formatting changes
- Also removed unused imports
- No removal of the `useAccounts` hook from `libs/client/shared/src/api/useAccountApi.ts` (its usage wasn't removed from the rest of the codebase either in #26)
- No `yarn.lock` changes (should be done in a separate PR imo)
- No Prisma migration (should be done in a separate PR imo)


Shpigford:
@MichaelDeBoey Let's go with yours (primarily due to the `useAccounts` bit).

tamirfyi:
I tried getting this one right before bed since it seemed easier than others, but I guess I was tired and overlooked some things. nice work michael!

PR #34, opened 2024-01-11
refactor: remove dead code

MichaelDeBoey:


Shpigford:
@MichaelDeBoey Can you provide some more insight on the changes?

MichaelDeBoey:
@Shpigford My IDE pointed out that these types and the `useNotificationsApi` hook were never used
I confirmed this manually, hence they can all just be removed from the codebase imo

Issue #17, opened 2024-01-10
Remove AWS dependencies


Shpigford:
There are slew of AWS-specific bits...so many that I honestly don't recall what all they were used for. But ultimately we need to make this platform-agnostic so that AWS isn't a requirement.

tamirfyi:
First attempt at an OSS contribution - I'm going to work on this and see if I can get a PR out tonight.

guptadeepak8:
Can i get assigned?

Shpigford:
Hey guys, we don't pre-assign issues at the moment without a longer history of commits to the repo. Thanks for the interest!

ssong:
What's the desired first outcome here?

Is the intention to strip out the existing AWS setup and provide guides for self-hosting or 'get it working locally' type situation?

Shpigford:
@ssong for this specific issue, it's just "remove any code associated with deploying to aws". we'll put together a separate issue about local deployment once we get the codebase functioning properly.

PR #40, opened 2024-01-13
feat: remove AWS dependencies

larswmh:
## Description
As requested in #17, here's a PR to remove anything from the repo that's associated with deploying to AWS.

/claim #17


Shpigford:
On the surface, looks good!

PR #46, opened 2024-01-14
fix: made uncaughtException arrow function like others

parth0105pluang:
Made uncaughtException handler as arrow function like that of unhandledRejection and other callback functions being used in the file.


parth0105pluang:
@Shpigford

amfrshawon:
fix: made uncaughtException arrow function like others

PR #48, opened 2024-01-14
changed docker-compose to docker compose in package.json

kartikay-agarwal:
in line 7, in package.json changed docker-compose to docker compose as docker-compose has been deprecated.
https://docs.docker.com/compose/migrate/


MichaelDeBoey:
Superseded by #134

PR #49, opened 2024-01-14
Update README.md for npm

PlotMasterKoushik:


coltonehrman:
I don't think this project uses `npm`. It uses `yarn`, from what I have seen.

Shpigford:
@coltonehrman is correct. sticking with yarn for now.

PR #62, opened 2024-01-14
#fix 55 Centralize Postmark env variables to .env file

### LINKED ISSUE #55: Centralize Postmark env variables to .env file
Postmark is already used pretty extensively throughout the app for email sending, however the environment variables for it are sprinkled all over the app.

We need to centralize all of those in to the main .env file and include them as part of setup.

We should be able to continue using it as the main email provider as they have a free plan (for up to 100 emails).

sy425191:
fixes #55

Centralize Postmark env variables to .env file


Shpigford:
Went with #59 as this particular PR had a bit to many additional changes.

PR #69, opened 2024-01-15
Remove all feedback related components

### LINKED ISSUE #65: Remove Feedback buttons
There are a couple of components that can be removed as we don't need "native" feedback buttons at this stage.

I _think_ it's only `FeedbackDialog` and `FeedbackButton` (and any mention of them), but would love someone who's more versed in React to dig through it and make sure there's nothing else obvious.

This includes removing `NEXT_PUBLIC_ZAPIER_FEEDBACK_HOOK_URL` from the .env.example file as it will no longer be needed either.

Greek:
Fixes #65. There shouldn't be any remaining components or other files referencing to these deleted components.


Shpigford:
Looks good! Thanks!

Issue #53, opened 2024-01-14
Remove Auth0 references


Shpigford:
Given we've now replaced Auth0, we should remove any references to it or unnecessary code.

The biggest thing there is likely the `/auth0` directory.

Shpigford:
Fixed with #70

PR #70, opened 2024-01-15
Remove (some) auth0.

anthonyshew:
👋  First-time in the repo!

This PR addresses #53. It may not include all of the required work to close out the issue - but should help to get part of the way there!

Note: I was having trouble getting tests to run right (and didn't know if they were expected to not run right given the history of the repo and current state). Accordingly, I didn't touch the e2e tests - but can if maintainers would like that done!


apps/server/src/app/__tests__/utils/user.ts
```diff
@@ -3,12 +3,12 @@ import prisma from '../../lib/prisma'
 
 const EMAIL = 'test@example.com'
 
-export async function resetUser(auth0Id = '__TEST_USER_ID__'): Promise<User> {
+export async function resetUser(authId = '__TEST_USER_ID__'): Promise<User> {
     const [_, [user]] = await prisma.$transaction([
-        prisma.$executeRaw`DELETE FROM "user" WHERE auth0_id=${auth0Id}`,
```
coltonehrman, line 8: Not sure changing in this way is correct? We likely want to rely on userId instead of any "authId" moving forward?

Shpigford: I believe @tmyracle can speak to if this is correct or not.

tmyracle: Yes this is correct. the AuthUser object handles the base information used by NextAuth and is associated with the pre-existing User object via with auth_id field, similar to how it was handled with Auth0.

Shpigford:
I think this solves #53, or at least the majority of it enough to call it solved for cleanup purposes. Thanks!

PR #71, opened 2024-01-15
GitHub Action quality checks.

anthonyshew:
- Setup Turborepo.
- GitHub action quality checks.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -1,0 +1,35 @@
+name: Quality
+
+on:
+    push:
+        branches: ['main']
+    pull_request:
+        types: [opened, synchronize]
+
+jobs:
+    quality:
+        name: Quality checks
+        timeout-minutes: 15
+        runs-on: ubuntu-latest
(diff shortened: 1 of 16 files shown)


tmyracle:
since this monorepo already uses NX is there a benefit to adding turbo as another tool? i believe there's already some pre-commit checks around linting/formatting, is it possible to achieve the same output without introducing a new package. love vercel btw, just trying to avoid more dependencies if we don't absolutely need them.

anthonyshew:
Agreed! We can reduce a ton of dependency and code bloat, ending up with a much simpler Turborepo. This was largely a demonstrative PR to explore migrating incrementally to hear from maintainers!

anthonyshew:
Realizing I should also note: I also did this keeping eventual self-hosters, self-deployers, etc. in mind since Vercel appears to be an intended first-class deploy target. I've seen rumblings to go for the cal.com strategy + Vercel-specific code in the repo.

We work really hard at making Turborepo (and Turbopack soon) a zero-config experience so hoping we could save tons of folks time and effort with a move. Lining up with the rest of the Vercel ecosystem would likely play out well long term.

PR #73, opened 2024-01-15
Country limitation while signup is removed

### LINKED ISSUE #64: Remove location limitation
As referenced in #60, we need to remove the blocker for folks outside the US.

> When we built Maybe we offered financial advice and as such had a LOT more regulatory stuff to account for. So, that just meant we couldn't support non-US folks.
> ...
> The problem will still be that we have to add data sources for folks outside North America, but otherwise you should still be able to actually create an account and manually add stuff.

There are some checks during signup/onboarding where it asks for location and then if you don't select United States it shows you a message about not being supported.

We need to remove that so there's no "location" limitation.

sy425191:
/fix #64 

This remove the checks added while signup, that only US citizens can use this app, Also removed the dead files i.e. **CountryWaitlist**


Neumand (changes requested):
A few spots were missed, actually:

- Line 214: `country !== 'US')`
- Line 250: `{country === 'US' && (`

sy425191:
Thanks for pointing it out @Neumand 🙏

Shpigford:
Agreed re: states. Looks good. Merging.

PR #82, opened 2024-01-15
feat: add basis api documentation

marclave:
hey y'all 👋 this is very much a work in progress, I noticed the [Wiki](https://github.com/maybe-finance/maybe/wiki/REST-API) had the REST API Documentation. But I figured we could some docs in code with [Swagger-JSDOC](https://github.com/Surnet/swagger-jsdoc/tree/v7) and [Scalar](https://github.com/scalar/scalar) ✨ 

we had a lot of fun making a Maybe theme

image

you can test it out here https://maybe.apidocumentation.com/

I made a few tags with a few endpoints, but the idea is we can add a little bit of documentation per endpoint and then get amazing API Docs that can be maintained in a simpler way ✨ 

I'm happy to dive in and finish the docs but I wanted to get some feedback first :)


Shpigford:
@marclave I appreciate the initiative on this! Right now it feels a tad premature as we're still sorting out what's even staying in the app. So it ultimately adds some overhead that I think we're not quite ready to handle just yet.

Certainly _looks_ really cool, though!

PR #89, opened 2024-01-15
Update broken bank/institution image location

### LINKED ISSUE #61: Update broken bank/institution image location
Right after successful registration the First Account page is missing bank icons:
image

Because they point to dead URLs.

Do you still have them? I can add them.

giuseppecutuli:
/fix #61

I added the missing asset files to the repository. I avoided using Image from nextjs because it didn't make sense since the files are contained in the static folder.
I avoided using require or dynamic imports because they are not recommended by nextjs.


Shpigford:
Be sure to mention the issue # you're working!

PR #90, opened 2024-01-15
docs: Add common issue section

coltonehrman:
- Added **COMPOSE_PROFILES** issue
- Added **ngrok** issue


coltonehrman:
Should update project to use something like this - https://www.npmjs.com/package/cross-env to set env variables inside scripts.

README.md
```diff
@@ -61,6 +61,45 @@ yarn prisma:seed
...
+error: server: unhandled promise rejection: [object Promise]: Error: connect ECONNREFUSED ::1:4551
+error: server: unhandled promise rejection: [object Promise]: Error: connect ECONNREFUSED ::1:4551
+error: server: unhandled promise rejection: [object Promise]: Error: connect ECONNREFUSED ::1:4551
+```
+
+This error will show on startup after running `yarn dev`.
+
+What this means is the local **ngrok** service is not running.
+
+To start up **ngrok** locally, you need to run `COMPOSE_PROFILES=ngrok docker-compose up -d`.
```
tmyracle, line 79: FYI: alternatively, you can run yarn run dev:services:all and we can remove stripe from that command in package.json for now.

coltonehrman: Right, I just mentioned this as the workaround now. Once we modify the scripts we can update this

tmyracle: updating the script in this PR now: [https://github.com/maybe-finance/maybe/pull/93](https://github.com/maybe-finance/maybe/pull/93) 👍

coltonehrman: Wondering if we should update the setup commands now?

tmyracle: I think this issue is now resolved with the latest changes to the commands in package.json

jbolda: Aye, docker compose should be set. Also since we already have NX, we can embed the env var requirements in that config. It will also be more flexible.

PR #93, opened 2024-01-15
Fix #92: Noop postmark sends if API key is not defined

### LINKED ISSUE #92: Suppress issues from missing NX_POSTMARK_API_TOKEN
When `NX_POSTMARK_API_TOKEN` is empty, an error is thrown:

```
error: uncaughtException: A valid API token must be provided.
PostmarkError: A valid API token must be provided.
```

We shouldn't _require_ `NX_POSTMARK_API_TOKEN` and instead should just ignore any email sending when that token is missing.

tmyracle:
Addresses #92 and updates yarn run dev:services:all to get rid of stripe for now and update README to use that command so that ngrok is started at the same time as postgres and redis

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "scripts": {
         "dev": "nx run-many --target serve --projects=client,server,workers --parallel --host 0.0.0.0 --nx-bail=true --maxParallel=100",
         "dev:services": "COMPOSE_PROFILES=services docker-compose up -d",
-        "dev:services:all": "COMPOSE_PROFILES=services,ngrok docker-compose up",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
     "scripts": {
         "dev": "nx run-many --target serve --projects=client,server,workers --parallel --host 0.0.0.0 --nx-bail=true --maxParallel=100",
         "dev:services": "COMPOSE_PROFILES=services docker-compose up -d",
-        "dev:services:all": "COMPOSE_PROFILES=services,ngrok,stripe docker-compose up",
```
coltonehrman, line 8: Do we want to add `-d` flag?

tmyracle: good catch, added

Issue #86, opened 2024-01-15
Migrate from Webpack to Vite


Neumand:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Update NX to use `vite` instead of `webpack`

**Describe alternatives you've considered**
I guess an alternative would be `esbuild`? I have less direct experience with other bundlers. Open to suggestions.

**Additional context**
Migrating here should improve the build time considerably. We made the shift at work and it saved us tons of time. This could be a good early improvement.

Shpigford:
Can you provide a high-level overview of what may be required here? Basically trying to gauge effort + impact.

Neumand:
Sure.

I _think_ this is the extent of the changes. If someone else has experience with this, I would love their opinion.

- Update `workspace.json` to use `@nx/vite` as the executor instead of `@nrwl/webpack:webpack`
- Upgrade Storybook to use `vite`
- Remove packages that leverage `webpack` (i.e. `@svgr/webpack`). It looks like most of them are not used.

I'd be happy to experiment with this myself to dig deeper into the complexity.

Shpigford:
Yeah, I'd suggest doing some work locally on it to quantify any potential improvements, then you can open a PR with that and have other folks try out the solution as well.

PR #99, opened 2024-01-16
Improvement: Webpack -> Vite

Neumand:
References #86 

Could be good to get someone to help review these changes. Running into CORS issues at the moment.


workspace.json
```diff
@@ -287,11 +281,9 @@
                     }
                 },
                 "serve": {
-                    "executor": "@nrwl/node:node",
+                    "executor": "@nrwl/vite:dev-server",
```
tmyracle, line 284: I'm not sure if vite:dev-server the right executor for the backend server? I thought vite was primarily for frontend apps, but I could be wrong. I've only used it as a replacement for create-react-app

Neumand: Ah, I think you're right.

I'll check if we can leverage another executor instead. Might be why I'm having issues.

Neumand:
I don't believe there's an executor besides `webpack` for Node at the moment.

Closing.

PR #110, opened 2024-01-16
feat(real estate): make real estate country custom

### LINKED ISSUE #109: Add global support to real estate addition
CleanShot 2024-01-16 at 09 41 22@2x

Currently the "Add real estate" modal is locked to the US, but there's no functional reason to keep it this way.

We should allow any country via a dropdown listing of all countries.

pablo-kitz:
Fixes #109

### DIFF
--- a/libs/client/features/src/accounts-manager/property/PropertyForm.tsx
+++ b/libs/client/features/src/accounts-manager/property/PropertyForm.tsx
@@ -36,7 +36,12 @@ export default function PropertyForm({ mode, defaultValues, onSubmit }: Props) {
             <section className="space-y-4 mb-8">
                 <h6 className="text-white uppercase">Location</h6>
                 <div className="space-y-4">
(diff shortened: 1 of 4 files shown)


libs/client/features/src/accounts-manager/property/PropertyForm.tsx
```diff
@@ -36,7 +36,12 @@ export default function PropertyForm({ mode, defaultValues, onSubmit }: Props) {
             <section className="space-y-4 mb-8">
                 <h6 className="text-white uppercase">Location</h6>
                 <div className="space-y-4">
-                    <Input type="text" label="Country" value="United States" readOnly disabled />
+                    <Input
```
Shpigford, line 39: This should be a dropdown list of all countries instead of a plain-text input.

pablo-kitz: adjusted on latest commit

Shpigford:
Has been implemented with #116

PR #114, opened 2024-01-16
Add ngrok instructions

### LINKED ISSUE #85: Add instructions for Ngrok usage
For development and for local self-hosting, we're defaulting to using Ngrok for HTTPS URLs (needed for basically any external data source).

We need to add instructions on how to set up an Ngrok account with a static domain and how to set the key + domain in .env.

Relevant discussion: #66

Neumand:
/fix #85 

This PR adds instructions for setting up ngrok with a free static domain.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
         ports:
             - 4551:4551
         environment:
-            - DOMAIN=${NGROK_DOMAIN}
+            - DOMAIN=${NGROK_DOMAIN:-host.docker.internal}
             - PORT=3333
             - AUTH_TOKEN=${NGROK_AUTH_TOKEN}
             - DEBUG=true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
         ports:
             - 4551:4551
         environment:
-            - DOMAIN=host.docker.internal
+            - DOMAIN=${NGROK_DOMAIN}
             - PORT=3333
             - AUTH_TOKEN=${NGROK_AUTH_TOKEN}
             - DEBUG=true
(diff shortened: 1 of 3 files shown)


docker-compose.yml
Shpigford, line 1: What happens if you *don't* setup ngrok and then try to run docker?

Neumand: Good point.

It will set it up with an empty string. I'll change it to use `-host.docker.internal` as the default value.

jbolda (approved):

Shpigford:
I'm currently getting the following error when trying to do this...

```
Your account is on the 'Pay-as-you-go' plan, which requires you to register a 
Domain through the ngrok Dashboard before starting an Endpoint. 
Visit https://dashboard.ngrok.com/cloud-edge/domains to register a domain 
and then use the `--domain` flag when starting your Agent.
```

I've registered a domain and have `NGROK_DOMAIN` set in `.env.`

Neumand:
> I'm currently getting the following error when trying to do this...
> 
> ```
> Your account is on the 'Pay-as-you-go' plan, which requires you to register a 
> Domain through the ngrok Dashboard before starting an Endpoint. 
> Visit https://dashboard.ngrok.com/cloud-edge/domains to register a domain 
> and then use the `--domain` flag when starting your Agent.
> ```
> 
> I've registered a domain and have `NGROK_DOMAIN` set in `.env.`

Hmm, I assume you also have `NGROK_AUTH_TOKEN` set?

And if you visit https://dashboard.ngrok.com/cloud-edge/domains you can see your domain, right?

image

Strange. I'd like some others to try and see what happens. I'm not getting that error.

Neumand:
@jbolda Can you try and see what happens?

Shpigford:
Here's what i've got...

CleanShot 2024-01-18 at 11 16 45@2x

CleanShot 2024-01-18 at 11 17 35@2x

Neumand:
Seems like it has to do with the account type.

It should work if you try with a different free account. Not sure how to proceed though...

image

PR #115, opened 2024-01-16
Set 'typescript.tsdk' version in VSCode settings

eneskaya:
Tell vscode to use the installed TypeScript version instead of the globally installed. So everyone (who is using vscode) is using the correct TS version :)


.vscode/settings.json
```diff
@@ -8,5 +8,6 @@
     },
     "[html]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode"
-    }
+    },
+    "typescript.tsdk": "node_modules/typescript/lib"
```
tmyracle, line 12: should we add `"typescript.enablePromptUseWorkspaceTsdk": true` to prompt folks? otherwise lgtm

eneskaya: That's a good idea! Didn't know about that option. Going to add it :)

eneskaya: Added!

Issue #19, opened 2024-01-10
Add Teller as primary bank connector option along with Plaid/Finicity


Shpigford:
The original app made extensive use of Plaid + Finicity for bank data. Unfortunately they both require enterprise contracts to make use of.

A much more human-friendly alternative is Teller. While they don't offer quite as much coverage on the bank connection front (yet), they do offer billing that's essentially free for a single user. So, this is likely the best option for a personal finance app like this that someone may run just for themselves.

https://teller.io 

We have the general concept/framework of multiple data providers already in the app where data from each gets cleaned up and distilled in to more common tables.

This project would mean adding Teller as another data provider and ultimately using whichever provider there are API keys for.

Ultimately this sets the stage for adding any data provider down the road (though that specifically is outside the scope of this).

This particular issue is about getting a banking provider added ASAP so Maybe can have utility without needing Plaid/Finicity accounts.

Rough deliverables:
1. Institution listing (for use in Institution search) — This needs to automatically update as Teller adds support for new institutions.
2. Abstracting bank connections away from user accounts — Right now we've got hard-coded columns for Plaid/Finicity. We need to pull those out with the goal of users ultimately being able to add many different connectors.
3. Checking account support
4. Savings accounts support
5. Investment account support
6. Remove any current blocking dependencies of Plaid/Finicity — I _think_ we've got some code in there that still relies on data sourcing from Plaid/Finicity
7. Readme user instructions on adding Teller support (presumably just via an .env variable API key)

## Note about the bounty
I've added a $500 bounty on this but want to be clear: this is a very complex task that may very well require multiple people working together on it.

We won't outright assign this issue to anyone at this point and depending on the number of people involved, there's a decent chance we'll split the bounty between contributors.

**EDIT: Teller (@stevegraham) has committed another $500 to this, bringing to total to $1000!**

jclusso:
Why not use [Stripe's Financial Connections](https://stripe.com/financial-connections). I thought I remember you stating data issues was a huge PITA. I feel like if anyone is going to get this right for certian, Stripe is your best bet.

Shpigford:
> Why not use [Stripe's Financial Connections](https://stripe.com/financial-connections). I thought I remember you stating bank data issues was a huge PITA. I feel like if anyone is going to get this right for certian, Stripe is your best bet.

Thinking about it from a "personal use" perspective, Stripe's solution isn't a great fit. You essentially need to have a business to even open a Stripe account. Plus, it's not free.

[...]

PR #119, opened 2024-01-17
Feature: Support Teller Part 3: Front End

tmyracle:
More progress on #19 This picks up where PR #98 left off.

This set of changes:
- Add Teller router
- Add front end hooks
- Updates to previous PR changes that had bugs
- Add Teller provider enum
- Add Teller Connect

Left to fix on this PR:
- [x] Figure out why I'm getting authentication errors (403) from Teller when syncing accounts (it also works sometimes though?)
- [x] Update the big bank images on the Add Bank modal to initialize Teller instead of Plaid
- [x] Update the dbgen expression for Transaction categories to map Teller categories
- [x] See if I can get rid of the screen flash when Teller Connect mounts
- [x] Study the syncing code path more and figure out the optimal way to initialized the accounts and transactions
- [x] Close Add Account modal after Teller Connect is resolved
- [x] ETL tests
- [x] Add test data
- [x] Integration tests
- [x] Add Teller instructions to README

Here's a demo video using the Teller sandbox:

https://github.com/maybe-finance/maybe/assets/1218724/09823c1c-2e47-48ce-9fd6-439fdd652c3a


libs/server/features/src/providers/teller/teller.service.ts
```diff
@@ -115,12 +126,67 @@ export class TellerService implements IAccountConnectionProvider, IInstitutionPr
...
+                type: 'teller' as SharedType.AccountConnectionType,
+                tellerEnrollmentId: enrollment.enrollment.id,
+                tellerInstitutionId: institution.id,
+                tellerAccessToken: this.crypto.encrypt(enrollment.accessToken),
+                userId,
+                syncStatus: 'PENDING',
+            },
+        })
+
+        await this.sync(accountConnection, { type: 'teller', initialSync: true })
```
tmyracle, line 180: there's probably a better way to do this, need to study the syncing paths better and the worker services

libs/teller-api/src/teller-api.ts
```diff
@@ -153,15 +153,10 @@ export class TellerApi {
...
-                // Add the access_token to the auth object
-                config.auth = {
-                    username: 'ACCESS_TOKEN',
-                    password: accessToken,
-                }
-                return config
+                auth: {
+                    username: accessToken,
+                    password: '',
+                },
```
tmyracle, line 159: there might be something funky going on here with the API singleton. The accessToken needs to be updated if you're getting accounts for different providers since the accessToken is unique to an enrollment/AccountConnection

libs/client/shared/src/hooks/useTeller.ts
```diff
@@ -0,0 +1,180 @@
...
+                    'teller.error.code': failure.code,
+                    'teller.error.message': failure.message,
+                },
+            })
+        },
+    } as TellerConnectOptions
+}
+
+// Custom implementation of useTellerHook to handle institution id being passed in
+export const useTellerConnect = (options: TellerConnectOptions, logger: Logger) => {
```
tmyracle, line 45: This is lifted from the Teller github and modified so that I can pass in the provider id that is selected in the AccountTypeSelector component which then gets fed to the config before initializing the Teller Connect instance. It feels kinda janky, would love advice here.

prisma/migrations/20240117191553_categories_for_teller/migration.sql
```diff
@@ -0,0 +1,223 @@
...
+
+ALTER TABLE
+  "transaction"
+ADD
+  COLUMN "category_user" TEXT;
+
+ALTER TABLE
+  "transaction"
+ADD
+  COLUMN "category" TEXT NOT NULL GENERATED ALWAYS AS(
```
tmyracle, line 20: I am not a fan of these generated expressions living in the database, I would prefer to move all of this logic into the ETL path in the future so that changes don't end up causing all sorts of migration issues. A future task and QOL improvement would be to rip out all these dbgenerated expressions from the database and schema and port them to server code.

tmyracle:
Alrighty let's do it live. One big thing to note: the generated column expressions that were added in the previous migrations in the life of the project make it an pain to change them. Unfortunately, if you already have a database running with data in it you will need to nuke it and start fresh or the migrations will fail. Fine for now, but this needs to be fixed (probably in a separate issue). I'd like to move that logic out of database generated statements and into the server or workers, most of it can be done in the ETL flows I'd imagine. 

TL;DR: if you run into issues running migrations drop your database/volume (delete from docker), start a fresh database, and then run them. Suggestions on how to unbork the migrations also very welcome, I spent a good hour or two trying to get around it in my migration but there's a fun web of materialized views and dependencies that makes it tricky. Someone better at databases than me can probably figure it out though 😄

Shpigford:
@tmyracle I may have (probably) missed this, but is there anything that needs to be run to get institutions to show up in the search?

I've got Teller .env variables set. But still getting no institutions in the search.

CleanShot 2024-01-18 at 08 23 12@2x

tmyracle:
> @tmyracle I may have (probably) missed this, but is there anything that needs to be run to get institutions to show up in the search?
> 
> 
> 
> I've got Teller .env variables set. But still getting no institutions in the search.
> 
> 
> 
> CleanShot 2024-01-18 at 08 23 12@2x

the worker should go out and grab those in the background and populate them, there's a sync-teller-institutions job that does it. might see if that cron job has run yet

Shpigford:
> the worker should go out and grab those in the background and populate them, there's a sync-teller-institutions job that does it. might see if that cron job has run yet

Not seeing anything in the console about that cron job running. Though I do see both `sync-finicity-institutions` and `sync-plaid-institutions`

PR #125, opened 2024-01-17
fix repeated data-testid

karan0handa:
It seems that the `data-testid` for the year selector was incorrectly copied from the month selector.

This fixes that.


Neumand (approved):
👍🏼

Issue #126, opened 2024-01-17
DatePicker.spec.tsx 'should be able to open modal and select a date' test failing because of faulty test setup.


karan0handa:
**Describe the bug**
Test failure. Probably something with the test setup because component works perfectly otherwise. This is the only test that fails in the `design-system` project.

**To Reproduce**

Run `yarn jest /libs/design-system/src/lib/DatePicker/DatePicker.spec.tsx`

**Expected behavior**
Test failure.

**Screenshots**
![image](https://github.com/maybe-finance/maybe/assets/79533543/9623fe51-8c59-4c81-a76e-3327da5479e6)

**Desktop (please complete the following information):**

-   OS: Ubuntu 22.04

tmyracle:
This is resolved now with #127 👍, can close @Shpigford

PR #127, opened 2024-01-17
fix date-picker tests

shyrynaz:
This PR resolves issue #126.

This fix removes the fixed date-time set by `beforeAll(() => jest.useFakeTimers().setSystemTime(new Date('2021-10-29 12:00:00')))`  and uses the current system time


Shpigford:
Can you provide some detail on what this fix is doing?

kundan621 (changes requested):

Issue #81, opened 2024-01-15
Date inputs should respect users date format


tqwewe:
**Is your feature request related to a problem? Please describe.**
I struggled to finish the onboarding process since I was entering my birthdate as `DD/MM/YYYY` instead of the `MM/DD/YYYY` it was expecting. I also didn't receive any validation errors and was able to continue onto the next question in the form.

**Describe the solution you'd like**
Date inputs should respect the users local format for dates. Some countries like the US have month first, and others such as Australia have the date first.

Additionally, the birthdate question in the on-boarding form should not let you continue if you've entered an invalid birthdate.

imanaspaul:
@Shpigford I would like to pick that up, Can I?

tmyracle:
Here's my proposed solution:

- Move location question before birthday so we know what country the user is in (rather than trying to guess via IP or something)
- Add a field in [geo-utils.ts](https://github.com/maybe-finance/maybe/blob/main/libs/shared/src/utils/geo-utils.ts) for the date format for that country according to [date format by country](https://en.wikipedia.org/wiki/Date_format_by_country)
- Add a prop to DatePicker component that takes in the date format and parses the date accordingly
- Update use of DatePicker in birthday question to use the selected country to return the appropriate date format from  [geo-utils.ts](https://github.com/maybe-finance/maybe/blob/main/libs/shared/src/utils/geo-utils.ts)
- Update all other uses of DatePicker to use the country field on the User to set the appropriate format

Shpigford:
@tmyracle Your solution sounds good.

bhardwajakshit:
@Shpigford @tmyracle I'd be happy to take up this issue and contribute.

ludvigrosenqvist:
@tmyracle Wouldn't a simpler solution be to use the standardized date format of ISO8601, YYYY-MM-DD? This skips having to take every locale date format into account. If needed, you can simply parse the date to US formatting like so:

```javascript
const date = new Date('2019-04-12');
const usFormattedDate = date.toLocaleDateString('en-US')

tmyracle:
> @tmyracle Wouldn't a simpler solution be to use the standardized date format of ISO8601, YYYY-MM-DD? This skips having to take every locale date format into account. If needed, you can simply parse the date to US formatting like so:
> 
> 
> 
> ```javascript
> 
> const date = new Date('2019-04-12');
> 
> const usFormattedDate = date.toLocaleDateString('en-US')

[...]

PR #132, opened 2024-01-17
Issue #81: Date inputs should respect users date format

Shivam0710:
I have done changes as follows:

- Moved location question before birthday so we know what country the user and based on the country of I've maintained a function where we can add the value based on the country code.
- Added fields in geo-utils for dateFormat so that we can pick the dateFormat based on country code.
- Added a prop to DatePicker component that takes in the date format and parses the date accordingly and sets an error message if the existing if (inputData.isValid) check fails.
- Removed hard coded formats in DatePicker component and made them dynamic based on the dateFormat provided.


libs/design-system/src/lib/DatePicker/DatePicker.tsx
```diff
@@ -25,11 +25,41 @@ export interface DatePickerProps {
...
+        .map((keyword) => keyword.toUpperCase())
+}
+
+function getDateFormatPattern(dateFormat: string): string {
+    const dateComponents = dateFormat.split('/')
+    const pattern = dateComponents
+        .map((component) => {
+            if (component.includes('d')) {
+                return ' ## ' // Placeholder for day
+            } else if (component.includes('m') || component.includes('M')) {
```
tmyracle, line 52: This is looking great, thanks for taking this on! Just a few small questions:

- Do we need to check both upper and lower case here? If so, should we do it for day and year as well?
- In the getMaskArray could you use a regex to remove slashes and spaces in one go, then upper case and split to array?

Shivam0710: Concerning the initial item, we only need to address "M," but I will adjust all of them to perform case-insensitive checks for added safety. As for the second point, I will manage the modifications accordingly.

Shivam0710: I've addressed the comments. Kindly review the pull request and inform me if any further modifications are necessary.

libs/design-system/src/lib/DatePicker/DatePicker.tsx
tmyracle, line 1: After the user has set their country can we use that to set the format throughout the app. I think the changes so far only update it for the Profile step but everywhere else in the app will still be the default it was before.

Shivam0710: Yes the issue is related to Profile only but we can save the country code in the cookies and retrieve it when picking up the format. Can you please list down the scenarios where we'll have to take care of this? I'll handle this accordingly.

tmyracle: DatePicker is used in multiple places in the app like <UserDetails />, <ValuationTableForm />, etc. Can just do a search for all the uses of DatePicker.

Shivam0710: Closed this by mistake. I have opened PR for the same

Issue #52, opened 2024-01-14
Update tests to replace Auth0


tmyracle:
The current tests use a lot of Auth0 helpers, we'll need to rip all of that out since we've moved to NextAuth. Guide here on how to set up cypress testing with NextAuth: [Testing with Cypress](https://next-auth.js.org/tutorials/testing-with-cypress)

As part of this we should ensure there is some basic e2e test coverage for login and registration.

PR #135, opened 2024-01-18
Fix the failing integration tests

tmyracle:
This change fixes the failing integration tests. Part of #52 , next change will be to add a new cypress e2e test
CleanShot 2024-01-18 at 01 55 00

- Replace hardcoded test auth ids so database calls find things based on the correct test user
- Added back in the proper values for finicityInstitutionId and finicityInstitutionLoginId from the test data
- Updated the mocked Axios instance to use a dummy JWT so that the tests that hit the server authenticate properly
- Add fallback JWT handling for auth middleware in the event you can't send cookies
- Make the cookie name we're looking for dynamic based on whether the request is https or not
- Reimplements the original crypto service, was running into some bugs with parsing the plaid accessTokens, can revisit this later

### DIFF
--- a/apps/server/src/app/middleware/validate-auth-jwt.ts
+++ b/apps/server/src/app/middleware/validate-auth-jwt.ts
@@ -2,17 +2,20 @@ import cookieParser from 'cookie-parser'
 import { decode } from 'next-auth/jwt'
 
 const SECRET = process.env.NEXTAUTH_SECRET ?? 'REPLACE_THIS'
-
 export const validateAuthJwt = async (req, res, next) => {
     cookieParser(SECRET)(req, res, async (err) => {
         if (err) {
             return res.status(500).json({ message: 'Internal Server Error' })
         }
 
-        if (req.cookies && 'next-auth.session-token' in req.cookies) {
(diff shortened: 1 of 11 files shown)


apps/server/src/app/middleware/validate-auth-jwt.ts
```diff
@@ -2,17 +2,20 @@ import cookieParser from 'cookie-parser'
...
 const SECRET = process.env.NEXTAUTH_SECRET ?? 'REPLACE_THIS'
-
 export const validateAuthJwt = async (req, res, next) => {
     cookieParser(SECRET)(req, res, async (err) => {
         if (err) {
             return res.status(500).json({ message: 'Internal Server Error' })
         }
 
-        if (req.cookies && 'next-auth.session-token' in req.cookies) {
+        const cookieName = req.secure
```
tmyracle, line 11: @6ixfalls I think this will address the issue we were talking about

6ixfalls: Alternatively you could also just force the cookie to be secure on the next auth side, which is what I did but this is fine.

Issue #151, opened 2024-01-18
Hide upgrade prompts + billing elements/references


Shpigford:
If `STRIPE_API_KEY` isn't present in .env, all upgrade prompts and billing references should be hidden.

Shpigford:
Implemented with #154

PR #154, opened 2024-01-18
Remove billing & upgrade references when STRIPE_API_KEY is missing

simonbukin:
This PR takes a stab at removing upgrade and billing references when STRIPE_API_KEY is not set in the `.env` file. #151 

I'm unsure if there are other places that reference billing and upgrade, but I would be happy to update those as well. Just spun up this repo this morning, so I'm unsure if there are other places these are mentioned.

### DIFF
--- a/apps/workers/src/main.ts
+++ b/apps/workers/src/main.ts
@@ -154,11 +154,13 @@ syncInstitutionQueue.add(
  */
 sendEmailQueue.process('send-email', async (job) => await emailProcessor.send(job.data))
 
-sendEmailQueue.add(
-    'send-email',
-    { type: 'trial-reminders' },
-    { repeat: { cron: '0 */12 * * *' } } // Run every 12 hours
-)
+if (env.STRIPE_API_KEY) {
+    sendEmailQueue.add(
+        'send-email',
+        { type: 'trial-reminders' },
+        { repeat: { cron: '0 */12 * * *' } } // Run every 12 hours
+    )
+}
 
 // Fallback - usually triggered by errors not handled (or thrown) within the Bull event handlers (see above)
 process.on(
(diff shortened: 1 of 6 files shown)


tmyracle:
There's also an email that's sent on a scheduled job for trial upgrade reminders in email.processor.ts, we can noop that in main.ts in the workers folder if no api key.

simonbukin:
@tmyracle done!

apps/workers/src/main.ts
```diff
@@ -154,11 +154,13 @@ syncInstitutionQueue.add(
  */
 sendEmailQueue.process('send-email', async (job) => await emailProcessor.send(job.data))
 
-sendEmailQueue.add(
-    'send-email',
-    { type: 'trial-reminders' },
-    { repeat: { cron: '0 */12 * * *' } } // Run every 12 hours
-)
+if (env.STRIPE_API_KEY) {
```
tmyracle, line 157: lets also make sure there's not a default value in the workers .env file that would cause a false positive

tmyracle: 🤦  yeah you're right, I was thinking of the NX_STRIPE_SECRET_KEY which I'm not sure if even used anywhere 🤔

simonbukin: Yeah, I saw that too... We're probably going to make some updates to the Zod schema once we figure out what each env variable is doing

tmyracle: For sure, will leave that to another PR. LGTM!

Shpigford:
Thanks @simonbukin!

cmosguy:
@simonbukin and @Shpigford there is still a strip container in the `docker-compose.yml` .  Should this be removed as well?

Shpigford:
> @simonbukin and @Shpigford there is still a strip container in the `docker-compose.yml` . Should this be removed as well?

No, as we'll still be making use of Stripe for our hosted solution. This PR is just about hiding UI elements.

cmosguy:
Ah understood - ok thank you!

PR #160, opened 2024-01-19
Fix test case in `date-utils.spec.ts`

### LINKED ISSUE #182: Failing test case `date-utils.spec.ts`
**Describe the bug**
`date-utils.spec.ts` test case fails.

**To Reproduce**

1. `yarn jest libs/shared/src/utils/date-utils.spec.ts` in console.

karan0handa:
### Description
Fixes #182.

"converts between years and ages › dob 2022-10-15 is 0 years old today" test case fails because of rounding error and outputs -1. This fixes by making the minimum age be 0 as opposed to -1. 

This happens because the `dobToAge` function uses `{ zone: 'utc' }` but the test case doesn't.

### Screenshots
Fixed - 
image


Lewiscowles1986:
Doesn't this just mask an underlying date utility issue?

karan0handa:
@Lewiscowles1986 I initially thought it was an issue with DateTime (usually is the case, which is why libraries like Moment and DayJS are used), but turns out that the the function was converting it to a different time zone. 

Have normalized the test timezones now.

tmyracle (approved):

PR #165, opened 2024-01-19
Fix session load

### LINKED ISSUE #181: Refreshing or manually changing URL automatically redirects to `/`
**Describe the bug**
If you refresh the page, or redirect to another page directly using a URL, you're redirect back to `/login` and then to `/`.

**To Reproduce**
1. Login
2. Navigate (using UI) to `localhost:4200/accounts`
3. Refresh
4. See error.
---
1. Login
2. Navigate (using URL) to `localhost:4200/accounts`
3. See error

**Expected behavior**
It should stay on the same route when refreshed and go to the appropriate route using the URL.

**Video**

[](https://github.com/maybe-finance/maybe/assets/79533543/08c6b2ef-eb32-44f2-a1a8-c8e64323de1a)

**Desktop (please complete the following information):**

-   OS: Ubuntu 22.04
-   Browser: Google Chrome

karan0handa:
### Description
Fixes #181. If you refresh the page, or redirect to another page directly using a URL, you're redirect back to `/login` and then to `/`.

This happens because the session doesn't load instantly and the `useEffect` in `WithAuth` redirects to `/login`.


apps/client/pages/_app.tsx
```diff
@@ -33,14 +33,18 @@ Sentry.init({
 
 // Providers and components only relevant to a logged-in user
 const WithAuth = function ({ children }: PropsWithChildren) {
-    const { data: session } = useSession()
+    const { data: session, status } = useSession()
+    const [isLoading, setLoading] = useState(true)
```
tmyracle, line 37: Is the state necessary since the useSession hook already provides that?

tmyracle (approved):

Issue #162, opened 2024-01-19
Unintuitive Logout Placement


karan0handa:
**Is your request related to a problem? Please describe.**
Spent atleast 5 minutes searching for the logout button. Expected it to be there when clicking on the user avatar.
Current flow involves maximizes sidebar, clicking on the three dots and then it being there. Not an intuitive flow.

**Describe the solution you'd like**
![image](https://github.com/maybe-finance/maybe/assets/79533543/f0563c0a-77fd-4fd6-a9c4-77327b12e23b)

The three dot menu should appear when I click the user avatar. I think this is much more intuitive.

**Describe alternatives you've considered**
Another alternative is adding a logout button in settings, but that wouldn't be needed if the above solution is implemented.

ikraamdaanis:
I'll like to tackle this: https://github.com/maybe-finance/maybe/pull/167

Shpigford:
Fixed with #167

PR #167, opened 2024-01-19
Logout Placement

ikraamdaanis:
Issue: https://github.com/maybe-finance/maybe/issues/162

Made the following changes in desktop mode:

- Removed the `<ProfileCircle/>`'s link to `/settings`
- Moved the `<MenuPopover/>` into the `<ProfileCircle/>`
- Added `buttonClassName` to `<MenuPopover/>` so we can customise the styling of `<Menu.Button/>`

Now when you press the profile icon, instead of going directly to settings, the menu popover appears and you can choose to go to settings, fix data or logout.

Before:

image

After:

image

Feedback is appreciated.

### DIFF
--- a/libs/client/features/src/layout/DesktopLayout.tsx
+++ b/libs/client/features/src/layout/DesktopLayout.tsx
@@ -16,7 +16,6 @@ import {
     RiFolderOpenLine,
     RiMenuFoldLine,
     RiMenuUnfoldLine,
-    RiMore2Fill,
     RiPieChart2Line,
     RiFlagLine,
     RiArrowRightSLine,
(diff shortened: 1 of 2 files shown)


simonbukin:
From what I can tell, this removes the ability to open user settings when clicking on the user profile icon. Is this intended? Clicking on the user profile to open up their settings seems like a pretty standard pattern to me, but would be happy to hear your thoughts.

ikraamdaanis:
> From what I can tell, this removes the ability to open user settings when clicking on the user profile icon. Is this intended? Clicking on the user profile to open up their settings seems like a pretty standard pattern to me, but would be happy to hear your thoughts.

This is still work in progress. Bringing up the menu instead when you click it:

Screenshot 2024-01-19 at 19 04 50

Shpigford:
Good job on this @ikraamdaanis!

Issue #131, opened 2024-01-17
Manually input + automatically track US stocks


Shpigford:
Currently we have the ability to manually enter the value of any asset (and then you can manually add updated values).

What need to add support for is entering a particular stock and then tracking the ongoing value of it.

For now, due to data sourcing, this would be restricted to the US stock market.

Data to input:
1. Stock
2. Number of shares
3. Cost basis (optional)

This will be the basis for allowing users to import CSV files of their various holdings, but that specific functionality is outside the scope of this issue.

The _backend_ essentially already supports this as we were importing this data via Plaid, but there's a current Issue to clean this up specifically for manual usage: #166

So, a lot of this (maybe most?) is essentially building the UI/UX to input the individual stocks.

---

Here's a rough mockup of potential UI.

![CleanShot 2024-01-18 at 15 30 26@2x](https://github.com/maybe-finance/maybe/assets/35243/4e5f9408-2146-4bbc-939f-c09b67d1322a)

1. Select an existing manual investment account (or create a new one).
2. Enter a stock symbol (though should probably pull from a database list of these so you can't just enter any random string)
3. Set the purchase date, total purchase value (cost basis) and the number of shares

From there, Maybe would then track the value of that investment automatically.

### Deliverables:

1. UI for adding a manual investment account — Would use the existing "Investment" block in the modal, where you'd either pick an existing account or add a new one.
2. UI for adding the stock — Should be a "search to select" UI for pulling from stock listing (Database: `Security`)
3. Creating the appropriate "transaction" for adding the stock.
4. UI for removing/selling the stock
5. Create the appropriate "transaction" for removing the stock.

### Bounty
There is a $250 bounty for this.

Shpigford:
> What would be the source for US stocks ? What about ETFs ?
Currently the app is using Polygon.io, which has all data for all securities in the US market (stocks, ETFs, etc). Highly recommend using that as they have a free plan for individual use.

> Thinking ahead for outside of US. What's a stock ? A ticker ? a ISIN code ? Both ? A ISIN code and a currency ?
We likely need to do a bit of research in this regard to see what international standard there is (if any). On the surface Symbol + Exchange _should_ be a unique identifier available everywhere, but again...need to do some digging there.

lasconic:
https://www.alphavantage.co seems to have a free plan and some international coverage (LVMH in Paris MC.PAR is in for example)

Shpigford:
[...]

PR #169, opened 2024-01-19
Front end for /claim #131

bskinny129:
partial /claim #131


Shpigford:
Thanks @bskinny129. There's quite a bit more work that would need to be done on this to really be up for a partial claim as this is essentially just a UI rough-in that doesn't really account for any interaction/functionality.

bskinny129:
Understood. Yup, I don't expect anything here, so no worries. Hopefully the next person to take it on can build on this to save some effort.

I just realized I wouldn't have enough time to see it all the way through in the next couple of days, so thought it best to pull the plug where I was!

gamedevsam:
I'm interested in picking up this bounty and seeing this through to completion. I plan on working on it tonight, see if I can make some progress and continue pushing this PR forward.

Issue #173, opened 2024-01-19
Login page and design system improvements


aditya-67:
### Improvements to the login page

- Password complexity bar and requirements text pop-up can be removed from the Login page
<img width="612" alt="Screenshot 2024-01-20 at 3 18 01 AM" src="https://github.com/maybe-finance/maybe/assets/9695866/890e82dd-3b01-4543-b05b-74ba6e7a59dd">
 
<img width="612" alt="Screenshot 2024-01-20 at 3 17 45 AM" src="https://github.com/maybe-finance/maybe/assets/9695866/12c91295-b38b-400d-9039-eee49ca1b1e2">

- Button does not have any loader/spinner on click. UX can be improved by adding a spinner to the Button.

<img width="612" alt="Screenshot 2024-01-20 at 3 29 07 AM" src="https://github.com/maybe-finance/maybe/assets/9695866/16686e8a-8c09-4460-b5db-fe091533d5d5">

Shpigford:
Implemented with #174

PR #174, opened 2024-01-19
Login and Register page improvements

aditya-67:
This PR fixes the issues mentioned in #173 

Following are the changes:
- Removed the validation and complexity bar from password input in login page
- Added a SVG spinner to the Button component and a flag to toggle the spinner 
- Updated the buttons in Login and Register page 

Screenshot 2024-01-20 at 3 42 33 AM

Screenshot 2024-01-20 at 3 43 18 AM

From the below recording, in Register page, do we need to clear the input on submit? If there is an error in registration, all the fields have to filled again. Would appreciate your thoughts on this

https://github.com/maybe-finance/maybe/assets/9695866/d01d3ac2-cd6a-4940-9836-c3bb2c178138

@Shpigford I can try and improve the current design system. If the current `Button` implementation is all good, maybe we can use the loader in buttons as need in the app

### DIFF
--- a/apps/client/pages/login.tsx
+++ b/apps/client/pages/login.tsx
@@ -1,6 +1,6 @@
 import { useState, type ReactElement } from 'react'
 import { FullPageLayout } from '@maybe-finance/client/features'
-import { Input, InputPassword, Button, LoadingSpinner } from '@maybe-finance/design-system'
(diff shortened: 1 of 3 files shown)


Shpigford:
This is a great start @aditya-67! Love little details and improvements like this. Thank you!

Issue #176, opened 2024-01-20
Unable to complete onboarding


tmyracle:
**Describe the bug**
When I sign up as a new user I land on the welcome page, when I click on the first button to go to the next step in the onboarding I hit an error. 

![CleanShot 2024-01-19 at 22 34 06](https://github.com/maybe-finance/maybe/assets/1218724/f98cc641-1a8b-4f4b-8d7f-6029ace63b1a)

This is what I'm seeing in the console:
`input parse error ZodError: [
  {
    "code": "invalid_union_discriminator",
    "options": [
      "main",
      "sidebar"
    ],
    "path": [
      "flow"
    ],
    "message": "Invalid discriminator value. Expected 'main' | 'sidebar'"
  }
]`

I checked the payload that's being sent to the backend and it is valid:
![CleanShot 2024-01-19 at 22 42 24](https://github.com/maybe-finance/maybe/assets/1218724/2860b011-6263-41c6-94dc-ed0c64b4f5c9)

**To Reproduce**

1. Pull latest main branch and start up app
2. Register as new user
3. Click on the continue button

**Expected behavior**
It should proceed through onboarding as expected.

**Screenshots**

**Desktop (please complete the following information):**

-   OS: MacOS 14.1.1.
-   Browser: Brave [Version 1.60.118 Chromium: 119.0.6045.163 (Official Build) (arm64)](https://brave.com/latest/), also verified on Safari Version 17.1 (19616.2.9.11.7)

**Additional context**
Steps taken to try to resolve:

- Fresh pull of the repo main branch
- rm -rf node_modules and reinstall
- traced codepath looking for potential bugs, logged and tested along the way
- tried incognito browser
- tried multiple different registrations
- tried hardcoding the request to a valid payload, validation still fails
- restarted computer

mrkirthi-24:
thanks @tmyracle for solving it. I needs to be merged asap.

Shpigford:
solved with #177

PR #157, opened 2024-01-18
feat: remove Finicity usage

MichaelDeBoey:
Closes #146

---

Alternative to #155 but with migrations and a couple of extra removals

PR #177, opened 2024-01-20
Add back json handling to express

tmyracle:
Looks like this was accidentally removed which breaks all the API requests
Solves issue #176


aditya-67:
Yes, was about to raise a PR for the same! Thank you

MichaelDeBoey (approved):
Oops, I should have only removed the limit in #157, not the usage 😅🙈

PR #183, opened 2024-01-20
Remove dead Maybe.co links from sidebar resources

noonfyi:
Removes from the sidebar the Resources/Learn More links to Maybe.co resources that are no longer available.
It would be spectacular for these to return in some form someday as internal documents, external resources, or back on Maybe.co, but for now they are inaccessible.


Shpigford:
Sorry to shut this one down, but I'll actually have these links working again in the next 7-10 days. 🫣

noonfyi:
That's great news!

Issue #137, opened 2024-01-18
Replace Postmark with a generic SMTP provider


6ixfalls:
**Is your request related to a problem? Please describe.**
Maybe relies on the Postmark API, a proprietary email sending service. Ideally, Maybe should support all kinds of mail providers, which can be easily done through SMTP, rather than forcing users to lock into Postmark.

**Describe the solution you'd like**
Postmark can either be replaced by an SMTP sender, or an option can be added to either send with Postmark or send through SMTP. Replacing Postmark with an SMTP sender should be the most useful, as Postmark also has SMTP relays/servers that can be used to send mail, removing the need to maintain two separate mail systems.

Shpigford:
Approved on our end. Reducing explicit third-party dependencies is definitely always a goal.

tmyracle:
Closed by #230

PR #185, opened 2024-01-20
Abstract away Postmark as email provider

tmyracle:
This is progress towards being able to support #137 

This creates a generic email service that can use one of multiple providers. In the future we can then just add a provider implementation for SMTP, Sendgrid, etc. Each *.provider.ts handles the client specific implementation. I can add some docs to the wiki on how to add support for a new email service.

@Shpigford let me know what you think, if you like this direction I'll convert it to ready for review.


tmyracle:
Follow up PR on this will add nodemailer to support SMTP to close out #137

tmyracle:
CleanShot 2024-01-22 at 09 43 24

Issue #150, opened 2024-01-18
Abstract data provider away from Transactions table


Shpigford:
We're undertaking a large project to abstract data providers away from tables as we're working towards support for dozens/hundreds of data providers globally.

Relevant columns:
* `teller_category`
* `teller_transaction_id`
* `teller_type`

mattgillooly:
I see several issues that look similar to this: #147, #148, #149.  Is the idea that these would initially all be refactored to use the same `TellerDataProvider`, or would `Account`, `AccountConnection`, `Transaction` and `User` each have a type-specific data provider that spoke to Teller?  e.g. `Teller::TransactionDataProvider`, etc.

PR #188, opened 2024-01-20
refactor: remove `Transaction` `category` DB generation

MichaelDeBoey:
As @tmyracle mentioned in https://github.com/maybe-finance/maybe/pull/119#issuecomment-1897697437

>  I'd like to move that logic out of database generated statements and into the server or workers, most of it can be done in the ETL flows I'd imagine

This also is a good step forward towards closing #150


tmyracle (approved):
Awesome stuff, thanks again!

PR #189, opened 2024-01-20
refactor: remove `InvestmentTransaction` `category` DB generation

MichaelDeBoey:
As @tmyracle mentioned in https://github.com/maybe-finance/maybe/pull/119#issuecomment-1897697437

>  I'd like to move that logic out of database generated statements and into the server or workers, most of it can be done in the ETL flows I'd imagine

This also fixes "Abstract data provider away from InvestmentTransactions table" (doesn't have a ticket though)


tmyracle:
Absolute legend, thanks for tackling this!

Issue #166, opened 2024-01-19
Import/maintain list of US stock tickers + prices


Shpigford:
## High level summary
We need to import all US stock tickers and then keep prices updated on an ongoing basis. This is foundational for future feature work.

There's code already written that uses Polygon.io for US stock data and that should be fine to get us started as they have a free plan for individual use.

A good bit of this is already done in the codebase, but it was based around only pulling in necessary tickers based on Plaid + Finicity imports.

We're about to allow people to manually enter any stock and have it track it, so we need to rework things a bit.

## Functional notes/scope
* Relevant database tables/models: Security, SecurityPricing
* A listing of all symbols: https://polygon.io/docs/stocks/get_v3_reference_tickers
* Stock data import will need to happen either on app setup or some sort of check each time the app is booted
* A daily job needs to be run to check for new tickers
* Should pull as much historical pricing data as Polygon allows for a given API key (they have different pricing plans if folks want to pay for more data). This is then stored in SecurityPricing.
* We should think about allowing for "real time" pricing. The Polygon free plan only has "daily" pricing data, but if you're on a paid plan you can get more updated data. Need some sort of background job for this.
* We need to track what stock exchange a given symbol is associated with (I don't believe this is currently tracked in Security)
* As far as database modeling goes, this should err towards flexibility to add additional stock exchanges and tickers in the near future.
* There's no frontend component to this. It's purely about establishing available data for use in other features.
* This will be used by other parts of the app, such as #131 

## Bounty
There's a $250 bounty for this.

mrkirthi-24:
Hi @Shpigford. Not able to login.

![Screenshot 2024-01-20 at 3 12 06 PM](https://github.com/maybe-finance/maybe/assets/53830546/781c97de-dbeb-4cb8-bf9a-3c8a61162602)

![Screenshot 2024-01-20 at 3 09 13 PM](https://github.com/maybe-finance/maybe/assets/53830546/306134f9-9acd-45e1-b824-81dd7e5182b3)

mrkirthi-24:
```
event - compiled client and server successfully in 917 ms (2693 modules)
input parse error ZodError: [
  {
    "code": "invalid_union_discriminator",
    "options": [
      "main",
      "sidebar"
    ],
    "path": [
      "flow"
    ],
    "message": "Invalid discriminator value. Expected 'main' | 'sidebar'"
  }
]
[...]

PR #195, opened 2024-01-21
Fetch Security Tickers Job

tmyracle:
Starting to get some things going for #166 
The rate limiting on Polygon is no joke and makes it a lot more annoying to deal with because anything hitting their APIs needs to be single threaded so you don't have 2 jobs going at the same time that are going to hit rate limits and crash the jobs. 

When you're on free I think this job takes like 30 minutes to finish or something. 

TIL, # of securities in for US stock markets (according to Polygon):
CleanShot 2024-01-21 at 03 11 53

US stock ticker sync:
CleanShot 2024-01-21 at 03 12 50

Pricing sync:
CleanShot 2024-01-22 at 11 01 22

Queue jobs:
CleanShot 2024-01-22 at 10 58 27

Dev tool added for syncing tickers:
CleanShot 2024-01-22 at 11 03 39

Test run with new test added:
CleanShot 2024-01-22 at 10 39 46


tmyracle:
Ok I think this is about ready to go!

Open question to resolve:
For syncing security prices everything works as expected on the starter tier. On the basic tier we're going to need to rethink the syncing strategy since the rate limits are pretty brutal (5 req/min). My plan is to have a follow up PR for syncing prices for the basic tier that runs once a day and just gets the closing price (basic doesn't support real time data). It'll also be scheduled for a time that shouldn't clash with the stock ticker sync since running both will hit rate limits.

PR #197, opened 2024-01-21
Add dev option for onboarding

### LINKED ISSUE #196: Seed the database with an account that has already completed migration
**Is your request related to a problem? Please describe.**
Whenever I reset the database migrations, I lose my account and have to go through onboarding again. 

**Describe the solution you'd like**
A simple solution to this is creating an account with the onboarding options already filled. An `AuthUser` needs to be created along with a connected `User` that has the correct onboarding JSON.

karan0handa:
### Description
Fixes #196. Builds on #200 to add the option to onboard a user.
1. Changes `AuthDevTools` to `UserDevTools`
2. Abstracts `UserDevTools`
3. Adds dev option to pre-onboard a user

image


prisma/seed.ts
```diff
@@ -52,6 +100,37 @@ async function main() {
             },
             update: {},
         }),
+        prisma.authUser.upsert({
+            where: {
+                id: 'test_f5ec79b4-8c49-4015-bc37-f2758923ef38',
+            },
+            create: {
+                id: 'test_f5ec79b4-8c49-4015-bc37-f2758923ef38',
+                firstName: 'Karan',
```
tmyracle, line 109: I think we should probably make the name something generic.

prisma/seed.ts
```diff
@@ -52,6 +100,37 @@ async function main() {
...
+            create: {
+                email: 'test@maybe.com',
+                firstName: 'Karan',
+                lastName: 'Handa',
+                authId: 'test_f5ec79b4-8c49-4015-bc37-f2758923ef38',
+                onboarding: onboarding,
+                household: 'single',
+                country: 'US',
+                maybe: 'Test',
+                createdAt: new Date('2024-01-20T15:21:17.889Z'),
```
tmyracle, line 129: Does this need to be set? I think the db will handle this.

prisma/seed.ts
karan0handa: Done.

prisma/seed.ts
karan0handa: Done.

6ixfalls:
Please add a check that only does this in development, I don't want to see a random user being created on my server especially if it's running in production.

tmyracle:
The more I've thought about this I don't feel great about creating users in the seeds file. I didn't like the pattern when I used it for testing and I just opened a PR to rework it to an alternate approach: #203 

What if instead we added a dev only button on the welcome/intro screen that lets you skip onboarding. The [AccountDevTools](https://github.com/maybe-finance/maybe/blob/main/libs/client/features/src/accounts-list/AccountDevTools.tsx) is an example of this dev only tool implementation. Thoughts?

karan0handa:
Your approach makes sense. I'll review your e2e PR is a bit, but I largely agree with that approach. There's no reason to keep the test user longer than it needs to exist.

With regards to having a dev option to skip onboarding. I'm onboard with that. The database shouldn't start with a user. With regards to the specifics, I think the dev options should give the ability to 
1. create an already onboarded user
2. create a completely new user
3. remove already onboarded user (if exists)
4. remove completely new user (if exists)
This dev option should be available at `/login` and `/register`. We can also have a dev option to skip onboarding for the current user in `/onboarding`.

This will be very similar looking to `AccountDevTools`.

apps/client/pages/register.tsx
```diff
@@ -1,11 +1,17 @@
 import { useState, type ReactElement } from 'react'
 import { Input, InputPassword, Button, Checkbox } from '@maybe-finance/design-system'
```
tmyracle, line 2: won't need the Checkbox here anymore

tmyracle:
Overall looking great! Just a few small changes and should be good to go.

apps/client/pages/register.tsx
```diff
@@ -43,6 +54,11 @@ export default function RegisterPage() {
             redirect: false,
         })
 
+        if (isOnboarded) {
```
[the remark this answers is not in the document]
karan0handa: Good catch!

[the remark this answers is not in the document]
karan0handa: `response.ok` should do the trick! (https://next-auth.js.org/getting-started/client)

```typescript
{
  /**
   * Will be different error codes,
   * depending on the type of error.
   */
  error: string | undefined
  /**
   * HTTP status code,
   * hints the kind of error that happened.
   */
  status: number
  /**
   * `true` if the signin was successful
   */
  ok: boolean
  /**
   * `null` if there was an error,
   * otherwise the url the user
   * should have been redirected to.
   */
  url: string | null
}
```

Shpigford:
@karan0handa You ready for me to merge this?

karan0handa:
@Shpigford Yes this is ready.

PR #203, opened 2024-01-22
Improvement: Rework account e2e test

tmyracle:
This change reworks how creating a test user for e2e. Rather than having it in the prisma seed file and it always being there we now dynamically create and destroy the AuthUser and User before and after tests run. It also adds back role based authorization to the e2e router. 

It also fixes some sql bugs :)


tmyracle:
CleanShot 2024-01-22 at 09 30 25

tmyracle:
CleanShot 2024-01-22 at 09 30 25

Issue #201, opened 2024-01-22
error - ./pages/_app.tsx Module parse failed: Unexpected token (51:11)


jijichen:
**Describe the bug**
```
error - ./pages/_app.tsx
Module parse failed: Unexpected token (51:11)
File was processed with these loaders:
 * ../../node_modules/next/dist/compiled/@next/react-refresh-utils/dist/loader.js
 * ../../node_modules/next/dist/build/babel/loader/index.js
You may need an additional loader to handle the result of these loaders.
|
|   if (session && status === 'authenticated') {
>     return <OnboardingGuard>
|                 <UserAccountContextProvider>
|                     <AccountContextProvider>
```

**To Reproduce**
- Followed README to setup the env and then start project
- Saw error when running `yarn dev`

**Desktop (please complete the following information):**

-   OS: macos 14.2.1 (23C71)
-   Browser chrome

**Additional context**
- commit : 
```
commit 06b861aea756247705537f49a95b2cf6285ef9cb
Merge: 42bb44b e5659ed
Author: Josh Pigford <josh@joshpigford.com>
Date:   Sun Jan 21 15:51:44 2024 -0600

    Merge pull request #200 from tmyracle/dev-tools-fix
    
    Fix #194: Fix dev tools

```

tmyracle:
This looks like some sort of weird babel error 🤔  it looks like whatever's compiling things has stripped out the parentheses from that return statement (definitely there in the code)?

Did everything yarn install successfully?

jijichen:
> This looks like some sort of weird babel error 🤔 it looks like whatever's compiling things has stripped out the parentheses from that return statement (definitely there in the code)?
> 
> Did everything yarn install successfully?

Thanks for taking a look @tmyracle !

Yea i had to use `yarn install --ignore-optional ` due to #44 , but it installed successfully!

Although I see 

```
error - ./pages/_app.tsx
Module parse failed: Unexpected token (51:11)
File was processed with these loaders:
 * ../../node_modules/next/dist/compiled/@next/react-refresh-utils/dist/loader.js
 * ../../node_modules/next/dist/build/babel/loader/index.js
You may need an additional loader to handle the result of these loaders.
|
|   if (session && status === 'authenticated') {
>     return <OnboardingGuard>
|                 <UserAccountContextProvider>
|                     <AccountContextProvider>
info  - Downloading WASM swc package...
info  - Using wasm build of next-swc
[...]

Issue #168, opened 2024-01-19
Implement Storybook for client code


simonbukin:
Currently, we have a Storybook set up in `/libs/design-system` for individual design system components, but none for the components used in the client code (`apps/client`). Having this set up would allow us to quickly test our changes to components (such as Onboarding) without having to hunt down the page in a running version of the app.

**Describe the solution you'd like**
Ideally, we'd have a command like `yarn nx run client:storybook`, which will spin up a Storybook for client components. I think that I can adapt some of existing Storybook code for the design system and point it at the client components, but I'll need to test that out.

This will also need to be added to the docs (ideally for frontend contributions) since this will allow us to quickly test out component changes. The Storybook functionality looks like it is undocumented at the moment.

karan0handa:
@Shpigford could this be opened again? work still remains #204 and adding more stories.

tmyracle:
Another cool next step could be to get the storybook theme updated to the Maybe branding, hopefully that's relatively easy since we can lift from the design library setup.

Shpigford:
Closing again as I believe #204 addressed things.

simonbukin:
@tmyracle I can take a quick look at this; it may be as simple as just just yoinking the design library definition but I'll have to make sure babel plays nice.

PR #204, opened 2024-01-22
Fix .babelrc issue and readd Storybook support in client

simonbukin:
This PR fixes readds Storybook support to `client` components, fixing the issue that was preventing a successful build in (#201). It also refixes #168. This actually ended up making the `.babelrc.json` a lot simpler, so that's positive :).

There had to be a small change to generic component, `TrendBadge.tsx`, due to a naming conflict (since both the type and variable were named `BadgeVariant`, the compiler was complaining). I aliased the type in the file and error was fixed, but I'm open to other solutions to this problem if anyone knows a better one.

I've included a screenshot here of both the Storybook and client running at the same time, from the same config. I tested the functionality of both on a fresh install, and had no issues.

Screenshot 2024-01-21 at 8 42 26 PM


karan0handa:
After this is done, could you create an issue that lists the next steps for completing client storybook integration? For instance - 
1. Supporting tailwind in storybook
2. Adding client components
3. etc.

PR #206, opened 2024-01-22
Made login/register buttons full width

BartoszJarocki:
Changed the CTA for login and register pages style to be full width as it looks a little bit cleaner. The text under the CTA is centered now, too. 

Before
CleanShot 2024-01-22 at 21 40 41@2x
CleanShot 2024-01-22 at 21 41 00@2x

After
CleanShot 2024-01-22 at 21 35 41@2x
CleanShot 2024-01-22 at 21 36 42@2x

### DIFF
--- a/apps/client/pages/register.tsx
+++ b/apps/client/pages/register.tsx
@@ -110,9 +110,8 @@ export default function RegisterPage() {
                                 </div>
                             ) : null}
 
-                            <AuthDevTools isAdmin={isAdmin} setIsAdmin={setIsAdmin} />
-
                             <Button
+                                className="w-full"
                                 type="submit"
                                 disabled={!isValid}
                                 variant={isValid ? 'primary' : 'secondary'}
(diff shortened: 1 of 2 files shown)


apps/client/pages/register.tsx
```diff
@@ -110,17 +110,16 @@ export default function RegisterPage() {
                                 </div>
                             ) : null}
 
-                            <AuthDevTools isAdmin={isAdmin} setIsAdmin={setIsAdmin} />
```
tmyracle, line 113: why was this removed?

Shpigford: danngggittt. i totally missed that when i looked at the code commits. can't speak to why it was removed, but 100% on me for not catching.

BartoszJarocki: my bad! PR with restoring it coming

tmyracle: all good, we can also use the fullWidth prop on the button rather than adding the class 👍

BartoszJarocki: follow up PR - https://github.com/maybe-finance/maybe/pull/208

sorry folks! regarding full width prop, good point. I find it a little bit weird to expose it that way but that's a topic for another time :)

PR #210, opened 2024-01-22
Add theme to client Storybook

simonbukin:
This PR adds company theming to the client Storybook. I mostly copied over and fixed some assets from the design system Storybook, but we can also think about centralizing assets in a top level folder and accessing that instead. This change doesn't cause any issues with launching the main app (double checked this time 😉)

Also, by including a css file with Tailwind directives, Tailwind now works in the Storybook (@karan0handa)


simonbukin:
Forgot to add a screenshot:

Screenshot 2024-01-22 at 3 48 46 PM

apps/client/assets/logo.svg
karan0handa, line 1: This logo already exists in `@maybe-finance/design-system`. Is this needed?

apps/client/assets/styles.css
karan0handa, line 1: Is this required since `apps/client/styles.css` already exists?

karan0handa, line 1: Is this required since `apps/client/styles.css` already exists?

karan0handa (commented):
Apart from the things mentioned, I think this is good for now. In a future PR, however, storybook should be switched over to Typescript so that import aliases can be used. 

https://storybook.js.org/docs/configure/typescript

apps/client/assets/logo.svg
simonbukin: I tried using a relative import, but ran into the following error:

```
Projects cannot be imported by a relative or absolute path, and must begin with a npm scope (@nrwl/nx/enforce-module-boundaries)
```

I'm thinking that import aliases would help here, but the temporary solution was duplicating the `logo.svg`. Do you think there's a better way forward here?

apps/client/assets/styles.css
simonbukin: Good catch. I removed this file and referenced the existing `styles.css` and it looks like it works just fine 👍

apps/client/assets/logo.svg
karan0handa: Oh I see. This happens during the pre-commit workflow, which is why I didn't see the error.

I think the way forward is to migrate Storybook to typescript and then it'll be able to use import aliases.

If the migration is too involved and better suited for a future PR the local `logo.svg` should be removed as the theme just concerns the Storybook UI and not the components themselves.

I definitely don't think we should have the same logo in two places though.

apps/client/.storybook/preview.js
```diff
@@ -1,4 +1,4 @@
-import '../assets/styles.css'
+import '..//styles.css'
```
karan0handa, line 1: Accidental typo? `../styles.css`

(though interestingly through this i discovered that the number of `/` doesn't matter! output remains the same)

simonbukin: Ack! Wild that it still worked 😅 Let me fix that

tmyracle (approved):

PR #215, opened 2024-01-24
feat: manually input and automatically track US stocks

karan0handa:
Solves #131.
- [x] UI for adding a manual investment account — Would use the existing "Investment" block in the modal, where you'd either pick an existing account or add a new one.
- [x] UI for adding the stock — Should be a "search to select" UI for pulling from stock listing (Database: `Security`)
- [x] Creating the appropriate "transaction" for adding the stock.
- [ ] UI for removing/selling the stock
- [ ] Create the appropriate "transaction" for removing the stock.

/claim #131


libs/client/features/src/accounts-manager/stock/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './AddStock'
```
gamedevsam, line 1: Ahh, good old index files. I'm not opposed to this, but find them to be a waste of time. VSCode can auto-import files, so why add this at all?

gamedevsam: Works for me, just pointing out it's a waste of time when working with modern JS / TS (aka with ES Modules). Just use named exports when possible, and let VSCode deal with imports for you.

PR #218, opened 2024-01-24
Migrate to pnpm

### LINKED ISSUE #87: Migrate from Yarn 1.x to pnpm
**Is your feature request related to a problem? Please describe.**
Yes. This repo is currently using an outdated package manager.

**Describe the solution you'd like**
Yarn Classic (1.x) is ~4ish years old and needs an upgrade. The purpose of this issue is to upgrade the workspace to explore using the pnpm package manager rather than Yarn 1.x

**Describe alternatives you've considered**
As of now, it's best to use Yarn 1.x. We can configure `package.json` to explicitly use the classic version of Yarn and add some documentation as to how developers can configure Yarn to use 1.x.

**Additional context**
[pnpm home page](https://pnpm.io/)

DestroyerXyz:
Closes #87.

Migrated over the lockfile and scripts.

Haven't tested all scripts yet, some help testing would be appreciated.


.dockerignore
```diff
@@ -1,5 +1,4 @@
-# Ignores everything except dist/ prisma/ apps/ and yarn.lock
+# Ignores everything except dist/ prisma/ and apps/
 *
 !dist/
-!prisma/
-!yarn.lock
```
tmyracle, line 5: do we not need the pnpm-lock.yaml file in docker? just curious why it wasn't replaced, I'm not as familiar with pnpm aside from basic use

BoscoDomingo: Yes, the lockfile is extremely recommended and sometimes even necessary (eg `pnpm install --frozen-lockfile`)

tmyracle:
Have you tried this with removing node_modules and doing a fresh install? When I pulled this branch it looked like everything worked fine, but when I removed node_modules and did a fresh install I got a ton of import errors. Might be due to lack of dependency hoisting compared to yarn? 

CleanShot 2024-01-24 at 10 27 36

apps/workers/Dockerfile
```diff
@@ -6,8 +6,8 @@
 FROM node:16.18.1 as builder
 
 WORKDIR /app
-COPY ./dist/apps/workers ./yarn.lock ./prisma ./
-RUN yarn install --production
+COPY ./dist/apps/workers ./pnpm-lock.yaml ./prisma ./
+RUN pnpm install --production
```
BoscoDomingo, line 10: `--production` is not accepted. It has to be `--prod` or `-P`. `--frozen-lockfile` is also highly recommended

BoscoDomingo (changes requested):
There's a couple things to change. I haven't had a chance to test it in my machine yet either

.dockerignore
DestroyerXyz: so @BoscoDomingo you're saying I should not add it right?

tmyracle:
> > Have you tried this with removing node_modules and doing a fresh install? When I pulled this branch it looked like everything worked fine, but when I removed node_modules and did a fresh install I got a ton of import errors. Might be due to lack of dependency hoisting compared to yarn?
> > CleanShot 2024-01-24 at 10 27 36
> 
> This was fixed by running `pnpm run prisma:gen` before trying to build.

Fixed the majority of the issues, but still seeing this. Any tips?

CleanShot 2024-01-24 at 10 59 09

DestroyerXyz:
Can't see how any of my changes would have caused this.
The `@bull-board/api` and `@sentry/types` packages weren't installed in the first place.

DestroyerXyz:
could it be that using yarn lets you access peerdeps in your code?
feels like a bad idea but that's the only explanation I can think of.

tmyracle:
> could it be that using yarn lets you access peerdeps in your code? feels like a bad idea but that's the only explanation I can think of.

Yeah after doing some reading it looks like pnpm requires explicit dependencies whereas yarn did not? So we might need to add those two and make sure the versions work with the related package. Possibly related discussion here: https://github.com/pnpm/pnpm/issues/4920

josephgoksu (approved):

tmyracle:
Ok I was finally able to get this working. From a totally clean install:

- Add "@sentry/types": "^7.22.0", to devDeps
- Add "@bull-board/api": "^4.6.4", to deps
- Fix the 13 or so type errors that show up in the console, I think they're all involving expecting a string but the date function could return string or null. Not sure why these weren't being surfaced before 🤷

gamedevsam:
Now this change I support. I <3 pnpm.

DestroyerXyz:
This should be working now. Can you guys test it out?

tmyracle:
> This should be working now. Can you guys test it out?

Everything seems to be working on my end 👍 
I feel like I want to add a preemptive help question in the Discord to tell people to nuke node_modules, run prisma:migrate:dev if they've run the app before. I still think the preinstall script isn't a bad idea to force pnpm use and prevent accidental yarn installs, but I guess we can add that if it becomes a common issue 🤷‍♂️ . Thanks for taking this on!

BoscoDomingo (approved):

tmyracle (approved):

DestroyerXyz:
@tmyracle can you merge this?

tmyracle:
> @tmyracle can you merge this?

Quick fix of the conflicts and I'll merge, sorry about that.

DestroyerXyz:
@tmyracle done

PR #222, opened 2024-01-24
Issue 81

Shivam0710:
I have done changes as follows:

    Moved location question before birthday so we know what country the user and based on the country of I've maintained a function where we can add the value based on the country code.
    Added fields in geo-utils for dateFormat so that we can pick the dateFormat based on country code.
    Added a prop to DatePicker component that takes in the date format and parses the date accordingly and sets an error message if the existing if (inputData.isValid) check fails.
    Removed hard coded formats in DatePicker component and made them dynamic based on the dateFormat provided.


Shivam0710:
@tmyracle should I store country id in cookies and pick it from cookies while showing other dates? Or are we storing user country code in db from which I can query and use this value?

tmyracle:
> @tmyracle should I store country id in cookies and pick it from cookies while showing other dates? Or are we storing user country code in db from which I can query and use this value?

It's stored in the db in the user table, let's use that as the source of truth. There's a useProfile hook in useUserApi that should return that.

libs/client/features/src/onboarding/steps/Profile.tsx
```diff
@@ -40,6 +40,15 @@ type FormValues = {
     maybeGoalsDescription?: string
 }
 
+function getDateFormatByCountryCode(countryCode: string): string {
+    const country = Geo.countries.filter((country) => country.code === countryCode)
+    if (country.length > 0) {
+        return country[0].dateFormat
+    }
```
pieterbeulque, line 47: Might as well do a `.find()`

PR #224, opened 2024-01-24
Fix: Teller cc balance

tmyracle:
Fixes Teller cc balance 🤦


MichaelDeBoey:
@tmyracle Didn't you fix this in #205 because there were some problems with Teller credit accounts?
If it's now reverted, could you explain why it was reverted and/or explain why your first thoughts from #205 were wrong?

Just trying to understand the rationale here

tmyracle:
> @tmyracle Didn't you fix this in #205 because there were some problems with Teller credit accounts?
> 
> If it's now reverted, could you explain why it was reverted and/or explain why your first thoughts from #205 were wrong?
> 
> 
> 
> Just trying to understand the rationale here

the Teller sandbox data for cc accounts always returns 0 for ledger balance for some reason, i should have tested with live data to be able to sanity check it better

Issue #223, opened 2024-01-24
Add stories for client code


simonbukin:
We currently have Storybook integration set up for client code (`/apps/client`). However, to make this useful, we need to start implementing stories for the components/pages for that part of the app. This issue will track that work.

# `components`
- [x] APM.tsx
- [x] Maintenance.tsx
- [x] Meta.tsx 
# `components/account-views`
- [ ] DefaultView.tsx
- [ ] InvestmentView.tsx
- [ ] LoanView.tsx
# `pages`
- [x] 404.tsx
- [ ] data-editor.tsx
- [ ] index.tsx
- [ ] login.tsx
- [ ] oath.tsx
- [ ] onboarding.tsx
- [ ] register.tsx
- [ ] settings.tsx
- [ ] upgrade.tsx
# `pages/plans`
- [ ] [planId].tsx
- [ ] create.tsx
- [ ] index.tsx
# `pages/card`
- [ ] [id].tsx
# `pages/accounts`
- [ ] [accountId].tsx
- [ ] index.tsx

# Miscelleanous
- [ ] Migrate to TypeScript for Storybook files, which will allow us to use import aliases for `assets`
- [ ] Upgrade to Storybook 7. This is mostly an automatic process, but we need to spot check the design system and client Storybook to make sure nothing is impacted by the change 

The above is just a listing of all the pages/components found in `apps/client`. We can whittle down this list or add more, depending on our needs and what is feasible to encode into a story or not.

anoopw3bdev:
@simonbukin Are contributions welcome for this issue? I would like to contribute

simonbukin:
Contributions are welcome @anoopw3bdev! Let me know if you need any help with anything

anoopw3bdev:
Cool @simonbukin I saw this draft PR https://github.com/maybe-finance/maybe/pull/229, should the new stories be written in top of this PR?

simonbukin:
No, generally you would want to open a new PR for stories you’re adding. A good example of this is #226

nikwhoa:
@simonbukin Are contributions welcome for this issue? I would like to contribute

simonbukin:
@nikwhoa go for it! Contributions are welcome, let me know if you have any questions.

tmyracle:
Meta and APM don't really warrant stories so marking those as done for now.

SamuelFanawopo:
[...]

PR #226, opened 2024-01-25
Add Storybook Page for 404 Error

SamuelFanawopo:
## Description

This PR adds a Storybook page for the 404 Error component. The addition of this Storybook page is part of the ongoing effort to improve documentation and component testing within the project.

## Changes

- Created a new Storybook page for the 404 Error component.
- The Storybook setup includes the default state of the 404 page for visual and interaction testing.

The implementation follows the standard structure for our Storybook pages, ensuring consistency across our documentation. This setup will help in visually testing the 404 Error component in isolation and will aid in future enhancements and bug fixes.

## Visuals

Attached is an image of the final Storybook implementation for the 404 Error page, showcasing how it appears in the Storybook environment.

Storybook Page for 404 Error

## Issue Fix

This PR addresses a part of Issue #223 - "Add stories for client code." Specifically, it resolves the task of creating a Storybook page for the 404 Error component, marking one out of the 21 components outlined in the issue.

## How to Test

1. Checkout to this branch.
2. Run the Storybook server using the command `yarn nx run client:storybook`.
3. Once Storybook is running, navigate to `pages/Page404` in the Storybook UI.
4. Review the rendered 404 Error page

### DIFF
--- a/apps/client/pages/404.stories.tsx
+++ b/apps/client/pages/404.stories.tsx
@@ -7,6 +7,4 @@ export default {
     component: Page404,
 } as Meta
 
-const Template: Story = () => <Page404 />
-
-export const Default = Template.bind({})
+export const Default: Story = () => <Page404 />;
--- a/apps/client/pages/404.stories.tsx
+++ b/apps/client/pages/404.stories.tsx
@@ -1,0 +1,12 @@
+import React from 'react'
+import type { Story, Meta } from '@storybook/react'
+import Page404 from './404'
+
+export default {
+    title: 'pages/Page404',
+    component: Page404,
+} as Meta
+
+const Template: Story = () => <Page404 />
+
+export const Default = Template.bind({})


apps/client/pages/404.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import type { Story, Meta } from '@storybook/react'
+import Page404 from './404'
+
+export default {
+    title: 'pages/Page404',
+    component: Page404,
+} as Meta
+
+const Template: Story = () => <Page404 />
```
tmyracle, line 10: does doing a named export here work as well? I think we can just do export const .... and skip the next line
still learning Storybook so apologies if this is a dumb question

SamuelFanawopo: > does doing a named export here work as well? I think we can just do export const .... and skip the next line 

Yes that's a valid approach, and is more suitable in the context of this page. The `bind({})` part is used to create a new instance of the function with its own context, but it’s more relevant in scenarios where you need to create multiple stories from the same template with different properties. 

For a simple single story like this for the 404 page directly exporting as a named export is fine. 

Instead of: 
```jsx
const Template: Story = () => <Page404 />;
export const Default = Template.bind({});
``` 
We would instead change it to:
```jsx
export const Default: Story = () => <Page404 />;
```

Thanks for the suggestion.

tmyracle: Interesting, learned something new!

tmyracle (approved):
lgtm!

PR #228, opened 2024-01-25
feat: add `/securities` route to get all securities to display to the user 

karan0handa:
This is in preparation to solve #131 and the first step of #215.
- Adds a `/securities` endpoint to get all securities (only symbol and exchangeName) from the database.


gamedevsam (approved):
A TypeScript repo without semi-colons? Who's in charge here? Lol.

karan0handa:
> A TypeScript repo without semi-colons? Who's in charge here? Lol.

https://chat.openai.com/share/dffc98d6-e9ea-4513-8914-d431c2a3cae6

gamedevsam:
> > A TypeScript repo without semi-colons? Who's in charge here? Lol.
> 
> https://chat.openai.com/share/dffc98d6-e9ea-4513-8914-d431c2a3cae6

Oof, got me good there, I apologize for my tone. Let me try again...

Interesting choice to omit semi-colons in this TypeScript repository. Could someone share the reasoning behind this decision?

xD

Milehigh-wrld (approved):

apps/server/src/app/routes/securities.router.ts
```diff
@@ -4,6 +4,22 @@ import endpoint from '../lib/endpoint'
 
 const router = Router()
 
+router.get(
```
tmyracle, line 7: Is this a placeholder or full implementation? Just curious on if you're planning to get all and filter client side or add some query params for searching (similar to institutions). I'm guessing this is setting up for the add investment form?

karan0handa: Yeah this is for setting up the investment form. I'll have two Listbox elements for picking a stock. One selects the exchange, and then the stocks of that exchange are available in the other Listbox (this happens after selecting / creating an account).

That's why I'll need all stocks, so this is the complete implementation. If I need something else I'll create another PR but it seems that this is all I need for the backend part of #131.

tmyracle: Can we get away with only one Listbox? I wonder how many people know the exchange vs just the ticker or company name. I'm imagining a single listbox and each result displays the relevant info needed to select the right security (name, ticker, exchange). Would be great to be able to search by ticker or name so I can just type Apple or Samsara and get the result I'm looking for.

karan0handa: Yeah that's actually a better idea. Search by symbol and then disambiguate with exchange.

tmyracle: Sending all of the 11k securities (and more in the future) to the client is going to have negative performance implications. I'd check the implementation of institution router and institution service for how server-side search is implemented elsewhere in the app. There should be an initial batch sent to the client (current standard is 50 results, see the enum PageSize in account-types.ts) and the search (debounced) should return the results that match the query using postgres full-text search.

PR #229, opened 2024-01-26
First draft of Storybook 7 + onboarding page

simonbukin:
This PR is a first draft of using Storybook to test out the onboarding flow. This required more work than expected.

First, upgrading to Storybook 7. This was automatic, using mostly `npx storybook@latest upgrade`. From first glance over the existing Storybook components in the client code, no stories are broken, but this requires more testing. 

Refactor of the `OnboardingPage.tsx` component. This is due to a build error that happens when exporting the component as default. For some reason, exporting the component as default explicitly at the bottom of the page, not inline, solves this error. This may be another issue that could be solved by babel, but this was my workaround for now.

Outstanding work right now is getting the onboarding page to load properly, as it currently is stuck loading. But it does render!

Screenshot 2024-01-26 at 12 03 32 PM


SamuelFanawopo:
> Outstanding work right now is getting the onboarding page to load properly, as it currently is stuck loading. But it does render!

So I've done some research into a similar process, I was developing stories for the login and registration pages. 

Storybook is ideally suited for visually testing components in isolation, and its environment is best leveraged with static, unchanging data. Relying on dynamic state, as is currently the case with the onboarding page, can lead to issues like the one you're experiencing.

To navigate around this, a more effective method would be to deconstruct the entire onboarding flow into smaller, manageable components. By creating individual stories for each step of the onboarding process you can concentrate on the UI and functionality of each step in isolation, without the overhead of managing the entire flow.

Moreover, while testing individual components has its benefits, simulating the entire flow can provide additional insights. For this purpose, you can craft a story that replicates the onboarding flow using static data. This involves creating a wrapper component in Storybook that controls the state of the onboarding flow, updating it based on user interactions within the story.

An essential part of this strategy is to replace any dynamic data fetching or asynchronous operations with mock data. This shift is crucial to prevent the component from being stuck in an endless loading state, waiting for data from external sources. Such an approach not only aligns with Storybook's strengths but also offers a more controlled and efficient environment for UI testing.

Implementing this strategy should help in effectively resolving the loading issue and provide a more streamlined and focused testing environment within Storybook.

Hope this helps.

simonbukin:
This makes a lot of sense @SamuelFanawopo, thanks for researching this! I think creating stories per component is a good way forward, and we can set up the "integration" tests for entire flows (like onboarding) as a step after that. I'll see if I can the onboarding steps into stories for this PR.

apps/client/pages/onboarding.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import type { Story, Meta } from '@storybook/react'
```
tmyracle, line 1: Could you move this to the new /stories directory? Having the page stories in client/pages does some wonky stuff with the build process since it's expecting everything there to be a route and page.

PR #231, opened 2024-01-27
fix: add indextabs scrollbar

karan0handa:
### Description
I couldn't scroll through the index bars and some of them were being hidden because of it. This can be seen in the below video. This PR fixes that.

### Video
https://github.com/maybe-finance/maybe/assets/79533543/e6721eb4-5978-4558-8d2b-c5544ff07f4b

### Before
image

### After
image

### DIFF
--- a/apps/client/styles.css
+++ b/apps/client/styles.css
@@ -132,6 +132,31 @@
             overflow: auto !important;
         }
     }
+
+    .show-on-hover-custom-gray-scroll::-webkit-scrollbar {
+        height: 14px;
+        width: 14px;
+    }
+
+    .show-on-hover-custom-gray-scroll::-webkit-scrollbar-thumb {
+        border: 4px solid rgba(0, 0, 0, 0);
+        background-clip: content-box;
+        @apply bg-transparent rounded-full;
+    }
+
+    .show-on-hover-custom-gray-scroll {
+        overflow: overlay !important;
+    }
+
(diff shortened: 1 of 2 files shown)


Shpigford:
Does this add an always-present scrollbar?

karan0handa:
> Does this add an always-present scrollbar?

Yup

karan0handa:
Something like this


https://github.com/maybe-finance/maybe/assets/79533543/ffa5879f-e4c1-4b39-a1fa-5eddc04fe0b0

Shpigford:
So, visually speaking, I'm not a big fan of that. Also, I'm able to side scroll without issue (Chrome & Safari).

https://github.com/maybe-finance/maybe/assets/35243/3b07d6b5-d2c8-4e2b-ae93-dabfc7ee667f

karan0handa:
> So, visually speaking, I'm not a big fan of that. Also, I'm able to side scroll without issue (Chrome & Safari).
> 
>  CleanShot.2024-01-27.at.08.22.36.mp4

That's interesting. I'm on ubuntu and I'm not able to side scroll without pressing "Shift". This seems to be a common issue on Ubuntu and probably other linux distros.

I've added a scrollbar that only appears when you hover over the `IndexTabs`.


https://github.com/maybe-finance/maybe/assets/79533543/06c81c55-da30-4098-a455-e1893436eb25

PR #247, opened 2024-02-02
Add basic devcontainer

robzolkos:
Makes it easier for developers to open the project and have all of its dependencies work in a consistently.

All that is required is a docker development environment (Docker + VS Code + DevContainers extension is the most common).


robzolkos:
Waiting on https://github.com/maybe-finance/maybe/pull/248 and a green CI to move forward.

robzolkos:
@Shpigford once this is merged, let me know if your local dev environment still works (assume you don't use a devcontainer).  If it does lets work out what the differences are to make it work and then I'll update the readme for both kinds of workflow (non-docker and docker).

Shpigford:
@robzolkos Merged. Error on `bin/dev`:

```
/Users/joshpigford/.rbenv/versions/3.3.0/lib/ruby/site_ruby/3.3.0/bundler/definition.rb:570:in `materialize': Could not find brakeman-6.1.2 in locally installed gems (Bundler::GemNotFound)
```

robzolkos:
`bundle` (as there was gems updated in previous PRs merged)

Shpigford:
@robzolkos Merged and successfully running `bin/dev`.

I've got zero experience with devcontainers, so if there's some other workflow to use at this point, let me know. 🙂

robzolkos:
All good.  If it works without devcontainers you can continue on as normal,  I'll add a piece into the readme for collaborators that prefer to use devcontainers.

JoshAntBrown:
Interestingly, my setup uses brew to install and manage Postgres and for some reason doesn't have the default superuser `postgres`. So the change to the database config to explicitly use this user breaks on my machine.

Happy to add the user to my database, but might be something others run into when trying to get started. Any thoughts @Shpigford @robzolkos?

robzolkos:
Yeah thought this may be the case. Will pr a fix

PR #251, opened 2024-02-02
Clear out broken tests

aviflombaum:
No reason to leave the broken controller tests in there.


Shpigford:
@aviflombaum GitHub is showing merge conflicts. Thoughts on how to resolve that when you're removing files like this?

PR #252, opened 2024-02-02
Fix account param safety

robzolkos:
This gets CI green again.  Not sure what the allowable account types are so I made some up..


app/controllers/accounts_controller.rb
```diff
@@ -33,8 +33,12 @@ def account_params
   end
 
   def account_type_class
-    params[:type].constantize
-  rescue
-    Account # Default to Account if type is not provided or invalid
+    valid_account_types = %w[Checking CreditCard]
```
robzolkos, line 36: These are made up account types.  We should have a list in the model probably.

Shpigford: Account types (for STI purposes) will likely be: Investment, Depository, Credit, Loan, Property, Vehicle, OtherAsset and OtherLiability

Then the `subtype` column will be used to specify the more fine-grained types.

PR #253, opened 2024-02-02
Add Rails LSP plugin

JoshAntBrown:
This PR adds the [ruby-lsp-rails](https://github.com/Shopify/ruby-lsp-rails) gem in development.

It injects a Rack app into development on /ruby_lsp_rails/ so that the ruby-lsp can show attributes and types in your editor, should be useful for helping anyone new to the codebase get to grips with all the things that exist.

I've also updated the Gemfile to read the ruby version from .ruby-version, which reduces some duplication and makes it a little easier to upgrade the version in future.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby "3.3.0"
+ruby File.read(".ruby-version").strip
 
 # Use main development branch of Rails
 gem "rails", github: "rails/rails", branch: "main"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby "3.3.0"
+ruby File.read(".ruby-version").strip
```
robzolkos, line 3: Would `ruby file: ".ruby-version"` work?

JoshAntBrown: It would! Have just updated.

Issue #254, opened 2024-02-02
Rework Accounts model to use Delegated Types


Shpigford:
We currently use STI for Accounts and their types, but given each type can have their own subset of attributes, going the Delegated Types route likely makes more sense.

These are the specific types we need:

* Depository
* Credit
* Investment
* Loan
* Property
* Vehicle
* OtherAsset
* OtherLiability

We'll go in and add the additional type-specific attributes in the future.

The only "STI-specific" bits that likely need addressing:
* The `type` column on Accounts
* Credit model
* Depository model

robzolkos:
What are the attributes common to all types?

(I believe Rails delegated types is the best fit for this problem, especially if there may be specific business logic tied to the various types later)

Shpigford:
> What are the attributes common to all types?

Currently...
* name
* balance

robzolkos:
Began on this [here](https://github.com/maybe-finance/maybe/pull/255)

PR #255, opened 2024-02-02
Rework Account to use delegated types

robzolkos:
Closes https://github.com/maybe-finance/maybe/issues/254

This adds a delegated type for each type of bank account.  It renders a custom new partial depending on the type of bank account selected.

CleanShot 2024-02-02 at 18 19 19@2x

- [x] Custom "new" forms per account type


app/models/account.rb
```diff
@@ -3,5 +3,7 @@ class Account < ApplicationRecord
 
   belongs_to :family
 
+  delegated_type :accountable, types: %w[ Credit Depository Investment Loan OtherAsset OtherLiability Property Vehicle]
```
christianrojas, line 6:
```suggestion
  delegated_type :accountable, types: %w[Credit Depository Investment Loan OtherAsset OtherLiability Property Vehicle]
```

app/models/account.rb
```diff
@@ -3,5 +3,7 @@ class Account < ApplicationRecord
 
   belongs_to :family
 
+  delegated_type :accountable, types: %w[Credit Depository Investment Loan OtherAsset OtherLiability Property Vehicle]
```
JoshAntBrown, line 6: Looks like we could reduce some duplication here?

```suggestion
  delegated_type :accountable, types: VALID_ACCOUNT_TYPES
```

robzolkos: Yep, will push shortly.  Just in draft mode as I work through it.

robzolkos:
@Shpigford have a play with this branch if you have a sec.  Let me know if this is the right track.  We can put customized "new" forms per account type if we want too (or keep it simple and create them with name and balance and then peeps can customize stuff in custom edit forms.  or whatever.

config/tailwind.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   content: [
     './public/*.html',
     './app/helpers/**/*.rb',
+    './app/models/**/*.rb',
```
robzolkos, line 7: This is temporary - we shall put colors in tailwind config.

Shpigford:
@robzolkos Nice! Each account type will definitely have customized "new" forms.

robzolkos:
@Shpigford ok ready for review.   Btw this nukes the Credit and Depository stuff you added in the last hour or so (sti) in favor of delegated types.

Shpigford:
@robzolkos Can you resolve the schema conflict? Came up because #262 added a new table. Then I'll kick the tires on this.

Shpigford:
Can we drop the Accounts `type` column?

robzolkos:
Yes I think so.  We don't have to do it in this PR but if its no longer needing to be used we can do it now or an upcoming one.

robzolkos:
We also need to get 4 icons for the 4 account types that we didn't have ones for...not sure which icons were being used?

Shpigford:
> Yes I think so. We don't have to do it in this PR but if its no longer needing to be used we can do it now or an upcoming one.

Let's go ahead and drop it just to get it cleared out since we aren't doing STI anymore.

> We also need to get 4 icons for the 4 account types that we didn't have ones for...not sure which icons were being used?

I'll tackle that after this gets merged. We've got the icons in Figma for it.

Shpigford:
@robzolkos Merged. Thanks a ton for tackling this!

Issue #259, opened 2024-02-02
Create basic early access system


Shpigford:
In preparation for having the hosted version of Maybe up and running from the beginning, we'd like to create a very rudimentary early access system.

Proposed scope/solution (but up for discussion)...

1. Invite codes model for storing a randomly generated code
2. Codes generated via command line (so that Maybe team members are the only ones that can generate) -- Could be a simple rake task that spits out a single invite code.
3. To keep it clean, should probably store the user_id on the invite code model to detect if a code has been used or not
4. Need to add a field to /registrations/new to input the code
5. Only necessary in the Maybe hosted environment, so could add a flag as an ENV variable

![CleanShot 2024-02-02 at 16 35 01@2x](https://github.com/maybe-finance/maybe/assets/35243/265a7ba9-09f3-4b45-bdd0-513e2b745a7d)

## Bounty
To work on this and receive the $100 bounty, you should follow the instructions around `/attempt` as well as creating a draft PR.

akimthedream:
The project currently employs UUIDs as primary keys for each model. Is there a specific rationale or vision driving this choice? An easy approach to ensure unique invite codes could involve using a bigint as the ID for the InviteCode model. In this scenario, a gem like https://github.com/sqids/sqids-ruby could be used to represent the model's ID as a concise, six-letter invite code 🙂

Shpigford:
No reason to add a gem just for this Just need a simple randomly generated string we can give to a potential user so they can register.

josefarias:
/attempt #259

A couple questions:
* Do we need invitation codes to be tied to a particular email/user? If not then we can keep the table from growing indiscriminately by deleting the code after it's used. The premise being that codes shouldn't be obtainable by more than one person.
* Do we want the codes to expire?

I can implement the solution as proposed

Shpigford:
@josefarias Great questions...

1. Invite codes do *not* need to be tied to a particular user. Good call re: deleting code after user.
2. No need for codes to expire.

mikeholford:
/attempt #259

https://github.com/maybe-finance/maybe/pull/263

Shpigford:
Amazon job on this, @josefarias!

Really appreciate how thorough you were, all the way down to writing tests for it!

Shpigford:
@jkostolansky and @mikeholford Thank you so much for the effort on this! Went with #262, in large part because of how thorough they were all the way down to the tests.

PR #260, opened 2024-02-02
Feature: Add basic invite codes

tmyracle:
Adds a basic invite code model, associates with user when signed up if provided. Sets ENV for whether Maybe hosted or not to display the invite code field. 

Haven't done rails in a while so would appreciate any tips/pointers

/claim #259


Shpigford:
First look is that it seems it doesn't actually require the invite code?

app/controllers/registrations_controller.rb
```diff
@@ -6,13 +6,14 @@ def new
   end
 
   def create
-    @user = User.new(user_params)
+    @user = User.new(user_params.except(:invite_code))
```
viktorianer, line 9: I think you can remove this line, as it does nothing what AR will not do anyway. AR will ignore this param by default.

viktorianer:
No tests at all? 🤔

PR #262, opened 2024-02-02
Implement invite codes

josefarias:
/claim #259

Closes https://github.com/maybe-finance/maybe/issues/259

### DIFF
--- a/app/views/layouts/auth.html.erb
+++ b/app/views/layouts/auth.html.erb
@@ -6,7 +6,7 @@
     <meta name="apple-mobile-web-app-capable" content="yes">
     <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
     <meta name="apple-mobile-web-app-title" content="Maybe">
-    
+
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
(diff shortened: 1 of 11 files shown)


Shpigford:
@josefarias Looking good! Is `rails c` then `InviteCode.generate!` the way to generate codes?

josefarias:
> @josefarias Looking good! Is `rails c` then `InviteCode.generate!` the way to generate codes?

@Shpigford correct!

app/views/layouts/auth.html.erb
```diff
@@ -18,15 +18,21 @@
   </head>
 
   <body class="h-full">
+    <% flash.each do |type, msg| %>
+      <div>
+        <%= msg %>
+      </div>
+    <% end %>
```
josefarias, line 25: Should probably style these. Could do that in this PR or separately.

Shpigford:
I wouldn't worry about the message styling. We need to overhaul that anyways.

Can you add a rake task for the invite code (so we don't even have to hop in to console)?

josefarias:
> I wouldn't worry about the message styling. We need to overhaul that anyways.

Sounds good!

> Can you add a rake task for the invite code (so we don't even have to hop in to console)?

Done! `bin/rake invites:create`

Issue #264, opened 2024-02-03
Centralize auth errors/notices/alerts


Shpigford:
This is specific to authorization messages/errors.

We've got messages sprinkled throughout various places in the app and we need to centralize them so they can consistently be styled.

There's this in the auth layout: https://github.com/maybe-finance/maybe/blob/main/app/views/layouts/auth.html.erb#L21-L25

And then form-specific validation errors: https://github.com/maybe-finance/maybe/blob/main/app/views/passwords/edit.html.erb#L4-L8

Auth-specific view directories to look through to find these...
* views/password_resets
* views/passwords
* views/registrations
* views/sessions

Ideally all messaging would be displayed above the first form field on any give form.

![CleanShot 2024-02-02 at 21 04 09@2x](https://github.com/maybe-finance/maybe/assets/35243/c1718bd0-7ec6-404a-a285-20059fb07075)

If you're interested in tackling this, don't overthink the styling itself. We can customize that. The main thing is getting it all in one place and then at least passing in the type of message (success/error/info/etc) and we can do more customized styling based on that.

josefarias:
I can tackle this tonight

jclusso:
What are your thoughts on errors being handled in one of two ways? When they are attached to a field, you put them with the field, but when they are generic, you throw them into a floating message. I've found this to work really well.

<img width="1288" alt="image" src="https://github.com/maybe-finance/maybe/assets/622516/e8cc3e1c-b435-44b7-ac87-6187dc9829c6"> (ignore phone being blank here... it typically would appear like below, but I didn't want my number public)
<img width="1050" alt="image" src="https://github.com/maybe-finance/maybe/assets/622516/06a74531-c9e6-466c-9da1-9832282bd8ce">

Shpigford:
@jclusso Long term I like that, but will be a much bigger project to tackle and we'll likely have a LOT of UI changes in the coming weeks. So just trying to keep them really straightforward for now.

jclusso:
@Shpigford, makes sense! When more things are fleshed out, feel free to tag me and I'd be happy to bring over some stuff I have for doing this nicely.

Issue #256, opened 2024-02-02
Start I18n Internationalization setup


ricsdeol:
Hello, please start using I18n to facilitate the app to be multi-lang.

I could start this weekend if you approve.

Short Guide about this.
https://guides.rubyonrails.org/i18n.html

Shpigford:
Yes! That was on our shortlist. 100% would love that.

Shpigford:
@ricsdeol If you do start on it, please submit a Draft PR associated with this issue so others can track progress. Helps us make sure issue work doesn't get stale.

ricsdeol:
> @ricsdeol If you do start on it, please submit a Draft PR associated with this issue so others can track progress. Helps us make sure issue work doesn't get stale.

Cool, I'm going to start this weekend :smile:

ricsdeol:
Start at #276

Shpigford:
@ricsdeol Does the I18n stuff that @josefarias has in #269 conflict with what you're doing at all?

jclusso:
Might want to consider using [i18n-tasks](https://github.com/glebm/i18n-tasks) paired with Google Translate or DeepL so translations don't slow down development and so you can lint them in the CI for missing ones.

ricsdeol:
> @ricsdeol Does the I18n stuff that @josefarias has in #269 conflict with what you're doing at all?

I solved the conflicts and use i18-tasks, I didn't know it, very interesting.

josefarias:
> I solved the conflicts and use i18-tasks, I didn't know it, very interesting.

Yeah i18n-tasks can sometimes slow you down a bit which isn't ideal.

But i18n can be a pain to manage when the files get super large. And the tasks + tests force you to be disciplined about keeping it tidy.

It's a pick your poison scenario in my mind 😅.

PR #269, opened 2024-02-03
Centralize auth messages

josefarias:
Closes https://github.com/maybe-finance/maybe/issues/264

At first, I thought by "centralize" we meant we wanted all the copy for the messages in a single place for easy editing. So I added rudimentary i18n, as well as [i18n-tasks](https://github.com/glebm/i18n-tasks), which helps keep i18n files tidy. (That's on me for not paying attention, sorry I jumped the gun here)

Then I realized we actually wanted the messages to all appear in a single place for auth-related forms, so I pulled out a helper for that, too.

Let me know if we don't want the i18n part of this. Happy to take it out.

To be clear, there's still plenty of i18n work left to do — this was more meant to organize the auth messages. The rest of the app needs to be configured still. We can tackle that work as part of https://github.com/maybe-finance/maybe/issues/256.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,6 +60,8 @@ group :development, :test do
   gem "dotenv"
 
   gem "letter_opener"
+
+  gem "i18n-tasks"
 end
 
 group :development do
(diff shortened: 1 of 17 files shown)


Shpigford:
@josefarias Couple of conflicts that popped up from a PR merge this morning.

PR #272, opened 2024-02-03
Apply default form styling

JoshAntBrown:
I'm not sure how others will feel about this. It depends on whether you want to keep a strict tailwind utility class approach, or whether the best solution is somewhere between with a mix of regular CSS with tailwind generating utilities and looking after the theme.

I find that a sprinkle of utilities mixed with some component and composition classes can often yield the best results, so I thought I'd throw together an example of what that might look like. It's an approach inspired by some of https://cube.fyi/. I noticed a similar pattern in the Campfire codebase, though they do without tailwind and roll their own utils.

On my own projects I've typically written a plugin for tailwind to expose some of the theme tokens as CSS custom properties, but that depends on postcss which isn't available here because we're using the standalone tailwind cli.

If you're happy with this approach I can go ahead and update all the form fields to use it.

Also, this effectively replaces #271 as I've applied the same padding styles here.


bryanbeshore:
Definitely an improvement IMO. This is how we organize our 'quite large' app as well.

app/assets/stylesheets/components/form-field.css
```diff
@@ -0,0 +1,36 @@
...
+    opacity: 0.5;
+    outline: none;
+    width: 100%;
+
+    &:focus {
+      opacity: 1;
+      @apply ring-0;
+    }
+  }
+}
```
josefarias, line 36: I'd suggest we keep using `@apply` here and avoid nesting.

```css
.fieldset {
  // @apply ... ;
}

.label {
  // @apply ... ;
}

.form-input {
  // @apply ... ;
}
```

josefarias (commented):
Totally get the reasoning here. It undoubtedly looks cleaner.

However, cleaner markup does not always mean it is easier to maintain.

Picking Tailwind as a framework means choosing maintainability over cleanliness. There's an argument to be made whether it's actually more maintainable or not. But if you fall in the "not more maintainable" camp, then why use Tailwind at all?

The choice is ultimately the maintainer's (in this case @Shpigford and team). But if they're picking Tailwind then I'd vote we use it as the framework suggests — with [some room for reusability](https://tailwindcss.com/docs/reusing-styles), but **mostly** not at the CSS level.

I'm using **mostly** as a key word. All this said, I think (and Tailwind thinks, per the link above) that _some_ componentizing at the CSS level is okay:

> While it’s highly recommended that you create proper template partials for more complex components, you can use Tailwind’s @apply directive to extract repeated utility patterns to custom CSS classes when a template partial feels heavy-handed.

I'd argue template partials _are_ heavy-handed for form fields in particular.

So summarize my take here: I don't think we should adopt the style proposed here as a philosophy in the codebase. Rather, I'd suggest we follow Tailwind's suggestions linked above — they know how to get the best mileage out of their framework. But I also do think this PR should be merged, with the changes I recommended in my comment.

JoshAntBrown:
Thanks, all good points. I find the qualities of clean and maintainable can often be very subjective so ultimately comes down to the choice of the maintainer.

If we take the approach I have proposed then I would suggest creating 'components' sparingly, and avoid making them more than a handful of CSS rules (max 80 - 100 lines).

It's a light pattern that, I believe, increases maintainability when you have a specific group of rules that only apply in that context. 

This form field is a good example of that where these rules would not make sense without the associated elements. Together they form a singular unit on the page.

josefarias:
> If we take the approach I have proposed then I would suggest creating 'components' sparingly

Sounds good! We might be saying the same thing then 👍🏻 

I'd advocate for extremely sparse use. So form elements would be an extremely rare exception. If in the future we end up with more than, say, seven or eight components — then that'd be plenty for me.

You might be thinking more like 15-20+. That'd be too much for me but I'm okay with it not being up to me.

I don't think we'll be setting this in stone in this PR. It's just early in the codebase lifecycle and I think we're both advocating for what we'd want it to look like. Up to the maintainers to enforce that!

> This form field is a good example of that where these rules would not make sense without the associated elements.

This is true. I'd still advocate for no nesting. Even if they don't make sense without the associated elements, and they should always be presented in that nested architecture, there's no reason to prematurely limit ourselves from where we can use these classes. There's no maintainability penalty if we only ever have seven or eight components, which goes back to my point above.

robzolkos:
I largely agree with @josefarias on this.   But it can really be painful with elements that repeat through a rails codebase.  Shared partials are easier to handle.  But for forms I would recommend considering the use of custom form builder.  Forms are used _everywhere_ and remembering to add this class or that class is a pain.

I just want to write `<%= f.text_field :name %>` and know that it will get rendered the same everywhere.   And if I want to change the background color I can do it in one place.

There's a tiny bit of initial boiler plate to setup but once done, its quite nice and maintainable.  And a native Rails feature.

```
# /app/lib/form_builders/tailwind_form_builder.rb
module FormBuilders
  class TailwindFormBuilder < ActionView::Helpers::FormBuilder
    def text_field(method, options={})
	  default_style = "mt-1 block w-full rounded-md shadow-sm focus:ring
        focus:ring-indigo-200 focus:ring-opacity-50"


	  super(method, options.merge({class: default_style}))
    end
  end
end
```

More info here https://blog.testdouble.com/posts/2022-12-05-blending-tailwind-with-rails-forms/

josefarias:
@robzolkos I could get behind that. I don't love specifying a form builder every time, but we could maybe extract a `tailwind_form_with` helper that just forwards everything to `form_with` and adds the form builder. This may have been mentioned in the article — I only skimmed through it. Sorry 😅.

Anyway, I'd suggest we wait and get some maintainer input before bikeshedding too much. I'm totally fine to punt on this philosophical decision that's actually kind of hard to capture in code. Save for the formbuilder route, that establishes a pretty clear precedent.

Shpigford:
In my mind, forms are about the only scenario where I think it could make sense to create some "roll-up" classes to make styling them more consistent. 

Especially in an app like this where there are an offensive number of form fields...of which we're taking a pretty "styled" approach to. Consistently applying a dozen+ class names will be a bit of a dumpster fire to maintain.

I don't have strong feelings about form builders. At the end of the day I just want consistency among styling so it's quick for anyone to put together a form.

JoshAntBrown:
Yeah I think on reflection of the suggestions here the FormBuilder approach paired with some helper methods gets us closer to that.

I can put together a solution that utilises FormBuilder and keeps things more in the spirit of Tailwind.

If there are other 'block' or 'component' type things that would benefit from reusability but don't justify a template partial then perhaps we could use the helpers, but that's something that can be solved later.

Shpigford:
> I can put together a solution that utilises FormBuilder and keeps things more in the spirit of Tailwind.

Sounds good to me. If you're keeping this PR for that solution, would you mind switching to draft mode for this?

leemcalilly:
Rather than build components using CSS inheritance, one thing I like to do is to go all in Tailwind utility approach, but then create some simple button components as helpers.

So things that are annoying to repeat in Tailwind can have a DRY partial where the Tailwind classes only have to be set once.

It's not as heavy-handed as something like ViewComponent, but makes Tailwind easier to work with.

```
module ComponentsHelper

  def add_new_button(path, button_text)
    render 'components/add_new_button', path: path, button_text: button_text
  end

  def form_button(form, cancel_path, cancel_text, button_text)
    render 'components/form_button', form: form, cancel_path: cancel_path, cancel_text: cancel_text, button_text: button_text
  end

  # Etc. 
end
```

Then you would render a button in the view like this:

```
  <%= render 'components/form_button', 
      form: form, 
      cancel_text: t('shared.cancel'), 
      cancel_path: reports_path, 
      button_text: t('reports.form.create') %>
```
      
It's also very easy to add unit tests for each component helper.

JoshAntBrown:
I've made some changes that brings this PR more inline with the recommendations from TW, and changes the default form builder to make it simpler to build forms that follow the standard styles.

With this change you'll be able to supply a label option to the field, setting it to true will render the field wrapper and label with the default options, passing a string will allow it to adjust the label text, passing a hash will allow it to set the label options, and passing an array will allow you to pass the args directly to label.

app/helpers/application_form_builder.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  private
+
+  def label_args(options)
+    return options[:label] if options[:label].is_a?(Array)
+    return [ options[:label], { class: "form-field__label" } ] if options[:label].is_a?(String)
+    return [ nil, options[:label] ] if options[:label].is_a?(Hash)
+    [ nil, { class: "form-field__label" } ]
+  end
```
josefarias, line 25: I think this might read easier. 

```suggestion
  def label_args(options)
    case options[:label]
    when Array
      options[:label]
    when String
      [ options[:label], { class: "form-field__label" } ]
    when Hash
      [ nil, options[:label] ]
    else
      [ nil, { class: "form-field__label" } ]
    end
  end
```

Will take a look at the whole PR after work, just spotted this and thought I'd mention for you consideration.

JoshAntBrown: Yeah, this definitely reads easier, thanks!

Shpigford:
I'm going to leave this open for the time being. We've got some updated form styling coming in the next ~48 hours and we'll see if that has any big changes to this one way or the other before merging.

Shpigford:
@JoshAntBrown Just pushed the form changes. Slightly new styling but not a dramatic departure.

Want to take a look and see if your solution here still fits the bill?

JoshAntBrown:
@Shpigford How's this? Should be updated with the latest styles across the entire app.

Shpigford:
> @Shpigford How's this? Should be updated with the latest styles across the entire app.

This certainly _seems_ amazing to me.

Probably worth getting another set of eyes on the code just for a sanity check.

Shpigford:
@JoshAntBrown Can you resolve the conflicts on this and then I'll get it merged?

JoshAntBrown:
All done!

PR #273, opened 2024-02-03
Refactor 'new' action in AccountsController for better readability and efficiency

AfolabiOlaoluwa:
- Used '.presence' instead of '.blank?' for handling both nil and empty string in 'params[:type]'
- Removed redundant condition for 'params[:type].blank?' by using 'type' directly
- Combined 'Account.new' calls into a single line

The distinction between `.blank?` and `.presence` lies in how they handle nil and empty strings.

`.blank?`: This method returns true if the object is nil, empty, or consists only of whitespace characters. However, it does not handle nil directly; instead, it relies on the truthiness of the object.

`.presence`: This method is particularly useful when you want to obtain a non-nil value. It returns nil if the object is blank (empty or consists only of whitespace characters), and it returns the object itself if it is present.

In the context of the action code, using `.presence` allows handling both nil and an empty string in a concise manner. If params[:type] is nil, it will be assigned to type as nil, and if it's an empty string, it will be assigned to type as nil as well. This makes the subsequent check (type.blank? || ...) more straightforward, as it covers both cases.

For more details, refer to the link https://api.rubyonrails.org/classes/Object.html#method-i-presence

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -2,27 +2,18 @@ class AccountsController < ApplicationController
   before_action :authenticate_user!
 
   def new
-    type = params[:type].presence
-
-    if type.blank? || Account.accountable_types.include?("Account::#{type}")
-      @account = Account.new(accountable_type: "Account::#{type}")
-    else
-      head :not_found
-    end
+    @account = build_account
+    head :not_found unless @account
   end
 
-  def show
-  end
+  def show; end
 
   def create
-    @account = Account.new(account_params.merge(family: current_family))
-    @account.accountable = account_params[:accountable_type].constantize.new
+    @account = build_and_associate_account
 
-    if @account.save
-      redirect_to accounts_path, notice: "New account created successfully"
-    else
-      render "new", status: :unprocessable_entity
-    end
+    render "new", status: :unprocessable_entity unless @account.save
+
+    redirect_to accounts_path, notice: "New account created successfully"
   end
 
   private
@@ -31,11 +22,26 @@ class AccountsController < ApplicationController
     params.require(:account).permit(:name, :accountable_type, :balance, :subtype)
   end
 
-  def account_type_class
-    if params[:type].present? && Account.accountable_types.include?(params[:type])
-      params[:type].constantizes
-    else
-      Account # Default to Account if type is not provided or invalid
+  def build_account
+    type = params[:type].presence
+    return Account.new if type.blank? || !Account.accountable_types.include?("Account::#{type}")
+
+    Account.new(accountable_type: "Account::#{type}")
+  end
+
+  def build_and_associate_account
+    current_family&.accounts.build(account_params).tap do |account|
+      account.accountable = build_accountable
     end
+  end
+
+  def build_accountable
+    account_type_class(account_params[:accountable_type]).new if account_params[:accountable_type].present?
+  end
+
+  def account_type_class(type)
+    return Account unless type.present? && ("#{type}").in?(Account.accountable_types)
+
+    type.constantize
   end
 end
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -2,12 +2,10 @@ class AccountsController < ApplicationController
   before_action :authenticate_user!
 
   def new
-    if params[:type].blank? || Account.accountable_types.include?("Account::#{params[:type]}")
-      @account = if params[:type].blank?
-        Account.new
-      else
-        Account.new(accountable_type: "Account::#{params[:type]}")
-      end
+    type = params[:type].presence
+
+    if type.blank? || Account.accountable_types.include?("Account::#{type}")
+      @account = Account.new(accountable_type: "Account::#{type}")
     else
       head :not_found
     end


app/controllers/accounts_controller.rb
```diff
@@ -2,12 +2,10 @@ class AccountsController < ApplicationController
...
-      else
-        Account.new(accountable_type: "Account::#{params[:type]}")
-      end
+    type = params[:type].presence
+
+    if type.blank? || Account.accountable_types.include?("Account::#{type}")
+      @account = Account.new(accountable_type: "Account::#{type}")
     else
       head :not_found
     end
```
josefarias, line 11: Nice thinking! I think we can take it a step further.

Consider the following:

```rb
maybe(dev)> params = {}
maybe(dev)> params[:type] = "Foo"
maybe(dev)> "Account::#{params[:type]}".presence_in(Account.accountable_types).constantize
(irb):23:in `<main>': undefined method `constantize' for nil (NoMethodError)

"Account::#{params[:type]}".presence_in(Account.accountable_types).constantize
                                                                  ^^^^^^^^^^^^
maybe(dev)> params[:type] = "Credit"
maybe(dev)> "Account::#{params[:type]}".presence_in(Account.accountable_types).constantize
=> Account::Credit(id: uuid, created_at: datetime, updated_at: datetime)
```

We could use that to refactor as follows:
```rb
class AccountsController < ApplicationController
  before_action :authenticate_user!

  def new
    @account = Account.new accountable_type: account_type_class
  end

  private

  def account_type_class
    "Account::#{params[:type].presence_in(Account.accountable_types)}".constantize
  rescue NameError
    Account
  end
end
```

We're presumably okay with defaulting to `Account` per https://github.com/maybe-finance/maybe/blob/48ade396ca3a1a9142350ff4e4620bb80e1d2f75/app/controllers/accounts_controller.rb#L40

But if not, we can just raise the `NameError` — fine to throw an exception if we got this far without a valid type param. I'll let @Shpigford confirm.

josefarias: Actually, per https://github.com/maybe-finance/maybe/blob/48ade396ca3a1a9142350ff4e4620bb80e1d2f75/app/models/account.rb#L4

I don't think rescuing and returning `Account` makes sense. I think we should just raise.

robzolkos: We just need to be careful as the new action renders a different partial depending on the presence of a valid type param.  I think I'll a quick test to make that doesn't break with refactors.  (But agree that this could be made cleaner)

AfolabiOlaoluwa: Well, it's a question of where the `.constantize` fails. If it won't fail, then `if..else` statement or a tenary operator is fine.

AfolabiOlaoluwa: Perhaps I need to look at this as a controller of its own rather than refactoring action by action. I also think a possible solution is to use `||` instead of raise even if `.constantize` happens to return a nit and fail.

josefarias: How is the type param obtained? Sorry I haven't looked myself. It shouldn't be open to freetext input. If we get to this point with an invalid type then someone's attempting something shady and it's fine to raise an exception. Arguably, even better to raise and capture in Sentry so we're aware of which attack vectors are being attempted.

robzolkos: Via links rendered when the type parameter is blank

AfolabiOlaoluwa:
I have changes based on what I think is better instead of `rescue NameError`. You might want to check.

josefarias:
> I have changes based on what I think is better instead of `rescue NameError`. You might want to check.

Ah yeah, I don't think we should `rescue NameError` either, per https://github.com/maybe-finance/maybe/pull/273#discussion_r1477075690

Can take a look after work! Thanks for putting this together

app/controllers/accounts_controller.rb
```diff
@@ -2,29 +2,18 @@ class AccountsController < ApplicationController
...
-      @account = if params[:type].blank?
-        Account.new
-      else
-        Account.new(accountable_type: "Account::#{params[:type]}")
-      end
-    else
-      head :not_found
-    end
+    @account = build_account
+    head :not_found unless @account
```
robzolkos, line 6: I would advise against using modifier clauses like this.  Standard Rails is to not use them in controller actions

For example here - a user will at first see that it returns `:not_found`.  Its not until the end of the line they realise its conditional.

I know this is subjective and both ways work and it saves a couple of lines, but for readability and consistency with other controller actions we should defer to the more verbose if...else...

app/controllers/accounts_controller.rb
```diff
@@ -2,29 +2,18 @@ class AccountsController < ApplicationController
...
-    @account = Current.family.accounts.build(account_params)
-    @account.accountable = account_params[:accountable_type].constantize.new
+    @account = build_and_associate_account
 
-    if @account.save
-      redirect_to accounts_path, notice: "New account created successfully"
-    else
-      render "new", status: :unprocessable_entity
-    end
+    render "new", status: :unprocessable_entity unless @account.save
```
robzolkos, line 14: See commend above re modifier clauses.

robzolkos (approved):
This is a nice cleanup.  I left a comment re my views on modifier clauses.

app/controllers/accounts_controller.rb
AfolabiOlaoluwa: I agree it is subjective. And I buy your point regarding consistencies with other controllers, which is rather also subjective in my opinion. 

Perhaps we can decide what things should look like but I don't think that "rhyme" is totally given in a large codebase mostly with `if..else` conditional statements or tennary operators or modifier clause. That uniformity is not given that we'll have only `if..else` statements accross the board. It is not achievable!

On the other hand, if I take it a bit further to avert concerns in your example and pass in some `..else` statement, this is what I think:

```
def new
  @account = build_account

  if @account.blank?
    head :not_found
  end
end
```

```
def create
  @account = build_and_associate_account

  unless @account.save
    render "new", status: :unprocessable_entity
  else
    redirect_to accounts_path, notice: "New account created successfully"
  end
end
```

Afterall, it's Ruby under the hood with dashes of Rails additions. So we don't have to follow Rails standards in such a situation. I really just want to clean these controllers one after the other and set some precedence we follow in this repo. Let me know what you think using the `@account`, but I certainly take preference in what I provided and I'll accept it in any repository since it works.

app/controllers/accounts_controller.rb
AfolabiOlaoluwa: Yes, I responded.

app/controllers/accounts_controller.rb
AfolabiOlaoluwa: Absolutely not seeing the benefits too. Just trying to regularize the uniformity in your opinion.

app/controllers/accounts_controller.rb
robzolkos: The other benefits of the `if...else` style is that it is clearer that the primary function of the action if to save the new account.   By placing the render clause first it reads left to right like the rendering of "new" if the primary function (which it only true in the exceptional case).

Furthermore it is prone to `AbstractController::DoubleRenderError` exceptions (as would be in this particular case if the @account did not save).

josefarias (commented):
So this PR has grown in scope since the last time I commented on it.

The last time I saw it, the refactoring was pretty close to what I would've done. So I thought it'd be good to comment and suggest what I felt was final polish.

Now the scope has grown and it makes multiple stylistic choices I wouldn't make. I don't think expounding on my thinking would be productive here as I don't own the codebase and there's no defined code style yet.

Debated whether I should comment at all but since I was asked for a review I didn't want to leave y'all hanging. It's good work, just too far removed from what I would recommend and I don't want to bikeshed here in the PR. Happy to pontificate elsewhere, though! Hit me up! 😄

I'll let the maintainers take it from here.

Shpigford:
My thoughts are a similar version of what @josefarias mentioned. Specifically, a lot of discourse over a relatively small amount of code change.

At this stage of the product, my general feeling is it's not worth spending this much time/energy on code that very well could completely get ripped out or rewritten over the next few weeks.

The focus at the moment should be primarily on adding new functionality and less on refactoring.

Going to close this for now.

If there are specific bits of this that have a substantial impact on features, then feel free to submit those smaller code changes.

PR #274, opened 2024-02-03
feat: add settings page

### LINKED ISSUE #266: Add user settings page
We currently have no controllers for users/families and as such also have no "settings" page.

I have no strong feelings on the organization of this, other than there being a page for settings. 🙂 

Things this settings page should allow updating...

1. Family name
2. First name
3. Last name
4. Email
5. Password (will also need a "confirmation" field, if these fields are blank then assume no password change)

I'd suggest using form styling found in views/accounts/new.html.erb

Add a "settings" link from the user avatar drop down.

CleanShot 2024-02-02 at 21 38 35@2x

edrickleong:
Closes #266 

First time using Ruby & Rails, so would appreciate any feedback. If this is good to go, I can put it as Ready for Review.

I used the form styles from `views/accounts/new.html.erb`

https://github.com/maybe-finance/maybe/assets/10529706/064e664c-1e50-48b4-920f-41a5c54a1efb

### DIFF
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -1,0 +1,18 @@
+class SettingsController < ApplicationController
+  def edit
+  end
+
+  def update
+    if current_user.update(user_params)
+      redirect_to root_path, notice: "Profile updated successfully."
+    else
+      render :edit, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def user_params
+    params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation)
+  end
+end
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -1,9 +1,11 @@
 class SettingsController < ApplicationController
+  before_action :authenticate_user!
+
   def edit
   end
 
   def update
-    if current_user.update(user_params)
+    if Current.user.update(user_params)
       redirect_to root_path, notice: "Profile updated successfully."
     else
       render :edit, status: :unprocessable_entity
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -13,6 +13,6 @@ class SettingsController < ApplicationController
   private
 
   def user_params
-    params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation)
+    params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation, family_attributes: [:name, :id])
   end
 end
--- a/app/controllers/settings_controller.rb
+++ b/app/controllers/settings_controller.rb
@@ -13,6 +13,7 @@ class SettingsController < ApplicationController
   private
 
   def user_params
-    params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation, family_attributes: [:name, :id])
+    params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation,
+                                 family_attributes: [ :name, :id ])
   end
 end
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -3,6 +3,7 @@
   resource :session
   resource :password_reset
   resource :password
+  resource :settings
```
pmareke, line 6: I'll limit to just the edit and update actions:

`resources : settings, only: %i[edit update]`

app/controllers/settings_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      redirect_to root_path, notice: "Profile updated successfully."
+    else
+      render :edit, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def user_params
+    params.require(:user).permit(:first_name, :last_name, :email, :password, :password_confirmation)
```
pmareke, line 16: I think @Shpigford wants also to update the family name, right?

app/views/layouts/application.html.erb
```diff
@@ -37,6 +37,9 @@
 
             <div class="absolute z-10 hidden w-screen px-2 mt-2 -translate-x-1/2 left-1/2 max-w-min" data-dropdown-target="menu">
               <div class="w-48 px-3 text-sm font-semibold leading-6 text-gray-900 bg-white shadow-lg shrink rounded-xl ring-1 ring-gray-900/5">
+                <%= link_to edit_settings_path, class: 'block p-2 hover:text-gray-600' do %>
```
pmareke, line 40: You can do it in one line:

` <%= link_to, "Settings", edit_settings_path, class: 'block p-2 hover:text-gray-600' %>`

app/views/settings/edit.html.erb
```diff
@@ -0,0 +1,34 @@
+<h1 class="text-3xl font-semibold font-display">Update settings</h1>
+
+<%= form_with model: current_user, url: settings_path, html: { class: "space-y-4" } do |form| %>
```
pmareke, line 3: Same here, I think the idea is to update also the family name.

pmareke (commented):
Great job!

edrickleong:
Thanks for the suggestions @pmareke!

I've updated based on your feedback.

app/controllers/settings_controller.rb
Shpigford: This is correct. Need to add the family name to this form as well.

pmareke (commented):
Thanks to you for the PR!

jclusso:
It's a good practice to require the current password when changing the password.

edrickleong:
@Shpigford should I look into verifying your password before changing the password? I could reuse the  current password reset flow.

Shpigford:
> @Shpigford should I look into verifying the current password before changing the password as part of this PR?
> 
> If so, I could reuse the current password reset flow.

Let's not worry about current password verification at this stage.

app/views/settings/edit.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  <%= form.fields_for :family_attributes do |family_fields| %>
+    <div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+      <label for="name" class="block text-sm font-medium opacity-50 focus-within:opacity-100">Family name</label>
+      <%= family_fields.text_field :name, placeholder: "Family name", value: current_user.family.name, class: "p-0 mt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100" %>
+    </div>
+  <% end %>
+
+
+  <div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+    <label for="first_name" class="block text-sm font-medium opacity-50 focus-within:opacity-100">First name</label>
```
ricsdeol, line 13: you should use Rails Forms Helper for label like:

```suggestion
    <%= form.label :first_name, class: "block text-sm font-medium opacity-50 focus-within:opacity-100" %>
```

ricsdeol (changes requested):

PR #276, opened 2024-02-03
Start I18n Internationalization setup

### LINKED ISSUE #256: Start I18n Internationalization setup

ricsdeol:
### Start Internationalization Setup
Close #256

#### TODOS:

- [x] Models
- [x] Layouts
- [x] Registration
- [x] Sessions
- [x] Passwords 
- [x] Pages
- [x] Account


adrienpoly:
tiny detail but some files are `.yml` some are `.yaml`

app/models/account/credit.rb
```diff
@@ -1,7 +1,3 @@
 class Account::Credit < ApplicationRecord
   include Accountable
-
-  def type_name
```
ricsdeol, line 4: Can use: `credit.model_name.human`

app/views/accounts/new_credit.html.erb
ricsdeol, line 1: isn't used in any action

ricsdeol:
@Shpigford i thinks the initial setup is done.

Shpigford:
@ricsdeol Please resolve conflicts + failing tests.

ricsdeol:
@Shpigford I fixed the missing keys, however the unused test is we can't use, it doesn't recognize when using some translations dynamically like here:

```erb
  <%= link_to new_account_path(type: type.class.name.demodulize), class: "flex flex-col items-center justify-center w-full text-center focus:outline-none" do %>
    <span class="absolute inset-0" aria-hidden="true"></span>
    <span class="flex w-10 h-10 shrink-0 grow-0 items-center justify-center rounded-xl <%= bg_color %> mb-2">
      <%= inline_svg_tag(icon, class: "#{text_color} stroke-current") %>
    </span>
    <%= type.model_name.human %>
  <% end %>
```

or 

```erb
<% Current.family.accounts.each do |account| %>
    <div class="flex items-center justify-between px-3 py-3 mb-2 bg-white shadow-sm rounded-xl">
      <div class="flex items-center text-sm">
        <%= account.name %>
      </div>
      <div class="flex items-center text-sm">
        <%= account.accountable.model_name.human %>
      </div>
      <p class="text-sm text-right">
        <span class="block mb-1"><%= number_to_currency account.balance %></span>
      </p>
    </div>
<% end %>
```

or others 'rails default' like here: 

```yml
en:
  date:
    abbr_day_names:
    - Sun
    - Mon
    - Tue
    - Wed
    - Thu
    - Fri
    - Sat
    abbr_month_names:
    -
    - Jan
    - Feb
    - Mar
    - Apr
    - May
    - Jun
    - Jul
    - Aug
    - Sep
    - Oct
    - Nov
    - Dec
    day_names:
    - Sunday
    - Monday
    - Tuesday
    - Wednesday
    - Thursday
    - Friday
    - Saturday
    formats:
      default: "%Y-%m-%d"
      long: "%B %d, %Y"
      short: "%b %d"
```

which is a good practice to facilitate translation into other languages or add date custom formats.

josefarias:
> I suggest disabling this test for now, then we can enable and check how to configure these cases.

Do the suggestions in https://github.com/glebm/i18n-tasks#dynamic-keys help with this?

ricsdeol:
> > I suggest disabling this test for now, then we can enable and check how to configure these cases.
> 
> Do the suggestions in https://github.com/glebm/i18n-tasks#dynamic-keys help with this?

Doesn't work with active record attributes and name like this:

```yml
---
en:
  activerecord:
    attributes:
      account:
        balance: Balance
        currency: Currency
        family: Family
        family_id: Family
        name: Name
        subtype: Subtype
    models:
      account: Account
      account/credit: Credit Card
      account/depository: Bank Accounts
      account/investiment: Investments
      account/loan: Loan
      account/other_asset: Other Asset
      account/other_liability: Other Liability
      account/property: Real Estate
      account/vehicle: Vehicle
```

josefarias:
> Doesn't work with active record attributes and name like this:

I think we should ignore those in the config instead of skipping the test. If we skip, enabling down the road will be a pain as we'll be missing context for which keys can safely be ignored.

For that same reason, I'd suggest adding a note to the config about why we're ignoring these in particular and how we are, in fact, using them (i.e. via model_name.human).

I see we use the model names in template. So fine to ignore those. Do we use the attribute names as well? (E.g in validations) If not, I'd avoid adding them now. I18n files can get unruly, so would rather only add what we absolutely need.

ricsdeol:
> > Doesn't work with active record attributes and name like this:
> 
> I think we should ignore those in the config instead of skipping the test. If we skip, enabling down the road will be a pain as we'll be missing context for which keys can safely be ignored.
> 
> For that same reason, I'd suggest adding a note to the config about why we're ignoring these in particular and how we are, in fact, using them (i.e. via model_name.human).
> 
> I see we use the model names in template. So fine to ignore those. Do we use the attribute names as well? (E.g in validations) If not, I'd avoid adding them now. I18n files can get unruly, so would rather only add what we absolutely need.

Thanks @josefarias I did your suggestion. 


@Shpigford Please could re-evaluate this changes?

Shpigford:
There's an error but I'm ignoring. Going to merge and then just kill of the system tests as they're getting in the way of our quickly-changing UI.

Shpigford:
Thanks for tackling this, @ricsdeol!

PR #279, opened 2024-02-03
Clean up registration and session views

jclusso:
- There was some lingering devise logic that was hiding some text and links for sign-up and sign-in in the auth layout.
- Added a `text-link` class since I feel this is a commonly used thing that is ideal to have an applied it to all links in these views.

### DIFF
--- a/app/assets/stylesheets/application.tailwind.css
+++ b/app/assets/stylesheets/application.tailwind.css
@@ -34,10 +34,6 @@
   @apply p-0 mt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100;
 }
 
-.text-link {
-  @apply font-medium text-gray-600 hover:text-gray-400 transition;
-}
-
 @layer base {
   @font-face {
     font-family: "GeneralSans";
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.tailwind.css
```diff
@@ -34,6 +34,10 @@
   @apply p-0 mt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100;
 }
 
+.text-link {
```
Shpigford, line 37: Generally we're trying to avoid/move away from these types of "roll-up" classes. At least at this stage where things are moving quickly on the styling side.

jclusso: For the record, I'm 100% on board with avoiding traditional CSS "roll-up" classes when using Tailwind. That being said, this is one of those select few I always have since text links are pretty common and it's not worth having a component or something for them. If you're very much against it, I can remove it.

Shpigford: I'd prefer to remove for now. We just don't enough established UI convention in place yet to warrant a specific class.

Issue #277, opened 2024-02-03
Docker file bundler version


huseyinulas:
Bundle version mismatch with Gemfile. 
Docker build gives an error below

There was an error parsing `Gemfile`: No such file or directory @ rb_sysopen - /rails/.ruby-version. Bundler cannot continue.

Shpigford:
Can you give more information about terminal commands you used when you got the error along with what OS you're using?

jkostolansky:
Probably because of this: https://github.com/maybe-finance/maybe/blob/e771735c68be6487c884c0acab96d19b6e6ac5de/Dockerfile#L25

There's missing COPY for `.ruby-version` file before `bundle instal`, only `Gemfile`.

Shpigford:
@huseyinulas can you confirm if what @jkostolansky mentioned fixes it for you? if so we can get that merged in.

robzolkos:
This may be related to [this](https://github.com/maybe-finance/maybe/pull/253/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR3) change.

I thought it may have been an older version of Bundler that caused this (the ability to specify a file was introduced in bundler 2.4.19), but running

`docker run --rm ruby:3.3.0-slim bundler --version` reports `Bundler version 2.5.3`

So the reason this breaks is like @jkostolansky said - we don't actually move the `.ruby-version` over prior to bundling at this step.

So there are two options:

1/ Revert that commit and specify the Ruby version in the Gemfile
2/ Modify the docker file to copy over the .ruby-version file before bundler is run.

I'm in favor of less re-configuring of the default Dockerfile.   There are a number of places that need to be changed when a new ruby version is released anyway (Dockerfile, .ruby-version, dev container config etc).  It happens so infrequently that a simple find and replace of `3.3.0` to `3.3.1` is normally sufficient and easy.  I don't think we are saving much by not just putting the ruby version at the top of the Gemfile (which is the default place anyway, and somewhere devs expect to find it).

Shpigford:
I generally agree. I honestly hadn't even seen the `ruby file: ".ruby-version"` syntax before.

Fine with me to revert that line.

zackgilbert:
For what it's worth, I've run into issues with using the `ruby file: ".ruby-version"` method when deploying to some hosts like fly.io. I've had to override projects that use it by commenting this out, write the hardcoded Ruby version in, and deal with the conflicts when merging upstream.

[...]

PR #285, opened 2024-02-03
Revert back to explicit Ruby version in Gemfile

robzolkos:
Closes https://github.com/maybe-finance/maybe/issues/277

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby file: ".ruby-version"
+ruby "3.3.0"
 
 # Rails
 gem "rails", github: "rails/rails", branch: "main"


seanwmitchell:
Hey @robzolkos, why did you change this back? It felt like an elegant solution. Just for my learning.

robzolkos:
> Hey @robzolkos, why did you change this back? It felt like an elegant solution. Just for my learning.

[comment[(https://github.com/maybe-finance/maybe/issues/277#issuecomment-1925452365) here explains what was found.   The default Dockerfile provided by Rails doesn't contain the `.ruby-version` file _before_ it does a bundle so therefore it fails (as it doesn't know what version of ruby to use).   I thought it best to just stick with the defaults for now (not a fan of changing the default Dockerfile _and_ the gemfile.

PR #287, opened 2024-02-04
Fix dynamic classname generation

robzolkos:
Tailwinds JIT compiler doesn't recognize concatenated partial class names and will therefore not include these in the generated css.

Ref: https://tailwindcss.com/docs/content-configuration#dynamic-class-names


Shpigford:
Good catch!

PR #288, opened 2024-02-04
Add PagesControllerTest with authentication

dwightwatson:
This adds a test for the dashboard (within `PagesController` as well as a test helper to handle authentication. 

Happy to move this to a [separate test helper file for authentication](https://guides.rubyonrails.org/testing.html#test-helpers) as mentioned in the guides.


test/controllers/pages_controller_test.rb
```diff
@@ -0,0 +1,9 @@
+require "test_helper"
+
+class PagesControllerTest < ActionDispatch::IntegrationTest
+  test "dashboard" do
+    sign_in users(:bob)
```
robzolkos, line 5: Can we move the sign in do a `setup do` block?  I think that should be a consistent pattern throughout regardless of how small the test file is (as it may grow in the future)

robzolkos (changes requested):

test/controllers/pages_controller_test.rb
dwightwatson: Fair call, done.

test/controllers/pages_controller_test.rb
```diff
@@ -0,0 +1,12 @@
+require "test_helper"
+
+class PagesControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @user = sign_in users(:bob)
```
robzolkos, line 5: Sorry nit pick but `@user` isn't being used yeah?

dwightwatson: Apologies - I was on Minitest autopilot and did what I'm used to. Should be sorted now.

robzolkos (approved):

PR #289, opened 2024-02-04
Split authentication concern from ApplicationController

dwightwatson:
This breaks out all the authentication methods from the root `ApplicationController` into a specific concern and includes it.

It might be interesting to replace the need for `current_user` and `current_family` helpers altogether - I'd imagine using the `Current` model directly should suffice. Happy to explore that if you're down for it.


robzolkos (approved):

Shpigford:
@dwightwatson Good work on this!

Re: replacing `current_user`/`current_family`, you're welcome to open a draft PR exploring that. Just be sure to layout pros/cons.

PR #290, opened 2024-02-04
Add FormBuilder for Maybe forms

JoshAntBrown:
This is an alternative solution to pull request #272. The approach used here is entirely different to the previous pull request, and just focuses on the forms.

I'm keen to get some feedback on this, as I'm not entirely sure I like the solution I have so far, there's an increased level of complexity that I have attempted to abstract away into a generic FormBuilder.

### DIFF
--- a/app/helpers/maybe_form_builder.rb
+++ b/app/helpers/maybe_form_builder.rb
@@ -1,0 +1,14 @@
+class MaybeFormBuilder < TailwindFormBuilder
+  TEXT_LIKE_FIELDS = field_helpers - [ :label, :check_box, :radio_button, :fields_for, :fields, :hidden_field, :file_field ]
+
+  has_styles_for :field_wrapper, "relative border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100"
+  has_styles_for TEXT_LIKE_FIELDS, "p-4 pt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100 w-full"
+  has_styles_for :label, "p-4 pb-0 block text-sm font-medium text-gray-700"
+  has_styles_for :submit, "flex justify-center px-4 py-3 text-sm font-medium text-white bg-black rounded-xl hover:bg-black focus:outline-none focus:ring-2 focus:ring-gray-200 shadow"
+
+  def field_wrapper(&block)
+    @template.content_tag(:div, { class: self.class.styles_for[:field_wrapper] }) do
+      block.call
+    end
+  end
+end
--- a/app/helpers/maybe_form_helper.rb
+++ b/app/helpers/maybe_form_helper.rb
@@ -1,0 +1,7 @@
+module MaybeFormHelper
+  # Returns form_with that uses the MaybeFormBuilder
+  def maybe_form_with(**options, &block)
+    options[:builder] ||= MaybeFormBuilder
+    form_with(**options, &block)
+  end
+end
--- a/lib/tailwind_form_builder.rb
+++ b/lib/tailwind_form_builder.rb
@@ -1,0 +1,78 @@
+class TailwindFormBuilder < ActionView::Helpers::FormBuilder
+  include ActionView::Helpers::TagHelper
+
+  class << self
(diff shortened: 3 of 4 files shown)


app/helpers/maybe_form_helper.rb
```diff
@@ -0,0 +1,7 @@
+module MaybeFormHelper
+  # Returns form_with that uses the MaybeFormBuilder
+  def maybe_form_with(**options, &block)
+    options[:builder] ||= MaybeFormBuilder
+    form_with(**options, &block)
+  end
```
robzolkos, line 6: I think we could use `default_form_builder TailwindFormBuilder` in application_controller and then the default form builder will be that, and we don't need to remember to say `maybe_form_with` we can just use `form_with` as normal.   I think the goal should be to be able to use normal Rails conventions.

app/helpers/maybe_form_builder.rb
```diff
@@ -0,0 +1,14 @@
+class MaybeFormBuilder < TailwindFormBuilder
+  TEXT_LIKE_FIELDS = field_helpers - [ :label, :check_box, :radio_button, :fields_for, :fields, :hidden_field, :file_field ]
+
+  has_styles_for :field_wrapper, "relative border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100"
+  has_styles_for TEXT_LIKE_FIELDS, "p-4 pt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100 w-full"
+  has_styles_for :label, "p-4 pb-0 block text-sm font-medium text-gray-700"
+  has_styles_for :submit, "flex justify-center px-4 py-3 text-sm font-medium text-white bg-black rounded-xl hover:bg-black focus:outline-none focus:ring-2 focus:ring-gray-200 shadow"
+
+  def field_wrapper(&block)
```
robzolkos, line 9: I think field_wrapper could be in a seperate PR.  Its unreleated to the form builder itself (its a styling concern on a field group) and I think could be a simpler helper?

lib/tailwind_form_builder.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  (field_helpers - [ :label, :check_box, :radio_button, :fields_for, :fields, :hidden_field, :file_field ]).each do |selector|
+    class_eval <<-RUBY_EVAL, __FILE__, __LINE__ + 1
+      def #{selector}(method, options = {})
+        merged_opts = apply_classes(options, self.class.styles_for[:#{selector}])
+        super(method, merged_opts)
+      end
+    RUBY_EVAL
+  end
```
robzolkos, line 26: Could we not have all the styles/config in this file rather than spreading over two?

robzolkos:
@JoshAntBrown This is much cleaner than #272 already.  Good job.

ricsdeol:
This (https://evilmartians.com/chronicles/viewcomponent-in-the-wild-embracing-tailwindcss-classes-and-html-attributes) is an interesting article about how to build and reuse components with tailwind with View Components.

About the form, I like to use [Simple Form](https://github.com/heartcombo/simple_form), and its easy configure a tailwind our any CSS framework. 

links: 
- https://github.com/tarellel/simple_form-tailwind/blob/master/lib/generators/simple_form/tailwind/templates/simple_form.rb
- https://dev.to/nejremeslnici/styling-simple-form-forms-with-tailwind-4pel
- https://viewcomponent.org/

Shpigford:
> About the form, I like to use [Simple Form](https://github.com/heartcombo/simple_form), and its easy configure a tailwind our any CSS framework.

Would prefer to not add a new gem/dependency.

JoshAntBrown:
I think on reflection of this PR, I can come up with a better solution that leverages a default form builder and a helper while also extracting the styles into re-usable class(es).

As it stands I'm concerned this solution as it stands is too restrictive and at some point will become a hinderance. I also don't like how I've introduced another way of applying styles through the use of the class method `has_styles_for`, this would be another thing someone has to learn if they want to modify the styles which is arguably just as bad as adding a new gem or dependency.

I'll close this PR, and pick up work on the original PR, moving some of the form builder work across to there.

PR #293, opened 2024-02-04
Fix devcontainer Dockerfile for Ruby 3.3.0 on Apple Silicon

robzolkos:
This compiles a Ruby 3.3.0 with a patch that fixes issues on Apple Silicon Macs.  It is a bandaid until Ruby 3.3.1 when we will revert back to using blessed ruby docker images.


robzolkos:
Waiting on feedback from @leemcalilly on if this fixes the devcontainer on his M2 Mac.

leemcalilly:
@robzolkos This is working for me now. Thanks!

plattenschieber:
I coped with the same issue and fixed it with the patch like you. Gladly, today I found out, that 3.3.1 is available and makes this fix obsolete. 
@robzolkos

PR #297, opened 2024-02-04
Opt-in to Current fully

dwightwatson:
As I floated in #289, this moves towards using `Current` more fully, and replacing the helper methods like `current_user` and `current_family`.

To me, `Current` feels like the more Rails-y way of doing things, and `current_*` style helpers feel more like the Devise way of doing things. Right now we're sitting half-in half-out of both (especially considering we use a before_action called `authenticate_user!`).

* I delegate `Current.family` to the current user's family
* I take inspiration from the [`Current` docs](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html) to set the user or redirect away in the one method

It might even be worth considering dropping `authenticate_user` to just `authenticate` but that's neither here nor there. I do think this PR is important because it makes access to the current user/family **consistent**.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -17,7 +17,7 @@ class AccountsController < ApplicationController
   end
 
   def create
-    @account = Account.new(account_params.merge(family: current_family))
+    @account = Account.new(account_params.merge(family: Current.family))
     @account.accountable = account_params[:accountable_type].constantize.new
 
     if @account.save
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -2,31 +2,21 @@ module Authentication
   extend ActiveSupport::Concern
 
   included do
-    helper_method :current_user
-    helper_method :current_family
     helper_method :user_signed_in?
   end
 
   private
 
   def authenticate_user!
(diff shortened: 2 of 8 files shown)


robzolkos (approved):
Love this.

ricsdeol (approved):

app/controllers/accounts_controller.rb
```diff
@@ -17,7 +17,7 @@ def show
   end
 
   def create
-    @account = Account.new(account_params.merge(family: current_family))
+    @account = Account.new(account_params.merge(family: Current.family))
```
jclusso, line 20: In the Current [docs](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html) they show how you can make this even cleaner by setting the default on the model using `Current` instead of having to merge it in with params. 

```ruby
belongs_to :creator, default: -> { Current.user }
```

dwightwatson: I was about to follow up with another PR that adds test for this controller and then uses the association fully...

```rb
@account = Current.family.accounts.new(account_params)
```

I think that's the clearest option. I personally quite like the `default:` option though.

Shpigford:
Great job on this, @dwightwatson!

PR #298, opened 2024-02-04
Use Current association in AccountsController and add tests

dwightwatson:
This adds some test coverage over accounts, and also uses the `Current` to build the account using the association.

* As mentioned in (#297)[https://github.com/maybe-finance/maybe/pull/297#discussion_r1477462566] we could instead set `Current.family` to be the default owner of new accounts

I'm not sure if these test PRs are helpful or not, or even in line with how you had envisioned testing Maybe. I generally tend to test the red paths as well as the green paths, but I've been taking inspiration from the Campfire source which is dead simple tests over the green paths.


Shpigford:
> I'm not sure if these test PRs are helpful or not, or even in line with how you had envisioned testing Maybe. I generally tend to test the red paths as well as the green paths, but I've been taking inspiration from the Campfire source which is dead simple tests over the green paths.

We're still feeling that out a bit. Generally speaking will likely err towards lighter on functional testing and very heavy on data-accuracy testing as the "data" part will always need to be "right" but actual functionality within the app will likely change a decent amount over the coming weeks/months as we get this thing off the ground.

At any rate, looks good to me. 🙂

PR #301, opened 2024-02-05
Add Simplecov Gem

JuanVqz:
This will help us to make sure we are testing all parts of our code.

It will be nice to implement the [Codecov Report](https://about.codecov.io/blog/getting-started-with-code-coverage-for-ruby/) when opening a PR.

https://docs.codecov.com/docs/pull-request-comments

Run the tests using `WITH_COVERAGE=true` to see the report under the `coverage` folder

https://github.com/simplecov-ruby/simplecov

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,7 @@ jobs:
         env:
           RAILS_ENV: test
           DATABASE_URL: postgres://postgres:postgres@localhost:5432
+          WITH_COVERAGE: true
           # REDIS_URL: redis://localhost:6379/0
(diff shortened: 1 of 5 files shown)


zachgoll:
Hey @JuanVqz!  Really appreciate the contribution.

We definitely value testing here, but we're trying to limit the number of additional Gems and overhead added to the project early on for the sake of simplicity.

I think we will definitely consider something like this in the future, but our primary focus at the moment is figuring out the core features that need to be built and optimizing for "minimalism" in our CI/CD process.

JuanVqz:
Sounds good 👍

PR #305, opened 2024-02-05
First pass at a system test

dwightwatson:
This introduces a simple system test for creating a new account:

* It fixes an issue with the label for account balance... using form helpers for labels would have made this easier to spot, but it looks like there is already a discussion around building custom form builders,
* I added labels for "New account" and "Submit" to interact with them appropriately. I'm not familiar with the original Maybe design but perhaps we can work on improve labelling,
* Added a `sign_in` system helper that matches the API of the controller helper.


test/system/accounts_test.rb
```diff
@@ -0,0 +1,18 @@
+require "application_system_test_case"
+
+class AccountsTest < ApplicationSystemTestCase
+  setup do
+    sign_in @user = users(:bob)
```
robzolkos, line 5: A pattern I see often and we could consider `sign_in 'bob'`  or `sign_in users(:bob)`

```
def sign_in(user)
  user = users(user) unless user.is_a? User
  ....
end
```

Gives a little more flexibility in usage

robzolkos (approved):
This is great!  I left a comment for consideration.

Shpigford:
This feels basic enough to be okay for now, but we need to be really careful to not over-test at this stage. Lot of UI/UX changes will likely happen on a near daily basis, so just don't want the community to constantly be fixing tests when they could be building out base features.

dwightwatson:
Gotcha, that's fair. I'll stop here. Don't feel the need to merge this one if you'd rather hold off.

test/system/accounts_test.rb
dwightwatson: This is really cute. I only saw it recently for the first time in Campfire.

robzolkos: same

PR #307, opened 2024-02-05
Add Windows compatibility and Docker setup; Upgraded README.md file

### LINKED ISSUE #306: Windows develop support
# Summary
Hey! It looks here are no Windows develop support. I am using Windows for developing stuff and I personally don't really like `devcontainers` (probably due to low JetBrains support of that)
## Task
Bring some documentation and do other fixes for supporting Windows as development platform.

yeskiy:
Resolves #306; Add Windows compatibility and Docker setup Upgraded README.md file

Added support for Windows system by including different versions of dependencies in `Gemfile`. Also introduced Docker setup to the project via a docker-compose file for easy setup and environment consistency. Files for Windows-specific development workflow and a Docker setup guide were also added to the README. Lastly, a .gitignore update included to ignore files from JetBrains products and Docker's PostgreSQL volume data.


robzolkos:
I'm concerned at the complexity this is introducing (and potential maintenance burden).   But having everyone being able to contribute is also important.

I think this addresses the segment of "windows users who use jetbrains ide's"  but unsure if this is worthy of introducing a third way (after wsl/*nix and devcontainer support).   (devcontainer can run on the command line and doesn't require an IDE,  you can use whatever ide/code editor you like).

Is there a away this could be addressed for those that prefer this combo (and would be used to the setup challenges it has) with a readme/wiki entry?

yeskiy:
Hey @robzolkos !
So the files that directly related to the windows developing process are [bin/dev.ps1](https://github.com/maybe-finance/maybe/pull/307/files#diff-8e7df44273fffb2898a9b58718e1d88911992339437a5d5cc0af2b7939c215b7) and [Procfile.windows.dev](https://github.com/maybe-finance/maybe/pull/307/files#diff-58133ead4b0fa55af2edba2875cfc52054123ac5584d594d44ae9ddb703af6da) which are additional replacement files for adding an ability to run project as previously intended in the documentation. As a JetBrains user, I can run devcontainer and even run the whole project without these changes. So the answer to the question:
> Is there a away this could be addressed for those that prefer this combo (and would be used to the setup challenges it has) with a readme/wiki entry?

Is definitially **yes**. But if you are talking about changes inside [Gemfile.lock](https://github.com/maybe-finance/maybe/pull/307/files#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731f), here is no way omitting of changing this file (maybe just not staging it, which is not really technically correct).

yeskiy:
>1/ For clarity - this is for peeps not using WSL correct?
Yep,
> 2/ Is there a way we can not introduce a second docker-compose.yml file (there is one already in .devcontainer folder). I'm thinking being able to run docker-compose -f .devcontainer/docker-compose.yml start db (or similar).
Second docker-compose file is really ambiguous, you're totally right! I did second one since `devcontainer ` one use `latest` tag, and I was wanted to test if I will get any problems with 16th version specifically.

robzolkos:
>  I was wanted to test if I will get any problems with 16th version specifically

We can update the docker-compose.yml file in the .devcontainers folder with ` image: "postgres:${POSTGRES_VERSION:-latest}"`  and it will use the latest if a specific version is not provided.

That clears up the need to have another docker-compose file introduced.  The remaining changes are ok and won't be a maintenance nightmare going forward I don't think. (aside: I don't use Windows but its surprising to me that windows needs all this extra cruft - I think most rails development happens in WSL nowadays).

zachgoll:
Hey @yeskiy, thanks for the setup here, I'm definitely in favor of making it easy for Windows devs to contribute.

Per @robzolkos comments, I'm a little worried that we're adding too many options for getting the project started this early on.  Just curious, is not using WSL just a personal preference?  Or did you run into some issues with it?

yeskiy:
Hey @zachgoll!
Yep, I can enjoy using WSL itself, but devil in the details here. JetBrains have “Beta” support of `devcotainer` (in real world, it is almost unusable when you try to develop something, a lot of bugs are still presented).
I personally don't use `VS Code` for developing, only for small stuff.  

P.S. @robzolkos did your requested change related to `docker-compose.yaml` in root folder

zachgoll:
Hey @yeskiy, I've opened a PR that addresses various setup related issues - https://github.com/maybe-finance/maybe/pull/339

I also created a [guide for getting started with Windows development](https://github.com/maybe-finance/maybe/wiki/Windows-Dev-Setup-Guide).  Ultimately, we'd like to stick with WSL as our recommendation for building with Windows since that closely mirrors our production environment and is what we think is the most common ways to run a Rails app on Windows.  Would you mind taking a read through that guide and letting me know if/where you get stuck?

zachgoll:
Closing this out as we now have a Windows Setup guide and updated README for local dev:

https://github.com/maybe-finance/maybe/wiki/Windows-Dev-Setup-Guide

To keep things simple for new contributors, we're recommending the WSL approach to running on Windows.  If any updates need to be made to the existing guide just let me know here in a comment and I'll get them added!

Issue #308, opened 2024-02-05
Avoid Forcing SSL in Production Environment


DomiiBunn:
**Description:**
Currently, the option to force SSL is hardcoded in the production environment. This can lead to complications for deployments behind reverse proxies, especially when utilizing services like Cloudflare Zero Trust or Nginx, which handle certificate management.

**Suggested Improvement:**
Consider making the SSL enforcement a configurable option through an environment variable. This enhancement would enhance flexibility and simplify deployments for users relying on reverse proxies and external services for SSL termination.

https://github.com/maybe-finance/maybe/blob/622fc07a7624f79d5d025997495c13efba8ff7d1/config/environments/production.rb#L46

**Benefits:**
1. **Deployment Flexibility:** Users deploying behind reverse proxies with SSL termination services can seamlessly integrate without conflicting SSL configurations.
  
2. **Simplified Configuration:** By allowing SSL enforcement to be set via an environment variable, deployment configurations become more straightforward, reducing potential errors.

3. **Adaptability:** Users can easily adapt the SSL enforcement setting based on their specific deployment requirements, making the application more versatile.

**Additional Context:**
Provide documentation updates reflecting the new environment variable and its impact on SSL enforcement. This ensures users are informed and can make informed decisions during the deployment process.

robzolkos:
Is setting the `assume_ssl` [config](https://github.com/maybe-finance/maybe/blob/622fc07a7624f79d5d025997495c13efba8ff7d1/config/environments/production.rb#L43) a better way to handle this?  (I've not ever seen the force_ssl bit switched off in prod and thinking assume_ssl was introduced for the scenarios you describe?)

DomiiBunn:
It could be however if a deployment is made via docker it might cause other issues for local enviroments that might not want to go into the configs.

Avoids future issues imo

ivarprudnikov:
> Currently, the option to force SSL is hardcoded in the production environment.

I see it as a sensible default. You should tweak the environment config file before the deployment to production anyway.

The already suggested use of [`assume_ssl`](https://api.rubyonrails.org/v7.1.0/classes/ActionDispatch/AssumeSSL.html) config seems what folks could use in this scenario.

DomiiBunn:
Why not just expost all the config options as environment flags, still set
everything correctly as the fallback but not require the end user to edit
multiple configs rather than just a .env file/docker env variables/system
environment while deploying (like 90% of other open source software does)

zachgoll:
[...]

PR #314, opened 2024-02-05
Avoid Forcing SSL in Production Environment Issue Fixed

amacleod1417:


dwightwatson:
I think the Rails default makes more sense here. 

If you genuinely wanted this off in production you could make the change on your end.

Shpigford:
What's the scenario for wanting to run this _without_ SSL in a production environment?

m4tt72:
> What's the scenario for wanting to run this _without_ SSL in a production environment?

When running behind a reverse proxy that already handles SSL

robzolkos:
I think this is what this [config](https://github.com/maybe-finance/maybe/issues/308#issuecomment-1927198572) setting is for

config/environments/production.rb
```diff
@@ -42,8 +42,11 @@
   # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
   # config.assume_ssl = true
 
-  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+# Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies. 
+#If the variable is set to 'true', SSL enforcement will be enabled; otherwise, it will be disabled.
+config.force_ssl = ENV.fetch("FORCE_SSL", "true") == "true"
```
haydenk, line 47: Should the default value not a return a boolean value instead of a string?

```ruby
ENV.fetch("FORCE_SSL", true) == "true"
```

If there's a better option than using `"true"` as the value to look for in the environment variable I would use that but I cannot think of one that isn't overly complex.

haydenk:
I agree with most everyone else that it isn't really necessary but I do see one circumstance where this could be useful, it just depends on the direction the project is going. 

It's useful if the application is dockerized and you need to change the setting but there isn't an official docker configuration in the repository, so you can just build your own image change the config setting.

If the application gets the updates for running the application in docker then this change could be useful.

zachgoll:
@amacleod1417 let's go ahead and implement the solution proposed here in this issue:

https://github.com/maybe-finance/maybe/issues/308#issuecomment-1934615353

Also, could you please fix the lint errors?  Will get this merged after!

zachgoll:
Closing for now.  Will address SSL in a dedicated issue alongside the creation of an official Dockerfile

PR #317, opened 2024-02-06
Make all account names human readable

varkum:
Screenshot 2024-02-05 at 8 40 15 PM
Screenshot 2024-02-05 at 8 40 09 PM

Fixed the above two names in the form to the following:

Screenshot 2024-02-05 at 9 04 27 PM
Screenshot 2024-02-05 at 9 04 37 PM

I converted the `type_name` method to a class method. To me, it makes more sense to pass the class in the `type` local for the `_account_type` partial instead of instantializing each type just to access the `type_name`. It also eliminates the need to call `.class` then on every instance of the type in the partial.

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -1,5 +1,2 @@
 module AccountsHelper
-  def human_account_name(camelCaseName)
-    camelCaseName.gsub(/(?<!^)([A-Z])/, ' \1')
-  end
 end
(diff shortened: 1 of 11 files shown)


robzolkos (approved):

robzolkos (approved):

varkum:
Is there a way to use a safe `constantize` here? We have to either rely on the accountable_type model attribute or the query param, both of which are marked as unsafe by scan_ruby

ricsdeol:
I think this coukd be closed after #256

Shpigford:
A lot of changes to the add account flow so going to close this.

Issue #319, opened 2024-02-06
`accounts` contains object information in 2nd column instead of the account type


gkeswani92:
Spun up the repository in development and added a couple of test entries for accounts. Every new account addition takes me to `accounts/new` that has the account object being shown in the 2nd column (see image below)

<img width="1480" alt="image" src="https://github.com/maybe-finance/maybe/assets/4531477/f045515b-f582-4c15-a402-4e3f11e6a4b6">

Happy to fix this if someone can clarify what the expected values in the column are. Is it the type of account?

gkeswani92:
I've put up a temporary fix here to ensure the right information is displayed: https://github.com/maybe-finance/maybe/pull/320

PR #320, opened 2024-02-06
Display account type instead of accoumtable object on accounts index

gkeswani92:
- Displays the type of account on the accounts index page

image
 
instead of the object information

image

Resolves: https://github.com/maybe-finance/maybe/issues/319

### DIFF
--- a/app/views/accounts/index.html.erb
+++ b/app/views/accounts/index.html.erb
@@ -8,7 +8,7 @@
         <%= account.name %>
       </div>
       <div class="flex items-center text-sm">
-        <%= account.accountable %>
+        <%= account.accountable.type_name %>
       </div>
       <p class="text-sm text-right">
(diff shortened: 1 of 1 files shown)


pmareke:
What do you think about align the column, neither center or to the left, and make the UI looks nicer?

Thanks for your work!

AfolabiOlaoluwa:
Was just about making a submission for this. I agree with your opinion. Account type is better. 

If "accountable" is necessary/must, perhaps using the `.accoubtable_type` is okay which is a bit devoid of rendering object on the template.

Shpigford:
Hey! Looks like a test is failing on this one. FWIW, that account view you're seeing on the right is a temp view while we finish up some design bits. It'll get completely replaced here in the next 7-10 days.

If you want to tackle that failing test and resubmit, happy to push this through, though!

ricsdeol:
Could be close after #256

PR #325, opened 2024-02-06
Add .editorconfig file

bopm:
This PR adds [.editorconfig](https://editorconfig.org/) file to the project to set clear expectations of file format and allow editor automation to correct small issues (like trailing whitespaces, final new lines, etc).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length=120


pmareke:
I prefer to use the same approach as Rails 8 will do for the formatting ([here](https://github.com/rails/rails/issues/50456)).

We can add a pre-commit hook and also a new step in the pipelines.

bopm:
But it's not a linter. Editorconfig is a way to tell your editor (or editor plugin) about expected file format conventions. For example that project uses 100 chars per line. Also, it gives you the ability to apply different formatting for subformats (like JSON or YAML).

pmareke:
> But it's not a linter. Editorconfig is a way to tell your editor (or editor plugin) about expected file format conventions. For example that project uses 100 chars per line. Also, it gives you the ability to apply different formatting for subformats (like JSON or YAML).

What do you mean is not a linter? Rubocop?

In their repo they define it as: "RuboCop is a Ruby static code analyzer (a.k.a. linter)".

bopm:
I am speaking about editorconfig, which is not a linter. It does not enforce or analyze code. It just tells the editor, that when you add a new file and forget to add a new line at the end of it, it does it for you automatically if you configured that setting in it. Or where to display the right margin line according to the chars per line setting.

pmareke:
I don't know, honesty I prefer to let the developer choose how they want to see their code in their editor and enforce some styles in the repository.

bopm:
Up to you, feel free to close this one.

pmareke:
No, not at all ahaha.

I'm here like you trying to help. Whatever the maintainers decide it ok to me, I just wanted to debate about it.

robzolkos:
Could a compromise solution be to have .editorconfig in the .gitignore file, then the developer can use whatever they like? (rubocop will pull them over and give them a ticket in CI if there are lint issues).

bopm:
@robzolkos I would believe that it'll defy the goal of having this file. It may be configured to be less restrictive, but things like tabs or spaces should not be debatable in big open-source repos. It's a way to remove the responsibility of thinking about such things from contributors. Rubocop linting for such is still needed for cases when a contributor is using an editor that does not support this config convention. But again, it's my opinion, not something rock solid.

Shpigford:
@bopm Thank you so much for submitting this!

In general I tend to agree with the sentiment that people can code however they want and then just let the linter do any necessary cleanup, so going to close this for now.

PR #327, opened 2024-02-06
Create docker push workflow

indykoning:
This will automatically push docker images into the github container repository under the organisation

### DIFF
--- a/.github/workflows/push-docker-container.yml
+++ b/.github/workflows/push-docker-container.yml
@@ -1,0 +1,57 @@
+name: "Push Docker Container"
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Additional tag to push'
+        required: false
+  push:
+    branches:
+      - 'main'
+    tags:
+      - '*'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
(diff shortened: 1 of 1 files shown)


zachgoll:
Hey @indykoning this is great, just want to make sure I'm following—this setup will push a _future_ `Dockerfile` (that hasn't been created yet) to Github Package repository correct?

.github/workflows/push-docker-container.yml
```diff
@@ -0,0 +1,57 @@
...
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Additional tag to push'
+        required: false
+  push:
+    branches:
+      - 'main'
+    tags:
+      - '*'
```
tebayoso, line 13: I would recommend to push only certain tags, for example:
```suggestion
      - 'RC*'
      - 'preview*
```

indykoning: During development i absolutely agree, though wouldn't we want real version numbers such as `1.0.0` or `2024.2` to have their image created as well?

zachgoll:
@indykoning this is great, but our Dockerfile is not quite ready and we'll want to make sure that once we start pushing images to GHCR, all files work.

I think we can either close this out and come back to it later or make the `Dockerfile` updates directly in this PR

zachgoll:
Closing out, will reference this when we introduce official support for a Dockerfile.

PR #330, opened 2024-02-06
Fix accountable text and alignment

### LINKED ISSUE #319: `accounts` contains object information in 2nd column instead of the account type

kiliczsh:
Resolves #319 with helper method.
Screenshot 2024-02-06 at 22 30 39


kiliczsh:
@Shpigford @robzolkos @sebastiancaraballo for your attention

PR #338, opened 2024-02-06
Implement Cloudless Flipper as feature flag management tool

kiliczsh:
Feature flags are very useful for product development. I have prepared initial implementation with Flipper. @Shpigford

Path: /flags


Shpigford:
For now we're keeping additional tooling to a minimum. Not out of the question long term, but for now we're laser focused on foundational product work and keeping complexity to a minimum.

Thank you, though!

PR #341, opened 2024-02-06
Feature to capture additional fields in credit form

### LINKED ISSUE #329: Add new "Credit card" account form
We're working on adding each new account type as they've got slightly different "field" needs.

All accounts have "Name" and "Balance" by default.

Additional fields:
* Type — Select (Mastercard, Visa, American Express, Discover)
* Limit — Number

View is in `/views/accounts/account/_credit.html.erb`

sarvaiyanidhi:
Closes #329

This PR implements below changes

- Migration to add news fields in account_credits table
- Enabling nested attributes on a delegated_type
- Capturing nested_attribute in create action of Account

### DIFF
--- a/app/models/account/credit.rb
+++ b/app/models/account/credit.rb
@@ -1,3 +1,5 @@
 class Account::Credit < ApplicationRecord
   include Accountable
+
+  CARD_TYPES = %i[mastercard visa american_express discover].freeze
 end
--- a/app/views/accounts/account/_credit.html.erb
+++ b/app/views/accounts/account/_credit.html.erb
@@ -1,0 +1,11 @@
+<div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+    <label for="type" class="block text-sm font-medium opacity-50 focus-within:opacity-100">Type</label>
+    <%= ff.select :card_type, options_for_select([["Mastercard", "mastercard"], ["Visa", "visa"],["American Express","american_express"],["Discover","discover"]], selected: ""), {}, class: "block w-full p-0 mt-1 bg-transparent border-none focus:outline-none focus:ring-0" %>
+  </div>
+  <div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+    <span class="absolute px-2 py-1 text-xs font-medium text-gray-400 uppercase bg-gray-200 rounded-full opacity-50 right-3">Optional</span>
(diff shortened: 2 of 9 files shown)


sarvaiyanidhi:
@Shpigford I have question here related to subtype field defined in app/views/accounts/account/_depository.html.erb
Is that field specific to account_depositories table? If so, why is it added in accounts table and in case it is for accounts table so can it be part of new.html.erb

Not sure the solution I followed of nested_attribute for delegate_type is correct approach but due to this solution subtype field is not saving up in database as it is considering it as nested attribute.

Please advice on implementation if I need to change the logic.

sarvaiyanidhi:
Just referred implementation done for #343 .. If that is the case I think I complicated requirement assuming we need to have separate fields to that type

jespr:
I think `subtype` is used as the "subtype" for the account. So for accounts where type is "credit", subtype can be any of: Mastercard, Visa, American Express, Discover

sarvaiyanidhi:
Yeah, got it I thought it needs to be separate fields related to specific table and complicated used nested_attribute form concept

jespr:
I like the way you've abstracted the types to the model - I think I'm going to do that in my PR as well!

sarvaiyanidhi:
I think let us wait for input from @Shpigford as not sure if what I implemented is required as depository type also follows your approach,

Shpigford:
subtypes just help specify the type.

So "depository" is just the high level with subtypes of "checking" and "savings" for instance.

And "investment" would have subtypes like "401k" and "roth ira"

app/views/accounts/account/_credit.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+    <label for="type" class="block text-sm font-medium opacity-50 focus-within:opacity-100">Type</label>
```
ricsdeol, line 2: Should use rails form help `ff.label :card_type`

And use internatiolizations...

sarvaiyanidhi:
Thank you for clarification I will just push new PR as this one is complicated by adding new attributes to account_credits table

app/views/accounts/account/_credit.html.erb
sarvaiyanidhi: Thanks @ricsdeol Noted. Will do in new PR

app/models/account/credit.rb
```diff
@@ -1,3 +1,5 @@
 class Account::Credit < ApplicationRecord
   include Accountable
+
+  CARD_TYPES = %i[mastercard visa american_express discover].freeze
```
ricsdeol, line 4: Coukd use rails enum
https://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

Issue #340, opened 2024-02-06
Dev container frontend not working


LosBagros:
I have a problem with accessing frontend

Curl inside container worked somehow...

Steps to reproduce:
1. Clone repo
2. Open dev container
3. `bin/dev`
4. Cannot access frontend

![image](https://github.com/maybe-finance/maybe/assets/45005533/256edce3-ae29-445c-a10a-489d6e31e250)
![image](https://github.com/maybe-finance/maybe/assets/45005533/80dc1d93-7014-4296-a9e2-bc737f9e5909)

robzolkos:
Try updating the line in `Procfile.dev` to `web: bin/rails server -b 0.0.0.0` and see if that makes a difference before running `bin/dev` again

zachgoll:
As we resolve this one, can we add the solution in a comment on #339 just so I can add it to the Dev Container section?

LosBagros:
Thx

caioluis:
@robzolkos I'm having the same issue, but as I already have the latest version of `Procfile.dev`, I wonder what could cause this.

PR #342, opened 2024-02-06
change server listening address to global

lcx:
listen to 0.0.0.0 instead of just local interface so it can work in Docker (and herby in devcontainers) refs #340
Credits go to @robzolkos

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,2 +1,2 @@
-web: bin/rails server
+web: bin/rails server -b 0.0.0.0
 css: bin/rails tailwindcss:watch


robzolkos:
Im concerned this will affect non devcontainer users.  Will need to be checked by someone who runs the app on metal.

jespr:
This works fine for me running it locally on my Macbook Pro (non-docker)

robzolkos (approved):

lcx:
I indeed haven't started it without devcontainer, was already midnight when I made the PR. But listening on 0.0.0.0 still allows you to connect via 127.0.0.1 should be no issue that's why I pushed without testing without devcontainer.

Issue #331, opened 2024-02-06
Add new "Investment" (balance) account form


Shpigford:
We're working on adding each new account type as they've got slightly different "field" needs.

All accounts have "Name" and "Balance" by default.

This specific "investment" form is for tracking account _balance_ (not tracking individual holdings).

Additional fields:
* Type — Select 

Types to add...
* Brokerage
* Pension
* Retirement
* 401(k)
* 529 plan
* Health Savings Account
* Mutual Fund
* Roth IRA
* Roth 401k

View is in `/views/accounts/account/_investment.html.erb`

jespr:
I can take a stab at this

Shpigford:
Implemented with #343

PR #343, opened 2024-02-06
Add Account type investment form

jespr:
This implements a new form for the the investment account type per https://github.com/maybe-finance/maybe/issues/331


jespr:
@Shpigford is this what you had in mind?

Shpigford:
Yep! This looks good to me.

PR #268, opened 2024-02-03
Integrate money-rails gem

sarvaiyanidhi:
Claim #265 

Closes #265 

Have used existing column balance of accounts field but value is stored in cents by using as option in model.

Default currency is set in money initializer file to USD

PR #344, opened 2024-02-07
Add new fields in Credit form

### LINKED ISSUE #329: Add new "Credit card" account form

sarvaiyanidhi:
Closes #329 

This PR address below changes
- New field 'limit' in accounts table
-  Credit type of account captures type and limit details

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -30,7 +30,7 @@ class AccountsController < ApplicationController
   private
 
   def account_params
-    params.require(:account).permit(:name, :accountable_type, :balance, :balance_cents, :subtype)
+    params.require(:account).permit(:name, :accountable_type, :balance, :balance_cents, :subtype, :limit)
   end
 
   def account_type_class
--- a/app/models/account/credit.rb
+++ b/app/models/account/credit.rb
@@ -1,3 +1,9 @@
 class Account::Credit < ApplicationRecord
   include Accountable
+  SUBTYPES = [
+    [ "Mastercard", "mastercard" ],
+    [ "Visa", "visa" ],
+    [ "American Express", "american_express" ],
+    [ "Discover", "discover" ]
+  ].freeze
 end
--- a/app/views/accounts/account/_credit.html.erb
+++ b/app/views/accounts/account/_credit.html.erb
@@ -1,0 +1,8 @@
+<div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+  <%= f.label :subtype, t('.type') ,class: 'block text-sm font-medium opacity-50 focus-within:opacity-100'  %>
+  <%= f.select :subtype, options_for_select(Account::Credit::SUBTYPES, selected: ""), {}, class: "block w-full p-0 mt-1 bg-transparent border-none focus:outline-none focus:ring-0" %>
+</div>
+<div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+  <%= f.label :limit, t('.limit') ,class: 'block text-sm font-medium opacity-50 focus-within:opacity-100'  %>
+  <%= f.number_field :limit, placeholder: number_to_currency(0), in: 0.00..100000000.00, required: 'required', class: "p-0 mt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100" %>
+</div>
--- a/db/migrate/20240207021033_add_limit_to_accounts.rb
+++ b/db/migrate/20240207021033_add_limit_to_accounts.rb
@@ -1,0 +1,5 @@
+class AddLimitToAccounts < ActiveRecord::Migration[7.2]
+  def change
+    add_column :accounts, :limit, :bigint, default: 0
+  end
+end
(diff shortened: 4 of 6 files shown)


OohBen (approved):

db/migrate/20240207021033_add_limit_to_accounts.rb
```diff
@@ -0,0 +1,5 @@
+class AddLimitToAccounts < ActiveRecord::Migration[7.2]
+  def change
+    add_column :accounts, :limit, :bigint, default: 0
```
rmarescu, line 3: Is this field meant to store money? If so, should use the `limit_cents` convention, as per a [previous discussion](https://github.com/maybe-finance/maybe/pull/268)?

app/views/accounts/account/_credit.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+  <%= f.label :subtype, t('.type') ,class: 'block text-sm font-medium opacity-50 focus-within:opacity-100'  %>
```
rmarescu, line 2: Nit - to match the surrounding code

```suggestion
  <%= f.label :subtype, t(".type") ,class: "block text-sm font-medium opacity-50 focus-within:opacity-100"  %>
```

app/views/accounts/account/_credit.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  <%= f.select :subtype, options_for_select(Account::Credit::SUBTYPES, selected: ""), {}, class: "block w-full p-0 mt-1 bg-transparent border-none focus:outline-none focus:ring-0" %>
+</div>
+<div class="relative p-4 border border-gray-100 bg-offwhite rounded-xl focus-within:bg-white focus-within:shadow focus-within:opacity-100">
+  <%= f.label :limit, t('.limit') ,class: 'block text-sm font-medium opacity-50 focus-within:opacity-100'  %>
```
rmarescu, line 6:
```suggestion
  <%= f.label :limit, t(".limit") ,class: "block text-sm font-medium opacity-50 focus-within:opacity-100" %>
```

app/views/accounts/account/_credit.html.erb
```diff
@@ -0,0 +1,8 @@
...
+  <%= f.number_field :limit, placeholder: number_to_currency(0), in: 0.00..100000000.00, required: 'required', class: "p-0 mt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100" %>
```
rmarescu, line 7:
```suggestion
  <%= f.number_field :limit, placeholder: number_to_currency(0), in: 0.00..100000000.00, required: "required", class: "p-0 mt-1 bg-transparent border-none opacity-50 focus:outline-none focus:ring-0 focus-within:opacity-100" %>
```

app/models/account/credit.rb
```diff
@@ -1,3 +1,9 @@
 class Account::Credit < ApplicationRecord
   include Accountable
+  SUBTYPES = [
```
rmarescu, line 3: Could `SUBTYPES` be an array, and the human-friendly version defined as translations?

```ruby
  SUBTYPES = %w[ mastercard visa american_express discover ]
```

app/views/accounts/account/_credit.html.erb
rmarescu, line 7: If `limit` must be positive, should add some model validation to enforce it.

app/controllers/accounts_controller.rb
```diff
@@ -30,7 +30,7 @@ def create
   private
 
   def account_params
-    params.require(:account).permit(:name, :accountable_type, :balance, :balance_cents, :subtype)
+    params.require(:account).permit(:name, :accountable_type, :balance, :balance_cents, :subtype, :limit)
```
rmarescu, line 33: Could the change have some test coverage?

rmarescu (commented):
@sarvaiyanidhi I've added some comments for consideration!

app/views/accounts/account/_credit.html.erb
pmareke: Good catch! I think we should think about in adding Rubocop asap and detect this kind of errors automatically.

db/migrate/20240207021033_add_limit_to_accounts.rb
sarvaiyanidhi: I was not sure that limit was also supposed to be money field. @Shpigford Can you confirm if we want to monetize limit field as well which will generate limit_cents and limit_currency in accounts table.

rmarescu: I've noticed that #268 introduced `balance_currency` (associated with `balance_cents`. Is there a case where there is a need for separate currency fields for each money fields? Like the balance be in USD, but the limit in CAD for example?

Is there a way to have just one currency field, that can be shared by all money fields?

sarvaiyanidhi: Yeah, even I was thinking on same. Even there is separate 'currency' field already present in accounts table. Not sure if we still need it as balance_currency is also added in table.

Shpigford:
Closing for new per the comment here: https://github.com/maybe-finance/maybe/issues/329#issuecomment-1932637291

PR #390, opened 2024-02-08
Fixes #382 Changed Add Account to use account model name

### LINKED ISSUE #382: Add account form should output the type of account
In the flow of adding an account, by the time you get to the form it should say which account type you're adding.

So if you clicked "Bank account" it should say "Add bank account", for example.

This is located in `views/accounts/new.html.erb`

CleanShot 2024-02-08 at 12 17 07@2x

2000rosser:
Fixes #382 The change currently only works for adding a 'Bank Account' as adding other types of accounts only creates a new Bank Account. When new types of accounts are added this will use the account name in the models as the title.

Signed-off-by: 2000rosser

### DIFF
--- a/app/views/accounts/new.html.erb
+++ b/app/views/accounts/new.html.erb
@@ -65,11 +65,11 @@
         <%= lucide_icon('arrow-left', class: 'text-gray-500 w-5 h-5') %>
       <% end %>
       <span>
-        <% if params[:type].in? Account.allowable_account_types %>
-          <% model_class = "Account::#{params[:type]}".constantize %>
-          Add <%= model_class.model_name.human(count: 2).downcase %>
+        <% model_class = find_model_class(params[:type]) %>
+        <% if model_class %>
+          Add <%= model_class.model_name.human(count: 2) %>
         <% else %>
-          Add account
+          Add Account
         <% end %>
       </span>
     </div>
--- a/app/views/accounts/new.html.erb
+++ b/app/views/accounts/new.html.erb
@@ -66,12 +66,12 @@
       <% end %>
       <span>
         <% if params[:type].in? Account.allowable_account_types %>
-          Account::#{params[:type]}".constantize
-          end %>
-          Add <%= model_class ? model_class.model_name.human(count: 2).downcase : "account" %>
+          <% model_class = "Account::#{params[:type]}".constantize %>
+          Add <%= model_class.model_name.human(count: 2).downcase %>
         <% else %>
-          Add Account
-        ß<% end %>
+          Add account
+        <% end %>
+      </span>
     </div>
     <%= form_with model: @account, url: accounts_path, scope: :account, html: { class: "space-y-4 m-5 mt-1", data: { turbo: false } } do |f| %>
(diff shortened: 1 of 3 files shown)


2000rosser:
I believe this new change will resolve the vulnerability scan issue

app/views/accounts/new.html.erb
```diff
@@ -65,11 +65,19 @@
...
+          <% allowable_types = ["Depository", "Investment", "Property", "Vehicle", "Credit", "Loan", "OtherAsset", "OtherLiability"] %>
+          <% if allowable_types.include?(params[:type]) %>
+            <% model_class = "Account::#{params[:type]}".constantize %>
+            Add <%= model_class.model_name.human(count: 2) %>
+          <% else %>
+            Add Account
+          <% end %>
+        <% else %>
+          Add Account
+        <% end %>
```
Shpigford, line 78: Super heavy on the if/else statements, which feels a little hard to follow/maintain.

What about something like this? Not 100% positive it'll pass the ruby checks. 
```
<span>
        <% model_class = if params[:type].in?(["Depository", "Investment", "Property", "Vehicle", "Credit", "Loan", "OtherAsset", "OtherLiability"])
                  "Account::#{params[:type]}".constantize
                end %>
        Add <%= model_class ? model_class.model_name.human(count: 2).downcase : "account" %>
      </span>
```

2000rosser: Yeah definitely worth a shot, I'll add it now

robzolkos: We should centralize/codify the allowable types as I'm sure they will be queried everywhere

```ruby
  # in Account model
  def self.allowable_accountable_types
    accountable_types.map { |type| type.demodulize }
  end
```

Then can do 

```ruby
<% if params[:type].in? Account.allowable_account_types %>
 ... 
<% else %>
   Add Account
<% end %>
```

And probably best to move to a helper to clean up conditionals out of the view.

2000rosser: Could it be worthwhile adding new models for singular and plural cases?
```yml
models:
      account/depository:
        singular: Bank Account
        plural: Bank Accounts
```

robzolkos: Rails handles that with the `pluralize` methods, so I'm not sure we do (yet)

2000rosser: The current implementation has Bank Accounts and Investments as plural by default, should they be changed to singular then?

2000rosser: My bad 😆

Shpigford:
Starting to wonder about how we're trying to be clever with the use of `constantize` and the account types.

It's causing problems in at least 3 different PRs right now.

Wondering if we should abandon ship on that and use something maybe slightly less succinct.

```
Confidence: High
Category: Remote Code Execution
Check: UnsafeReflection
Message: Unsafe reflection method `constantize` called on parameter value
Code: "Account::#{params[:type]}".constantize
File: app/views/accounts/new.html.erb
```

Shpigford:
This was ultimately addressed in #395

Issue #386, opened 2024-02-08
Account sidebar buildout


Shpigford:
We need to build out the sidebar with a new collapsible/grouped design.

## Scope

1. Accounts should be grouped by their account type
2. Account type groups should be collapsible
3. Each account grouping should have the appropriate "New X account" link at the bottom of it
4. The number for the account should be pulled from that account's `balance` field
5. The account type grouping total balance should come from simply summing the balances of the accounts

## Design

https://www.figma.com/file/PEsvixyFSgKtp069qOzwaD/Sidebar-Accounts?type=design&node-id=1%3A2&mode=dev

<img width="387" alt="Sidebar Accounts" src="https://github.com/maybe-finance/maybe/assets/35243/e31efc5d-cf47-4d40-8170-ff78f4eb15d2">

## Bounty

There is a $100 bounty on this. We award bounties based on first to successfully meet the scope requirements. In the event that multiple PRs are made from different people, we generally go with ones with cleanest code and test coverage (when appropriate).

Shpigford:
Implemented with #391

PR #391, opened 2024-02-08
Build out account sidebar

JoshAntBrown:
I've put this together for the account sidebar. Should be a near 1 to 1 match with Figma since I checked all the dimensions to the tailwind ones or used a custom number where required.

I'm not particularly happy about the way I've queried the accounts within the partial right now, it might be nicer if we could just do `type.all` and have the model scoped by the Current.family automatically but that's a wider problem than the scope of this issue.

/claim #386


app/views/accounts/_account_list.html.erb
```diff
@@ -0,0 +1,32 @@
+<%# locals: (type:) -%>
+
+<% accounts = Current.family.accounts.where(accountable_type: type.name) %>
+
+<div class="text-sm mb-1" data-controller="reveal">
+  <button class="flex gap-4 px-2 py-3 items-center w-full rounded-[10px] font-medium hover:bg-[#f2f2f2]" data-action="reveal#toggle">
```
robzolkos, line 6: @JoshAntBrown any thoughts on using summary/detail element for this?

JoshAntBrown: Good call, will refactor, wanted to get a first version out

JoshAntBrown: Much better with the detail element, no need for the stimulus controller, and toggling between the icons feels more robust being based on the state of the dialog. Thanks!

Shpigford:
@JoshAntBrown Would you mind adding a GIF or short video showing the functionality? Just a super quick little thing to show that it's functioning. 🙂 Much easier to give feedback on that side of things without pulling the code down.

app/views/accounts/_account_list.html.erb
```diff
@@ -0,0 +1,30 @@
+<%# locals: (type:) -%>
+
+<% accounts = Current.family.accounts.where(accountable_type: type.name) %>
+
+<details class="text-sm mb-1 group">
+  <summary class="flex gap-4 px-2 py-3 items-center w-full rounded-[10px] font-medium hover:bg-[#f2f2f2]" data-action="reveal#toggle">
```
robzolkos, line 6: Is this data-action still needed?

Shpigford:
@JoshAntBrown Amazing! Looks good to me. Merging now.

PR #392, opened 2024-02-08
Use account_type name for Add Account modal

### LINKED ISSUE #382: Add account form should output the type of account

nickostendorf:
fixes #382 

image

Just saw #390, but had already started working on this.

@Shpigford what do you think about this approach with the locales?

A little new to ruby, but `t('.'+params[:type]` feels messy. Let me know if there is a better way to implement the translation

### DIFF
--- a/app/views/accounts/new.html.erb
+++ b/app/views/accounts/new.html.erb
@@ -64,7 +64,9 @@
       <%= link_to new_account_path(step: 'method', type: params[:type]), class: "flex w-8 h-8 shrink-0 grow-0 items-center justify-center rounded-lg bg-[#141414]/5" do %>
         <%= lucide_icon('arrow-left', class: 'text-gray-500 w-5 h-5') %>
       <% end %>
-      <span>Add account</span>
+      <% if params[:type].present? && Account.accountable_types.include?("Account::#{params[:type]}") %>
+        <span>Add <%= t('.'+params[:type]) %></span>
+      <% end %>
     </div>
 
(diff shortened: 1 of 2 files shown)


app/views/accounts/new.html.erb
```diff
@@ -64,7 +64,9 @@
       <%= link_to new_account_path(step: 'method', type: params[:type]), class: "flex w-8 h-8 shrink-0 grow-0 items-center justify-center rounded-lg bg-[#141414]/5" do %>
         <%= lucide_icon('arrow-left', class: 'text-gray-500 w-5 h-5') %>
       <% end %>
-      <span>Add account</span>
+      <% if params[:type].present? && Account.accountable_types.include?("Account::#{params[:type]}") %>
+        <span>Add <%= t('.'+params[:type]) %></span>
```
JoshAntBrown, line 68: The whole string within span ought to be captured for internationalisation, and then the model's name passed down as a variable.

You really need access to the constantized account type though, which is where it gets more tricky. That should probably be done in the controller.

Assuming the accountable type was made available you could do something like:

```ruby
t('.new', account_type: @accountable_type.model_name.human.downcase)
```

Then in the locale file:
```
account_type: "New %{account_type}"
```

nickostendorf: Sounds good. Thanks for the explanation!

PR #394, opened 2024-02-08
Account Type using view helpers

### LINKED ISSUE #381: Entry method step not properly setting type
Currently the "entry method" step is hard-coded to type "Depository" when it should detect which type from the `params[:type]`

https://github.com/maybe-finance/maybe/blob/main/app/views/accounts/new.html.erb#L45-L47
```
<%= render "entry_method", type: Account::Depository.new, text: 'Enter account balance manually', icon: "keyboard" %>
<%= render "entry_method", type: Account::Depository.new, text: 'Securely link bank account with data provider (coming soon)', icon: "link-2", disabled: true %>
<%= render "entry_method", type: Account::Depository.new, text: 'Upload spreadsheet (coming soon)', icon: "sheet", disabled: true %>
```

eddbot:
Should help to close #381 and #382 🤔 

https://github.com/maybe-finance/maybe/assets/42304980/0f16cd0d-3bb2-4b2c-9186-de8a86c87949

I see these are quite popular issues to tackle, so will just open this as a draft if anyone wants to steal/borrow my approach 😄

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -2,4 +2,16 @@ module AccountsHelper
   def to_accountable_title(accountable)
     accountable.model_name.human
   end
+
+  def humanized_account
+    account_type.model_name.human
+  end
+
+  def account_type
+    type = params[:type]
+
+    Account unless type.present? && Account.accountable_types.include?(params[:type])
+
+    "Account::#{type}".constantize
+  end
 end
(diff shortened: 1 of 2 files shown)


app/helpers/accounts_helper.rb
```diff
@@ -2,4 +2,31 @@ module AccountsHelper
   def to_accountable_title(accountable)
     accountable.model_name.human
   end
+
+  def humanized_account
+    account_type.model_name.human
+  end
```
eddbot, line 8:
```suggestion
  def humanized_account
    to_accountable_title(account_type)
  end
```

🤔

PR #415, opened 2024-02-09
Currency Dropdown

### LINKED ISSUE #380: Currency drop-down addition
We need some help creating this currency dropdown.

CleanShot 2024-02-08 at 10 55 08@2x

Should use Stimulus + Tailwind to pull it off.

The view for this is `/views/accounts/new.html.erb` and is around that "balance" input field.

crolla97:
closes #380

WIP: Initial styling and interactivity has been setup, just need to figure out how to allow dropdown to overflow the modal and fix placement

- [x] Currency button styles
- [x] Currency dropdown styles
- [x] Update currency label on select
- [x] Highlight active currency inside dropdown
- [x] Fix dropdown modal overflow
- [ ] Fix dropdown screen cutoff


crolla97:
I didn't have enough time today to implement the solution for the screen cutoff issue, so I'm open to someone else taking over this task.

To address the overflow issue, I chose to set the dropdown's position to fixed. This allows it to bypass the modal's overflow cutoff. An alternative approach involves setting the dialog element's overflow to visible. However, this method can cause problems if the modal content exceeds the modal's height, leading to overflow at the bottom. Suggestions are welcome for this solution also!

Current
Screenshot 2024-02-10 at 00 15 29

robzolkos:
I tried adding a fixed height and adding an overflow scroll and that _seems_ ok but not really elegant.

CleanShot 2024-02-09 at 19 48 29

I'm going to research how other apps have attacked this problem..

Shpigford:
FWIW, I don't mind the fixed-height solution at this point. It will default to whatever your account currency is set to and ultimately isn't something most folks will have to interact with all that often.

Certainly would love to come up with a really great solution *long term* but at this stage we probably don't need to spend a massive amount of time on it.

Johnson1s:
581053e52064ad06c2f4a9a4adecff9a9776a331

app/helpers/application_form_builder.rb
```diff
@@ -8,15 +8,17 @@ def initialize(object_name, object, template, options)
 
   (field_helpers - [ :label, :check_box, :radio_button, :fields_for, :fields, :hidden_field, :file_field ]).each do |selector|
     class_eval <<-RUBY_EVAL, __FILE__, __LINE__ + 1
-      def #{selector}(method, options)
+      def #{selector}(method, options, &block)
         default_options = { class: "form-field__input" }
         merged_options = default_options.merge(options)
+        content = block_given? ? @template.capture(&block) : ''
```
JoshAntBrown, line 14: This feels a little off to me, it may be better to just use the helpers directly for a complex field like the currency dropdown. Sometimes a little repetition is okay.

```
<% form_field_tag do %>
  <% label :balance, class: "form-field__label" %>
  <%# inputs here %>
<% end %>
```

This would give you more control over the styling and placement of the contents of the field. The designs indicate at some point we might want to put the currency symbol at the beginning for example.

This would also enable you to wrap the symbol and inputs with a container and style them with flexbox then.

JoshAntBrown (commented):
Minor styling detail, the currency should be in alignment with the input text (have drawn a red line on the original design below to highlight).

Screenshot 2024-02-10 at 11 57 57

app/helpers/forms_helper.rb
```diff
@@ -1,5 +1,6 @@
 module FormsHelper
-  def form_field_tag(&)
-    tag.div class: "form-field", &
+  def form_field_tag(&block)
```
crolla97, line 2: @JoshAntBrown  Might have made things more complicated again as I couldn't use the form_field_tag without added the &block unless I've missed something. Happy to change this in favour of directly using something like: 

`<div class="form-fields">content</div>`

crolla97:
Reverted the change of updating the accounts currency value as this currency [branch](https://github.com/maybe-finance/maybe/tree/feature/currency) looks to add currency support. This is purely a front end update

app/helpers/forms_helper.rb
JoshAntBrown: I was going to ask about this, what was the issue you had?

I've just pulled your branch down to my machine and reverted this back to the original implementation and it looks like it works for me.

crolla97: Ahh yes you're right it works for me reverting now, I was fixing merge conflicts whilst looking at this and couldn't get it to work without explicitly setting the context. Reverting now

app/javascript/controllers/currency_dropdown_controller.js
```diff
@@ -0,0 +1,58 @@
...
+    this.menuTarget.classList.add("hidden");
+  }
+
+  connect() {
+    document.addEventListener("click", this.hideMenu);
+  }
+
+  disconnect() {
+    document.removeEventListener("click", this.hideMenu);
+  }
```
antlio, line 23: what do you think about binding the event to the element instead as follow:

```suggestion
  connect() {
    this.element.addEventListener("click", this.hideMenu.bind(this);
  }

  disconnect() {
    this.element.removeEventListener("click", this.hideMenu.bind(this);
  }
```

however i'm wondering if it would make more sense to add `this.hideMenu()` in the `selectOption` function instead.

Shpigford:
@antlio How do you feel about me going ahead and merging this? I'm comfortable with it not being perfectly optimized as we'll be able to work out kinks as we go.

app/javascript/controllers/currency_dropdown_controller.js
JoshAntBrown: I believe the intention here is to capture clicks that occur outside of the dropdown so that the dropdown becomes closed.

Binding the event to the element wouldn't capture those clicks.

antlio:
@Shpigford absolutely I should have highlight these are non blocking comment 🙏

app/javascript/controllers/currency_dropdown_controller.js
antlio: Makes sense!

I saw that this is done in the js `modal_controller` like this:
```js
click_outside(e) {
    if (e.target === this.element) {
      this.element.close();
    }
  }
 ``` 
 
 stimulus build in events could also be used like:
 ```js
 data-action="click@window->currency-dropdown#hideMenu"
 ```
 
 but already looks good to go as it is!

Issue #295, opened 2024-02-04
Production Dockerfile for Self Hosting


DomiiBunn:
## Planned Work

See https://github.com/maybe-finance/maybe/issues/295#issuecomment-1959490796

- [ ] Build an "all in one" Dockerfile that can run the app on a single VPS
- [ ] Build a Github action that pushes the Dockerfile to ghcr as scaffolded out in both #416 and in #327
- [ ] Implement SSL config change from discussion in #308 

**An update will be posted directly to this issue when it is ready to be worked on!**

## Original Issue 

**Issue Description:**

*Background:*
Currently, it seems that the development is in its early stages. While acknowledging this, I believe that having a public Docker image along with a Docker Compose example could be beneficial for the community.

*Proposal:*
I suggest creating and sharing a public Docker image for easier deployment. Additionally, including a Docker Compose example would facilitate users in setting up the environment seamlessly, potentially reducing future dependency-related issues.

*Benefits:*
1. Simplifies deployment process.
2. Mitigates potential dependency issues.
3. Enhances user experience and accessibility.

*Considerations:*
- Ensure documentation is comprehensive for both the Docker image and Docker Compose setup.
- Set up a CI to deply the Docker image as updates are made.

*Note:*
I understand that this might be premature in the development phase, but planning for this inclusion could prove valuable in the long run.

yeskiy:
Hey! You can use [devcontainer](https://github.com/maybe-finance/maybe/blob/main/.devcontainer/devcontainer.json) for that!

K-J-VV:
Would you consider uploading & releasing images for those less-savvy Githubers to be able to spin our own docker containers?

e.g., ghcr.io/maybe-finance/maybe:latest

DomiiBunn:
My exact point ^

I spun it up with devcontainer but I think people would rather coppy a docker-compose file. change some env vars and do `docker compose up` rather than clonning a reppo and seting up with dev containers.

^.^

K-J-VV:
> My exact point ^
> 
> I spun it up with devcontainer but I think people would rather coppy a docker-compose file. change some env vars and do `docker compose up` rather than clonning a reppo and seting up with dev containers.
> 
> ^.^

I can absolutely vouch as being a person who would rather use a docker-compose file. I consider myself a creative homelaber, not at all a developer/coder. I'm sure I can figure out how to clone a reppo but am entirely dependent on trying out new projects from available images & then updating my tags as new releases are published

[...]

PR #416, opened 2024-02-09
Created docker-compose configuration and tested in local to run as container

jinnabaalu:


hkamran80:
Wouldn't it be better to use a `.env` file that is referenced across each Compose service?

jinnabaalu:
> Wouldn't it be better to use a `.env` file that is referenced across each Compose service?

That is individual choice, rather complicating the compose, i want to keep it plain, when you run in your environment it should be based on their thought to apply.

dodwmd (approved):

config/environments/production.rb
```diff
@@ -43,7 +43,7 @@
   # config.assume_ssl = true
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  config.force_ssl = ENV.fetch("SSL_ENABLE", "true") == "true"
```
aleonmon, line 46: I would suggest using either `SSL_ENABLED` or `ENABLE_SSL` and keeping it consistent for the upcoming env vars

jinnabaalu: You want me to have the variable `ENABLE_SSL`  confused. Please let me know I can update

aleonmon: `ENABLE_SSL` or `SSL_ENABLED` => OK

`SSL_ENABLE ` => NOT OK

VDuda:
Tested this branch in my environment -  ```FATAL:  database "maybe" does not exist``` - I believe the database should be set to `maybe_production`? Correct me if I'm wrong, working through the issue now.

.github/workflows/create-docker-image-push-toghcr.yml
```diff
@@ -0,0 +1,28 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    branches:
+      - main
```
thomas-mc-work, line 6: Wouldn't it suffice to only build and push on creating tags?

zachgoll:
Closing this out for now.  I think these changes belong in a dedicated PR to build out a Dockerfile, address SSL environment variables, and an automated push flow for the Dockerfile.  This is all being tracked in #295

PR #422, opened 2024-02-10
Add UI for Assets Overview

elithecho:
This PR implements the base for the interface shown on the repo's main readme.

Screenshot 2024-02-10 at 3 36 16 PM

The CSS are hardcoded as tailwind JIT cannot pick up dynamic classes.

Question remains whether we want to implement https://github.com/rails/dartsass-rails for managing non-tailwind related classes?

It also does not implement the nicer total display since it's something that could be broken down as a component/template of its own.

Screenshot 2024-02-10 at 3 38 52 PM

## In the case of No assets

Screenshot 2024-02-10 at 3 57 00 PM


app/assets/stylesheets/accounts.css
```diff
@@ -0,0 +1,74 @@
...
+  color: var(--account-loan);
+}
+
+.text-account-other-asset {
+  color: var(--account-other-asset);
+}
+
+.text-account-other-liability {
+  color: var(--account-other-liability);
+}
```
JoshAntBrown, line 74: We should let Tailwind do the heavy lifting here by extending the theme in tailwind.config.js as per their documentation https://tailwindcss.com/docs/customizing-colors#adding-additional-colors.

This will ensure we have utility classes available for all places where colour makes sense.

app/models/asset_group.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  attr_reader :name, :type, :total_asset_value, :percentage_held, :param
+
+  def initialize(type, accounts, percentage_held: nil)
+    @type = type
+    @param = type.model_name.param_key.gsub("_", "-")
+    @total_asset_value = accounts.sum(&:balance)
+    @percentage_held = percentage_held
+  end
+end
```
JoshAntBrown, line 24: I feel like there is some overlap here between this and the accountable types themselves, and I wonder if some of this would be better suited to the Accountable concern. I'm also not sure if it leads to some confusion as we were talking about accounts, but now we're talking about assets?

Is there a distinct difference between an AssetGroup, and a given accountable type like Account::Credit?

Perhaps we could move total_value, and percentage_held into the concern as methods or derive some further concern that allows the retrieval of these values? Maybe AssetGroup is a concern?

I feel like an interface like the following might be preferable:
```
Account::Credit.total_value(user)
Account::Property.percentage_held(user)
```

The user/family argument could default to Current.user, so needn't be passed every time further cleaning the interface. It could also perhaps - at some point - accept a collection so we may be able to query for the total value of all accounts of a given type held in Maybe or for a segment of users/families.

I'd also consider adding a method on the accounts relation to group by type and return the real class for us since it feels like a common thing we need to do.

Could lead to something like this:
```
<% Current.user.accounts.by_type.each do |type| %>
  <%= type.total_value %> (<%= type.percentage_held %>)
<% end %>
```

app/assets/stylesheets/accounts.css
elithecho: hey there, so the problem I've faced with this was when using dynamic classes, tailwind JIT isn't able to track them.
any suggestions on going the tailwind route?

app/models/asset_group.rb
elithecho: I get what you mean with the naming, think `AccountGroup` might be a better naming.
`AccountGroup` is a View model, kind of like a Presenter pattern.

The goal here is to group them, and store information about the group.
eg an AccountGroup model contains information of a particular type like `Account::Credit`

It allows us to extend the information to the next part of the UI.

Screenshot 2024-02-10 at 11 54 41 PM

Apologies if I don't understand your approach, as grouping by type and returning a type will not allow us to aggregate the data of a particular Account type like Account::Credit.

And this pattern would perform an N+1 query since they do not have context of the total account value of the total account value.
```
Account::Credit.total_value(user)
Account::Property.percentage_held(user)
```

elithecho:
@JoshAntBrown I've made some updates based on your suggestions.
I'm not too sure how I'm able to return a single type without aggregating them into a group.
Also, let me know how we can tackle extending Tailwindcss with dynamic attributes

app/models/account.rb
```diff
@@ -6,4 +6,19 @@ class Account < ApplicationRecord
...
+    grouped.map do |accountable_type, accounts|
+      total_value = accounts.sum(&:balance)
+
+      AccountGroup.new(
+        type: accountable_type.constantize,
+        total_value:,
+        percentage_held: (total_value / total_account_value) * 100
+      )
+    end
+  end
```
elithecho, line 23: the reason for using `all` gets all the accounts as object, then using `sum` accounts for balance + their respective currency rather than depending on SQL `sum` function

app/assets/stylesheets/accounts.css
robzolkos: There doesn't seem to be dynamic classes being used in this case.  Everything here could be defined in the tailwind config.

JoshAntBrown: What are you referring to when you say dynamic classes? I don't see anything dynamic about the class names in this approach either, so not sure why tailwind's lack of ability to track dynamic classes would matter.

app/models/asset_group.rb
JoshAntBrown: Yeah good points, I was mostly thinking about the interface rather than how it would be queried underneath. There might be a way to avoid the N+1 problem, but not sure it's worth the added complexity cost.

Thinking of it as a view model makes sense to me. Still not sure on the name though - one thing that came to mind was the concept of a Portfolio that takes all your Accounts, but I'll avoid bike shedding too much.

app/views/pages/_dashboard_assets_overview.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <div class="flex -mx-0.5 my-6">
+      <% @account_groups.each do |account_group| %>
+        <div class="block mx-0.5 h-2.5 rounded bg-<%= account_group.param %>" style="flex-basis: <%= account_group.percentage_held %>%"></div>
+      <% end %>
+    </div>
+
+    <div class="flex justify-center mt-8">
+      <% @account_groups.each do |account_group| %>
+        <div class="flex items-center mx-4">
+          <span class="inline-block rounded-full w-3 h-3 bg-<%= account_group.param %>"></span>
```
elithecho, line 18: @robzolkos @JoshAntBrown 
This is the dynamic class definition I'm talking about.

If we extend tailwind confit with a new color `account-credit`.

Without actually writing `bg-account-credit` in template, tailwind JIT can't find it and will purge the class from the build.

robzolkos: You can use tailwinds [safelist](https://tailwindcss.com/docs/content-configuration#safelisting-classes) feature for this.   I'm just wondering if its premature however as I don't think we're using these colors anywhere else yet (and a solution may become clearer when we do?)

app/models/asset_group.rb
elithecho: Yeah the naming is hard to nail down. Initial thought with AssetGroup was because of the section title, "Total assets".

JoshAntBrown:
> I've made some updates based on your suggestions. I'm not too sure how I'm able to return a single type without aggregating them into a group. Also, let me know how we can tackle extending Tailwindcss with dynamic attributes

Thanks! Your refactor highlighted something to me: it feels like the core of the complexity we're trying to abstract is just the percentage calculation. The sum of the balance is straight forward and wouldn't feel too out of place in the view for me. 

I wonder if we can get away with just having the percentage calculation as a helper that takes either the grouped accounts or the balances themselves?

As I write this I'm also coming around to the idea of AccountGroup, but perhaps as Account::Group, or Account::Set. Perhaps at some point we might want to calculate this information on different groups/sets of accounts. Maybe by type, or by currency, or something else? What percentage of my accounts are in USD?

I think one of my concerns with the AssetGroup was with having too many models at the top level.

app/views/pages/_dashboard_assets_overview.html.erb
elithecho: That makes sense, but yes it's quite premature now to add IMO. But let me know if you'd prefer we go that route.

config/tailwind.config.js
```diff
@@ -18,13 +23,21 @@ module.exports = {
...
         black: '#242629',
         offwhite: '#F9FAFB',
+        'account-depository': '#2E90FA',
+        'account-investment': '#32D583',
+        'account-property': '#F23E94',
+        'account-vehicle': '#6172F3',
+        'account-credit': '#36BFFA',
+        'account-loan': '#F38744',
+        'account-other-asset': '#12B76A',
+        'account-other-liability': '#F04438',
```
zachgoll, line 33: I think we will want to avoid any arbitrary names like this.  Ideally, we'll just have a bunch of standardized color names here in the config.  I think @Shpigford may have a Figma file around this shortly!

elithecho: @Shpigford let me know how we can proceed, especially with the color naming!

zachgoll: @elithecho I'm going to push a change to the Tailwind config here shortly that you can use.

zachgoll: Color palette is now live - https://github.com/maybe-finance/maybe/pull/452

I've updated these lines so once you pull down `main` you can accept the incoming changes.

zachgoll:
Hey @elithecho, just getting caught up with some of the earlier discussions with @JoshAntBrown.  Had a few questions/comments:

1. Does this PR reference an existing issue or Figma design?  Or is it from the README screenshot?
2. Regarding the data modeling, I think I tend to agree with @JoshAntBrown that if possible, we avoid any early abstractions.  The idea of a "view model" probably makes sense at some point, but we still have a lot of designs to work through which may affect these account groupings.  Coming up with the appropriate name will be a lot easier once we see a couple different scenarios where aggregations need to be calculated.

elithecho:
@zachgoll this was done abstracted from the README before we had that figma link! (oops)
Happy to close this if you think it will create any complications.
Just decided to do something to contribute since we did have an interface before that.

zachgoll:
@elithecho gotcha, no problem.  Really appreciate you diving in and tackling things that will move us forward like this!

Unfortunately I think our best path forward here is to close this one out and wait until we release a full feature spec with guidelines on implementation.  There are lots of good learnings from this PR that we can use for that moving forward!

PR #424, opened 2024-02-10
Improve account modal keyboard navigation

sergey-tyan:
Add back button support to account keyboard navigation and autofocus text input on the form
demo video: 
<video src="https://github.com/maybe-finance/maybe/assets/6569308/55b0bae6-9fa3-4616-8be8-27c99ca54281"></video>

### DIFF
--- a/app/javascript/controllers/list_keyboard_navigation_controller.js
+++ b/app/javascript/controllers/list_keyboard_navigation_controller.js
@@ -31,6 +31,11 @@ export default class extends Controller {
   }
 
   get focusableLinks() {
-    return Array.from(this.element.querySelectorAll("a[href]"))
+    const backLink = this.element.parentElement.querySelector("a[href].back")
+    const links = Array.from(this.element.querySelectorAll("a[href]"))
+    if (backLink) {
+      links.unshift(backLink)
+    }
+    return links
   }
 }
--- a/app/javascript/controllers/list_keyboard_navigation_controller.js
+++ b/app/javascript/controllers/list_keyboard_navigation_controller.js
@@ -17,7 +17,10 @@ export default class extends Controller {
 
   getLinkTargetInDirection(direction) {
     const indexOfLastFocus = this.indexOfLastFocus()
-    return this.focusableLinks[indexOfLastFocus + direction]
+    let nextIndex = (indexOfLastFocus + direction) % this.focusableLinks.length
+    if (nextIndex < 0) nextIndex = this.focusableLinks.length - 1
+      
+    return this.focusableLinks[nextIndex]
   }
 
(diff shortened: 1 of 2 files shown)


JoshAntBrown (commented):
It would probably be better to move the controller further up the DOM so that the back link is included in the scope for the query for focusable links instead of changing the javascript controller to handle this edge condition of there being a back link.

app/javascript/controllers/list_keyboard_navigation_controller.js
```diff
@@ -17,7 +17,10 @@ export default class extends Controller {
 
   getLinkTargetInDirection(direction) {
     const indexOfLastFocus = this.indexOfLastFocus()
-    return this.focusableLinks[indexOfLastFocus + direction]
+    let nextIndex = (indexOfLastFocus + direction) % this.focusableLinks.length
+    if (nextIndex < 0) nextIndex = this.focusableLinks.length - 1
```
sergey-tyan, line 21: if last item is focused and you press "down" then next focus will go to the first item

sergey-tyan:
@JoshAntBrown added an optional attribute for passing `data-controller` to the modal, please take a look

app/views/accounts/new.html.erb
```diff
@@ -1,12 +1,12 @@
 <h1 class="text-3xl font-semibold font-display"><%= t('.title')%></h1>
 
-<%= modal do %>
+<% content_for_modal = capture do %>
```
JoshAntBrown, line 3: We should update the modal helper to accept the args, and directly rather than bringing it's complexity into our view.

The modal helper is just a function at the end of the day so can accept any args we want:
```
<%= modal(controllers: "list-keyboard-navigation")
```

That said, if we take this approach, I also feel like it should accept args as if it were any regular tag builder helper. This would add a fair amount of complexity to the helper as we would need to figure out how to merge it's base properties and attributes with those we pass in here.

I think we can take an even simpler approach and just render a div inside the modal that wraps all the content and applies our list-keyboard-navigation controller.

app/javascript/controllers/list_keyboard_navigation_controller.js
JoshAntBrown: Sounds good! Nice improvement.

app/views/accounts/new.html.erb
sergey-tyan: sure, will update the pr, thanks for the feedback!

sergey-tyan: @JoshAntBrown updated, please take a look

Shpigford:
Great job on this, @sergey-tyan!

sergey-tyan:
thank you @Shpigford!

PR #425, opened 2024-02-10
Multi-currency support

Shpigford:
There's so much here that it may genuinely be difficult to actually fully review this (not to mention, it requires a $12/mo Open Exchange Rates account to pull in exchange rate data).

But, putting it here for sanity checks on the code.


zachgoll (approved):
The direction of this makes sense to me.  Eventually it might be nice to abstract away the "original" and "converted" concept away from the views and have a `account.balance` instance method to use, but nothing that can't be improved later!

db/migrate/20240209174912_redo_money_storage.rb
```diff
@@ -0,0 +1,12 @@
+class RedoMoneyStorage < ActiveRecord::Migration[7.2]
+  def change
+    add_column :accounts, :original_balance, :decimal, precision: 19, scale: 4, default: 0.0
+    add_column :accounts, :original_currency, :string, default: "USD"
+    add_column :accounts, :converted_balance, :decimal, precision: 19, scale: 4, default: 0.0
+    add_column :accounts, :converted_currency, :string, default: "USD"
+
+    remove_column :accounts, :balance_cents
+    remove_column :accounts, :balance_currency
+    remove_column :accounts, :currency
```
zachgoll, line 10: Guessing we're cool with dropping existing balances at this stage?

config/solid_queue.yml
```diff
@@ -0,0 +1,18 @@
...
+#       polling_interval: 0.1
+#
+# development:
+#  <<: *default
+#
+# test:
+#  <<: *default
+#
+# production:
+#  <<: *default
```
zachgoll, line 18: Delete?

db/migrate/20240209174912_redo_money_storage.rb
Shpigford: Yeah. Especially since those `balance_` columns were associated with that Money gem. Not worth the hassle.

config/solid_queue.yml
Shpigford: It's all just commented out for now. We'll likely customize some of it as we add more background jobs.

Shpigford:
> The direction of this makes sense to me (GoodJob looks interesting, nice that we can avoid the extra Redis dependency!). Eventually it might be nice to abstract away the "original" and "converted" concept away from the views and have a `account.balance` instance method to use, but nothing that can't be improved later!

I agree that I don't loovvvvve having both, but yeah...can abstract once we get more data coming in for things like transactions. That way we don't prematurely abstract things.

Shpigford:
YOLO. Let's do this. 🫣

Gemfile
```diff
@@ -23,12 +23,15 @@ gem "lucide-rails", github: "maybe-finance/lucide-rails"
...
 
+# Background Jobs
+gem "good_job"
+
 # Other
 gem "bcrypt", "~> 3.1.7"
 gem "inline_svg"
 gem "jbuilder"
 gem "tzinfo-data", platforms: %i[ windows jruby ]
-gem "money-rails", "~> 1.12"
```
sergey-tyan, line 31: could you please explain the reasoning behind not using this gem?

Shpigford: No major benefit to using it.

jhliberty: https://github.com/maybe-finance/maybe/blob/4d5d35b277ce89002f02b3a63d3380abdcc384e7/db/migrate/20240206031739_replace_money_field.rb#L3

This migration fails without it!

adrienpoly: I had this issue, It should be fixed by #438

Issue #429, opened 2024-02-11
Save "collapse" state of account sidebar


Shpigford:
The account groupings in the sidebar can be expanded/collapsed.

We need to save the "collapse" state.

What's up for discussion here is _where_ to save it.

* Database column associated with the user?
* Local storage?
* Something else?

![CleanShot 2024-02-10 at 19 40 49@2x](https://github.com/maybe-finance/maybe/assets/35243/da4c21b3-d64b-4b38-824b-35b6f25e6e88)

## Bounty
There is a $25 bounty on this. We award bounties based on first to successfully meet the scope requirements. In the event that multiple PRs are made from different people, we generally go with ones with cleanest code and test coverage (when appropriate).

We suggest submitting a draft PR as soon as you begin work on it.

ashleypean:
@Shpigford I'm not familiar with this repo's rules around bountied issues. Can I claim this or just put up a PR for the task?

Shpigford:
@ashleypean It's available to everyone! We generally award the bounty to the first person to complete the task. In the event of multiple submissions, we generally go to the most succinct solution.

Shpigford:
@sergey-tyan I've awarded you the bounty!

Thank you to both @sergey-tyan and @ashleypean for working on this!

Went with #432 due to passing code brevity and passing CI.

sergey-tyan:
thank you!

PR #433, opened 2024-02-11
use localstorage to save account sidebar collapsed state

ashleypean:
/claim #429

### DIFF
--- a/app/views/accounts/_account_list.html.erb
+++ b/app/views/accounts/_account_list.html.erb
@@ -3,7 +3,7 @@
 <% accounts = Current.family.accounts.where(accountable_type: type.name) %>
 
 <% if accounts.sum(&:converted_balance) > 0 %>
-  <details class="mb-1 text-sm group" data-controller="account-list" data-account-type="<%= type.name.demodulize%>" data-action="click->account-list#updateDetailState"> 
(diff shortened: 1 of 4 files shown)


app/views/accounts/_account_list.html.erb
```diff
@@ -3,29 +3,29 @@
...
-  <% accounts.each do |account| %>
-    <%= link_to account_path(account), class: "flex items-center w-full gap-3 px-2 py-3 mb-1 hover:bg-[#f2f2f2] rounded-[10px]" do %>
-      <div>
-        <p class="font-medium"><%= account.name %></p>
-        <% if account.subtype %>
-          <p class="text-xs text-[#737373]"><%= account.subtype&.humanize %></p>
-        <% end %>
-      </div>
-      <p class="ml-auto font-medium"><%= format_currency account.converted_balance %></p>
+  <details class="mb-1 text-sm group" data-controller="account-list" data-account-list-category-value="<%= type.name %>" data-action="toggle->account-list#handleToggle">
```
JoshAntBrown, line 6: It might be better to make a more generic stimulus controller that persists the state of any details element rather than specifically just the account list.

Perhaps we could give the details element an id, and use that as it's key?

The name of the action could also be more descriptive, perhaps
`toggle->details-state-controller#persistState`?

ashleypean: Good point, udpated!

Shpigford:
Closing in favor of #432

PR #438, opened 2024-02-11
fix migration and add smoke test for migrations

adrienpoly:
somehow money-rail gem was removed in a previous commit 
as a consequence I wasn't able to run the migration to start the project

~~this PR adds it back~~ and adds a smoke database reset test to ensure a new developper can always run migrations to start fresh on this project

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
   push:
-    branches: [ main ]
+    branches: [main]
 
 jobs:
   scan_ruby:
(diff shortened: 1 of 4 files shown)


JoshAntBrown:
I think it may be better to remove the dependency from the migration than bring back dead dependencies?

adrienpoly:
Sure I don’t have a detailed history of the project looking through the PRs I was under the impression the gem was removed by mistake. If that’s not the case then for sure I would prefer to fix the migration

JoshAntBrown:
Ah yes of course, checking git blame looks like it was removed in this PR: https://github.com/maybe-finance/maybe/pull/425. There's some discussion about this issue there too.

adrienpoly:
Great will update the PR to fix this migration without the gem them

adrienpoly:
Pr updated accordingly. Thanks for pointing me in the good direction

PR #443, opened 2024-02-11
Add more currencies to the seed

adrienpoly:
This PR adds more currencies to the seed so that by default we have 5 currencies.

The currencies are added as test fixtures. The seed use the fixtures to create the record. 

If this approach is of interest, I think this could be extended to other models 

CleanShot 2024-02-11 at 22 39 25@2x


Shpigford:
For testing is one thing, but shouldn't be added to the core seed file as actually using multiple currencies means you need an Open Exchange Rates API key (which costs money).

PR #446, opened 2024-02-12
Alternate static source for exchange rate data

stevegeek:
Thanks so much for starting Maybe as an opensource project!

I see currency support is being added using the `openexchangerate.org` API. However there is potentially a useful free alternative (at least at the moment) which is https://github.com/ismartcoding/currency-api 

This PR is a first pass at using it as a source for fetching currency data. 

Let me know if this is something you might be interested in adding.

### DIFF
--- a/lib/tasks/exchange_rates.rake
+++ b/lib/tasks/exchange_rates.rake
@@ -1,3 +1,5 @@
+require "bigdecimal"
+
 namespace :exchange_rates do
   desc "Fetch exchange rates from openexchangerates.org"
   task sync: :environment do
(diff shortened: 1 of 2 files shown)


lib/tasks/exchange_rates.rake
```diff
@@ -80,4 +82,46 @@ namespace :exchange_rates do
...
+        based_rates = { "USD" => rates_in_usd }
+        rates_in_usd.each do |currency_iso_code, value_in_usd|
+          based_rates[currency_iso_code] = rates_in_usd.transform_values do |other_currency_value|
+            (BigDecimal(other_currency_value, NDIGITS) / BigDecimal(value_in_usd, NDIGITS)).round(PRECISION)
+          end
+        end
+
+        # Now store the exchange rates in the database
+        Currency.all.each do |currency|
+          based_rates[currency.iso_code].each do |currency_iso_code, value|
```
stevegeek, line 115: Actually this should only iterate over the supported currencies

Shpigford:
Thanks for this! Unfortunately pulling from a random repo with very little history is too risky for a financial app like Maybe, so we'll pass on this for now.

stevegeek:
Ok makes sense 👍 ! 

I stuck it in a gist in case anyone comes across this and finds it useful while developing etc : https://gist.github.com/stevegeek/61898dc78857f5adcc0f029f989c9593

PR #450, opened 2024-02-12
Improve active class assignment in sidebar links

mattia-malnis:
I noticed an issue where the active class was not being applied to sidebar links when visiting a nested route, such as `/accounts/:id`:

immagine

To enhance user navigation experience, I think it's a good idea to maintain the sidebar menu item highlighted when its corresponding page or section is active:

immagine

### Changes made

I've created the `is_current_path?` helper method, which utilizes the `recognize_path` method to extract the controller from the given path. This extracted controller is then compared with the current controller to determine if the path corresponds to the current page.


app/helpers/application_helper.rb
```diff
@@ -27,7 +27,7 @@ def sidebar_link_to(name, path, options = {})
     hover_class_names = [ "hover:bg-white", "hover:border-[#141414]/[0.07]", "hover:text-gray-900", "hover:shadow-xs" ]
     current_page_class_names = [ "bg-white", "border-[#141414]/[0.07]", "text-gray-900", "shadow-xs" ]
 
-    link_class_names = if current_page?(path)
+    link_class_names = if is_current_path?(path)
```
zachgoll, line 30: I'm not sure how often we'll be using the helper outside of this specific scenario, so I'm leaning towards a one-liner to handle this:

```
if current_page?(path) || (request.path.start_with?(path) && path != "/")
```

Is there another set of scenarios you're thinking of where we might need a dedicated `is_current_path` helper?

mattia-malnis: Actually, at the moment, I'm not sure in which other scenarios the helper might be used. I created the `is_current_path` helper primarily for potential reuse in cases where there might be pages containing submenus. 
Your solution seems suitable for our current purpose!

mattia-malnis: @zachgoll I've updated the code!

PR #451, opened 2024-02-12
Sidebar link active

cmhhelgeson:
Just adds styling from when the sidebar element is active, making it appear as if it is "pushed in" and signaling to the user that their input was received.


zachgoll:
@cmhhelgeson a few lint errors here that need to be fixed, then will get this merged!

app/helpers/application_helper.rb
```diff
@@ -24,14 +24,17 @@ def currency_dropdown(f: nil, options: [])
 
   def sidebar_link_to(name, path, options = {})
     base_class_names = [ "block", "border", "border-transparent", "rounded-xl", "-ml-2", "p-2", "text-sm", "font-medium", "text-gray-500", "flex", "items-center" ]
+
     hover_class_names = [ "hover:bg-white", "hover:border-alpha-black-50", "hover:text-gray-900", "hover:shadow-xs" ]
     current_page_class_names = [ "bg-white", "border-alpha-black-50", "text-gray-900", "shadow-xs" ]
+    active_class_names = [ "active:bg-gray-200", "active:shadow-xs"]
```
zachgoll, line 30:
```suggestion
    active_class_names = [ "active:bg-gray-200", "active:shadow-xs" ]
```

zachgoll:
@cmhhelgeson just one lint error that needs to be fixed.  Would recommend setting up "format on save" in your code editor to prevent these moving forward.

app/helpers/application_helper.rb
cmhhelgeson: Fixed the issue, but could not get format on save to work correctly. If I may ask, what linter is being used?

app/helpers/application_helper.rb
```diff
@@ -24,14 +24,16 @@ def currency_dropdown(f: nil, options: [])
 
   def sidebar_link_to(name, path, options = {})
     base_class_names = [ "block", "border", "border-transparent", "rounded-xl", "-ml-2", "p-2", "text-sm", "font-medium", "text-gray-500", "flex", "items-center" ]
-    hover_class_names = [ "hover:bg-white", "hover:border-alpha-black-50", "hover:text-gray-900", "hover:shadow-xs" ]
-    current_page_class_names = [ "bg-white", "border-alpha-black-50", "text-gray-900", "shadow-xs" ]
+    hover_class_names = [ "hover:bg-white", "hover:border-[#141414]/[0.07]", "hover:text-gray-900", "hover:shadow-xs" ]
+    active_class_names = [ "active:bg-gray-200", "active:shadow-xs" ]
+    current_page_class_names = [ "bg-white", "border-[#141414]/[0.07]", "text-gray-900", "shadow-xs" ]
```
zachgoll, line 29: Looks like when you merged `main` in, this got reverted back to the old file.

zachgoll:
@cmhhelgeson do you plan to update this?  Or should I close out for now?

PR #459, opened 2024-02-14
Fixed Migration Errors

cmhhelgeson:
When attempting to migrate new model/database code into the schema on a different branch, I encountered errors in redo_money_storage.rb and replace_money_field.rb which I described here:
https://discord.com/channels/1191501936281784361/1194410011863035904/1207137319572742144

This branch implements the changes discussed in the subsequent Discord posts (I've only commented out balance_currency under the assumption that it will be implemented later).

### DIFF
--- a/db/migrate/20240206031739_replace_money_field.rb
+++ b/db/migrate/20240206031739_replace_money_field.rb
@@ -1,7 +1,6 @@
 class ReplaceMoneyField < ActiveRecord::Migration[7.2]
   def change
-    add_column :accounts, :balance_cents
-    change_column :accounts, :balance_cents, :integer, limit: 8
+    add_column :accounts, :balance_cents, :integer, limit: 8
 
     Account.reset_column_information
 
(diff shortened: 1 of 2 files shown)


zachgoll:
@cmhhelgeson for now, I think running `bin/rails db:reset` (which does not execute migrations, but rather loads the DB directly from `schema.rb`) is probably the best way to get things working right now.

I'm thinking we may need to prune some old migrations to make it easier for new contributors.

adrienpoly:
in https://github.com/maybe-finance/maybe/pull/438 I added I smoke test to ensure `rails db:reset` always works. It could be updated with a `rails db:drop db:create db:migrate db:seed` which would run the full circle and ensure we can always run the migration (it would have caught this error).

This being said I think from time to time it is nice to prune older migration by merging all of them in a single migration (basically coping the schema rb in a migration). In our app we typically do this once every 6 months

This assumes that we don't define triggers at the db level as they are not reflected in the schema.rb (we would need to use structure.sql)

zachgoll:
@adrienpoly yep, I think running the entire lifecycle as a smoke test would be good.

In our old app we heavily relied on triggers, functions, and views in PG.  I think this time around we're trying to push ActiveRecord as far as it will go, but my guess is that we'll end up with `schema_format :sql` at some point.

How have you handled the pruning process when triggers/functions/views are defined?

adrienpoly:
> How have you handled the pruning process when triggers/functions/views are defined?

I had none up to recently so that was simple ...

PR #462, opened 2024-02-14
Update database migrations and strengthen db smoke tests

adrienpoly:
Follow up to the talks in #459 

- fix migrations
- strengthen the db smoke test to run the full db:drop db:create db:migrate db:seed steps

Co-authored-by: Christian Helgeson <62450112+cmhhelgeson@users.noreply.github.com>


zachgoll (approved):
Awesome, thank you both for the help on this!

PR #463, opened 2024-02-15
Speedup docker build time

anonychun:
Previously individual layers within the Dockerfile installed the same some dependencies repeatedly, causing redundancy and build time increase.

This change:
Moves commonly used dependencies to the base layer of the Dockerfile.
Other layers now inherit these dependencies from the base layer, avoiding redundant installations.
This approach leverages Docker's layering mechanism to ensure all layers have access to required dependencies while eliminating unnecessary repetition.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,10 @@ FROM registry.docker.com/library/ruby:$RUBY_VERSION-slim as base
 # Rails app lives here
 WORKDIR /rails
 
+# Install base packages
+RUN apt-get update -qq && \
+    apt-get install --no-install-recommends -y curl libvips postgresql-client
+
 # Set production environment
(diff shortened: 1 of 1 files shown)


zachgoll:
@anonychun changes make sense, but were you able to get this Dockerfile to build successfully?  If so, what command are you running? (I was not able to myself)

robzolkos:
Since this is using the stock standard production Dockerfile that comes with Rails - would it be better to have this upstreamed into Rails (where more eyes could get a look at it).  Thoughts @anonychun ?

anonychun:
I already created pull request to Rails, and already being merge https://github.com/rails/rails/pull/51063

robzolkos (approved):

Issue #475, opened 2024-02-22
Implement turbo/broadcast into account pages


Shpigford:
We've just rolled out account balance syncing such that when you add a new entry to your balance, it triggers a background job (`AccountBalanceSync`).

We need to implement Turbo to properly refresh the items on the account page.

Here's a video showing the issue:

https://github.com/maybe-finance/maybe/assets/35243/833278f9-9cd2-45eb-91ff-fb9c7d6cc8bd

## Scope
1. Should show "syncing" message
2. Once the background job finishes it should remove the sync message and update the graph

Presumably using something like `broadcast_update_to` + `turbo_stream_from` is the solution here, but that's your call on the best way to implement. We're using Turbo 8 so if there's any new fancy pants magic in there for this sort of thing, feel free to use it.

## Bounty
There is a $50 bounty on this. To participate in the bounty:

1. Comment `/attempt #447` to indicate you plan on working on this.
2. Submit a draft PR once you start and include `/claim #447` along with your general game plan for how you intent to implement this.
3. Change out of draft mode on the PR once you're ready to submit your attempt.
4. Include a short video showing functionality in your PR (drastically speeds up us approving).

In the event of multiple attempts, we generally pick the one that most closes matches the scope with the cleanest code and test coverage (when appropriate).

alexandredelaitre:
/attempt #475 I will be create an attempt to solve this. When the add/update button is pressed I will display the syncing message, once the frontend receives a signal from the backend I will remove the syncing message and update the graph, either using broadcast_update_to + turbo_stream_from or some sort of Promise.

TheVixhal:
Is this issue still open?

PR #476, opened 2024-02-23
Refresh account on update

JoshAntBrown:
/claim #475

Uses the new refresh feature as part of Turbo 8 to broadcast refreshes and morph page updates into the page, avoiding the need to manually refresh the page, or have to build manual updates to parts of the page.

Video attached below, I did the whole setup as demonstrated in the original issue so the main change features about half-way through.

https://github.com/maybe-finance/maybe/assets/1793797/9f324c44-dba4-4a44-bb72-c4f053a1cc14

### DIFF
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -1,6 +1,5 @@
 development:
-  adapter: redis
-  url: redis://localhost:6379/1
+  adapter: postgresql
 
 test:
   adapter: test


app/models/account.rb
```diff
@@ -1,4 +1,5 @@
 class Account < ApplicationRecord
+  broadcasts_refreshes
```
robzolkos, line 2: ❤️

robzolkos (approved):

zachgoll (commented):
Nice solution here!  Just a few small questions/comments:

1. State after broadcast

Small nitpick, but any chance we could get this to reset to the default "closed" state after the refresh broadcast?

CleanShot 2024-02-23 at 08 34 58

2. ActionCable Adapter

To keep things simple for local development, I was thinking we'd use the `async` adapter for Action Cable, but I realized that `turbo-rails` [deliberately updates this setting](https://github.com/hotwired/turbo-rails/commit/29fe2cf14b0055804fcd426aadc7d79e67286635) to Redis.  I'm still getting up to speed with the Rails/Turbo ecosystem, so wondering if there was a rationale behind this published somewhere?

If Redis is a hard requirement, probably worth updating the README and I can get the Wiki setup guides updated as well.

zachgoll:
@JoshAntBrown awesome, thanks for the fix on that!

And regarding Postgres, I'm kind of sitting in the camp of, "let's see how far we can get with it".  That way, we avoid the Redis dep for now and only add it if Postgres causes significant problems.  @Shpigford any preference here?

Shpigford:
@zachgoll My hunch is we're gonna end up using Redis regardless (even if just as a cache), but I'm fine with a limited bit of exploration. I don't think it's worth going to great lengths to avoid it, though.

JoshAntBrown:
I've updated the adapter for development, it just uses the Active Record's connection pool so nothing else to configure. I think for caching you can probably use Solid Cache if you want to avoid Redis completely.

JoshAntBrown:
Btw, I think we can probably further simplify how the `valuation#create` action works.

We could remove [valuations/create.turbo_stream.erb](https://github.com/maybe-finance/maybe/blob/7e324f1b53f38f669761c9ee169224a9b21fb83a/app/views/valuations/create.turbo_stream.erb#L1-L4) and replace it with a simple redirect. This would help reduce a chunk of complexity since we can just have the `accounts#show` page render the current state and then have Turbo refresh it either as part of the redirect or by broadcasting it's changes.

The only blocker to doing that right now is how to render the 'Syncing' message. The `account#show` view uses `@account.status == "SYNCING"` to determine whether the account is currently syncing. However, this flag isn't set until the job starts and the job completes almost immediately. Therefore, the user would never see this message.

To work around this, we could update the status on the account as part of the `valuation#create` action. I feel like this might be the wrong approach though, and just adds complexity elsewhere.

Instead, it might be better if we had a way to query any pending or active sync jobs. This could perhaps be a status on each valuation that gets updated once the valuation has synced, or another model that tracks syncing jobs. It feels a bit beyond the scope of this PR though, especially since a refactoring of the syncing mentioned in #474, so perhaps something to consider at a later date?

zachgoll:
@JoshAntBrown I agree that the valuation CRUD flows could be simplified.  You're more than welcome to move things around if you see a clear path with that.  Most of what is on `main` is a "sketch" that we'll need to tidy up once all the core stuff is in there, so improvements are definitely welcome!

Regarding the sync status message—I think what we have here is okay for now even though the user will likely just see a "flash" due to how fast the job is.  That message is mostly a UI placeholder until the official designs are completed for how to show these "sync" states.

JoshAntBrown:
Thanks, and yeah makes sense. I'll have a think about it and may follow up with a separate PR at some point if I can think of a nice way of tidying up some of the sync management. Hopefully what we have should be good enough for now.

zachgoll:
@JoshAntBrown sounds great!  I have a PR coming shortly that adds a barebones `Transaction` model and updates the sync job to incorporate transactions.  Once that's in, I think we'll start to get a much clearer picture of how this sync job, `Account`, `Valuation`, `Transaction`, and `AccountBalance` should all be working together.

zachgoll (approved):

PR #479, opened 2024-02-24
docs: add dev containers config gotcha in README

nshki:
# Overview

This is a small README update for devs who choose to use dev containers to work on the project. When setting this up on my machine (Fedora Asahi Linux running on an M2 MacBook Air) via the [dev containers CLI](https://github.com/devcontainers/cli), I ran into an issue where the Rails server kept throwing a `could not connect to server: Cannot assign requested address` error. The issue was that the `.env` file that I copied from `.env.example` was trying to connect to `DB_HOST=localhost` when it needed to be pointed at `DB_HOST=db` in the Docker Compose setup.

This adds a sentence in the README to address this and it also adds a comment in `.env.example` next to the relevant line.

And as a side note, I'm very excited that you all have decided to open source this lovely project. I'm looking forward to keeping up with its developments and would love to contribute when I can.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,6 +25,6 @@ SMTP_PASSWORD=
 TLS=true
 
 # Database Configuration
-DB_HOST=localhost
+DB_HOST=localhost # May need to be changed to `DB_HOST=db` if using devcontainer
 POSTGRES_PASSWORD=postgres
 POSTGRES_USER=postgres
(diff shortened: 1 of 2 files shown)


zachgoll (approved):
Good idea, this is always a tricky thing to remember when setting up with Docker!

PR #481, opened 2024-02-25
Add currency to account form to support multiple currencies

sriram2520:
Fixing some minor bugs which caused issues when using a currency other than USD

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -90,32 +90,8 @@ module ApplicationHelper
   def format_currency(number, options = {})
     user_currency_preference = Current.family.try(:currency) || "USD"
 
-    case user_currency_preference
-    when "USD"
-      options.reverse_merge!(unit: "$", precision: 2, delimiter: ",", separator: ".")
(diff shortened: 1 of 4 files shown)


zachgoll:
@sriram2520 I think we may want to consider moving all of these pre-defined currency configurations to `config/locales` and directly integrate into the Rails I18n API.  There's a good amount of work to do around that, so it might be worth opening an issue to define the scope.  Here are a few starting tasks that I can think of:

- Set up routing for locales (likely want to go with the `/[locale]/[...rest]` convention (e.g. `/en/accounts` or `es/accounts`)
- Write logic for setting user's preferred locale (via `Family`) and a fallback locale
- Define the currencies in `config/locales/defaults`
- Use `number_to_currency` throughout app

zachgoll:
@sriramsfs I think the "Discussions" tab here on the repo is probably the best place for getting feedback on new features, system design, etc.  Once you open up a discussion, it may also be helpful to post a link to it in the "Contributors" channel of the Maybe Discord to get more eyes on it.

PR #487, opened 2024-02-26
Find account first and build transaction through association

dwightwatson:
`find` should throw `ActiveRecord::RecordNotFound` if it can't find the record, so there's no need to manually raise.

You can then simply build the transaction through that association.

(It may make sense to nest this endpoint like a Rails resource... `/accounts/123/transaction` and access the account ID that way instead of passing it through as a param).

### DIFF
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -17,12 +17,9 @@ class TransactionsController < ApplicationController
   end
 
   def create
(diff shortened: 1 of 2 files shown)


dwightwatson:
Looks like the test is flaky because of `Transaction.last` not returning the last transaction consistently. I updated the test to give the new transaction a name by which it can be looked up.

zachgoll (approved):
Looks great, thanks for the contribution!

PR #488, opened 2024-02-26
Update hard-coded currency UI with currency specific params

sriram2520:


app/views/accounts/show.html.erb
```diff
@@ -29,9 +29,9 @@
         <p class="text-sm text-gray-500">Total Value</p>
         <%# TODO: Will need a better way to split a formatted monetary value into these 3 parts %>
         <p class="text-gray-900">
-          <span class="text-gray-500"><%= number_to_currency(@account.original_balance)[0] %></span>
-          <span class="text-xl font-medium"><%= number_with_delimiter(@account.original_balance.round) %></span>
-          <span class="text-gray-500">.<%= number_to_currency(@account.original_balance, precision: 2)[-2, 2] %></span>
+          <span class="text-gray-500"><%= @account.original_currency.unit %></span>
+          <span class="text-xl font-medium"><%= format_currency(@account.original_balance, precision: 0, unit: '') %></span>
+          <span class="text-gray-500"><%= @account.original_currency.separator %><%= @account.original_balance.cents %></span>
```
zachgoll, line 34: If the currency is something like `INR`, won't this default to a precision of 2 still?

zachgoll (approved):
I think overall this looks good.  Will have to be careful how much monkey patching we end up doing, but these are pretty core concepts to the app so makes sense for the time being until we make that final decision on how to handle currencies throughout the app.

app/views/accounts/show.html.erb
sriram2520: You are right, have fixed it now 👍

PR #498, opened 2024-02-28
Show calendar when clicking anywhere on the input field

sriram2520:
Currently datepicker opens only when clicking on the calendar icon. This click event handler will open the calendar when clicking anywhere on the input field.


zachgoll (commented):
I think we should generally avoid overriding browser default behavior on inputs like this unless we have a strong reason to do so.  Is this primarily for convenience?

sriram2520:
@zachgoll Yes, its just for convenience. Typically I do this in most of my forms. Also usually I do onfocus event to show picker as well. This helps when someone uses keyboard tab key to navigate form elements.

Shpigford:
This makes a strong assumption that people _always_ want to use the date picker, which I don't believe is the case.

Unless there's stronger reasoning other than convenience, I'd pass on this.

PR #499, opened 2024-02-28
Hide the additional arrow shown on safari browser for summary tag

sriram2520:
Default browser arrow gets added on safari browser.
Screenshot 2024-02-28 at 5 14 04 PM

Fixing it based on [this](https://stackoverflow.com/a/66814239) stackoverflow answer

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1,9 +1,1 @@
 /* Application styles */
-details > summary {
-  list-style: none;
-}
-
-details > summary::marker, /* Latest Chrome, Edge, Firefox */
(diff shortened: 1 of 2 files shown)


zachgoll:
@sriram2520 good catch, can we throw the following line directly in `application.tailwind.css` in the `base` layer for organizational purposes?

```
@layer base {
  details > summary::-webkit-details-marker {
    @apply hidden;
  }
}
```

zachgoll:
@sriram2520 also, looks like the email on your Github account does not match the email setup with your `git config --global user.email`, causing multiple users on each PR.  Would you mind either adding that email to your Github -> Settings -> Emails, or updating your local git utility?

PR #500, opened 2024-02-28
Show currency placeholders with user family currency settings

sriram2520:


app/views/accounts/new.html.erb
```diff
@@ -71,12 +71,12 @@
           <%= render "accounts/#{permitted_accountable_partial(@account.accountable_type)}", f: f %>
           <%= form_field_tag do %>
             <%= f.label :balance, class: "form-field__label" %>
-            <%= f.number_field :balance, step: :any, placeholder: number_to_currency(0), in: 0.00..100000000.00, required: 'required', class: 'form-field__input max-w-[80%]' %>
+            <%= f.number_field :balance, step: :any, placeholder: format_currency(0), in: 0.00..100000000.00, required: 'required', class: 'form-field__input max-w-[80%]' %>
```
zachgoll, line 74: I agree that this should be dynamic, but rather than the family currency, shouldn't it reflect what has been selected in the currency dropdown below?

sriram2520: Yes, that would be better. But as far as I know we can't do it without javascript/stimulus controller update. 
It feels a bit of an overkill at the moment.

zachgoll: Yep, probably a bit overkill for now.  I think we we're going for any sort of change at this point, it should probably be to _remove_ the currency symbol altogether and just use a `0.00` placeholder.  That way we're being clear and avoiding the extra JS.

PR #501, opened 2024-02-28
Basic Account Balance Sync Algorithm

zachgoll:
This PR introduces a basic account sync flow and algorithm.  The classes and code organization may change in the future, but this lays the foundation for the overall "sync strategy" that populates the `account_balances` table for an account and allows us to show a historical balance graph.

## Strategy

The basic idea is that a series of `AccountBalance` records for an `Account` can be _derived_ from the following inputs:

1. `account.balance` - the current day balance of the account, provided by the user (or a 3rd party data provider like Plaid)
2. Transactions - financial events that alter the account balance
3. Valuations - point-in-time balances which can be thought of as "version controlled balances"

The general logic is:

1. Find the oldest possible balance of the account, which is equal to the oldest `Valuation` or `Transaction` record date
2. Derive the start balance by summing all transactions and adding to current day account balance (or if `Valuation` is the first record, use its value instead)
3. Walk sequentially from the start date to the current day, adjusting the balance based on new `Valuation` or `Transaction` records

## Example

Let's say we have an `Account`:

- Current day `balance` is `5000` (assume current date is `2024-01-06`)
- Oldest transaction is on `2024-01-02`
- Net transaction flows from Jan 2nd -> Jan 6 is -1000
  - Inflows are negative, outflows are positive
  - In this case, `-1000` means "account increased by 1000 over the period"

We can derive the start balance (Jan 1st) to be equal to `current_balance + sum(txns)`, or `5000 + -1000 = 4000`.

In other words, the account starts at 4000 and increases by 1000 over the period to end at the current balance of 5000.

| Date     | Valuation | Transaction | Balance |
|----------|-----------|-------------|---------|
| 1/1/2024|                   |                     | 4000   |
| 1/2/2024|                  | -500             | 4500   |
| 1/3/2024|                |                       | 4500   |
| 1/4/2024 |              |      -500                 | 5000   |
| 1/5/2024 |                 |                    | 5000   |
| 1/6/2024 |                |                     | 5000   |

Now, consider the user creates a manual "adjustment" on 1/3/2024 and assigns the value of the account on this date to 5000:

| Date     | Valuation | Transaction | Balance |
|----------|-----------|-------------|---------|
| 1/1/2024|                   |                     | 4000   |
| 1/2/2024|                  | -500             | 4500   |
| 1/3/2024|      5000          |                       | 5000   |
| 1/4/2024 |              |      -500                 | 5500   |
| 1/5/2024 |                 |                    | 5500   |
| 1/6/2024 |                |                     | 5000   |

In this case, the user-defined valuation alters the balance from the valuation date forward until it reaches the final date, whereby the balance goes back to `account.balance` (source of truth).

If the user wants a smooth balance graph without this sudden change at the end, they will need to update the account balance to match the most recent valuation.

## Future improvements

- Will need to support partial syncs (only sync from the date of the last change)
- Will need to incorporate various transaction types (i.e. investment transactions)
- Will need to handle accounts that have transactions in multiple currencies (current implementation of `transactions.sum` won't work in these cases)


app/models/account_syncer.rb
```diff
@@ -0,0 +1,53 @@
+class AccountSyncer
```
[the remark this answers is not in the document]
zachgoll: I think for now I'll throw it in `app/models/account`, but I do think there could be value in a `syncable` concern eventually as it would be nice to be able to use `family.sync`, `user.sync`, and `account.sync`.  In other words, `account.sync` would sync the account, while `family.sync` would sync all accounts owned by the family (i.e. global syncing)

app/jobs/account_sync_job.rb
```diff
@@ -0,0 +1,16 @@
...
+    account.update!(status: "SYNCING")
+
+    begin
+      AccountSyncer.new(account).sync
+      account.update!(status: "OK")
+    rescue => e
+      account.update!(status: "ERROR")
+      Rails.logger.error("Failed to sync account #{account_id}: #{e.message}")
+    end
+  end
```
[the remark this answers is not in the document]
zachgoll: Yeah I agree, that's a lot cleaner.

app/models/account/syncer.rb
```diff
@@ -0,0 +1,53 @@
...
+    private
+        # TODO: Support partial syncs via start_date
+        def sync_balances(start_date = nil)
+          sync_start_date = [ start_date, @account.effective_start_date ].compact.max
+
+          valuations = @account.valuations.where("date >= ?", sync_start_date).order(:date).select(:date, :value)
+          transactions = @account.transactions.where("date > ?", sync_start_date).order(:date).select(:date, :amount)
+          oldest_entry = [ valuations.first, transactions.first ].compact.min_by(&:date)
+
+          net_transaction_flows = transactions.sum(&:amount)
```
sriram2520, line 22: since there is a currency attribute to transaction model, is it possible that we are adding amounts in mutiple currencies together?

zachgoll: @sriram2520 yes, very good point—this PR assumes a single currency for an account.  

I'm going to add this point to the list of "Future improvements" (in PR description) as I think that would be a good isolated PR to open and nail down how we handle the conversion of "non-standard" transactions for an account.

rangerscience:
I recommend:

1) Consolidating all the mutations "in one place" - so, either the `Account::Syncer#sync` handles the `update!(status: "SYNCING")`, _or_, `Account#sync` does the `upsert_all` (and the syncer returns the `balances_to_upsert array)`. I recommend putting the mutations in the model, but either is fine from my perspective.

2) If you split up the body of the `sync_balances` function so that (in this case) pretty much every assignment is a function (`def sync_start_date; [ start_date, @account.effective_start_date ].compact.max; end` - you get some neat potential benefits:
a) It's really easy to develop and debug (through the console) - `syncer = Account::Syncer.new(self)`, and then poke around; ie  `syncer.sync_start_date`
b) It's really easy to test, since you get tiny little tests for each

app/models/account/syncer.rb
```diff
@@ -0,0 +1,53 @@
...
+    def sync
+        sync_balances
+        purge_balances
+
+        Rails.logger.info("Synced account #{@account.name}")
+    end
+
+    private
+        # TODO: Support partial syncs via start_date
+        def sync_balances(start_date = nil)
```
rangerscience, line 15: Name suggestion: `balances_by_day`

app/models/account/syncer.rb
sriram2520: @zachgoll I can't think of a scenario where the transaction currency would be different from the account's currency. My suggestion would be to remove currency attribute from transaction model, as I dont see it being used anyway.

zachgoll: @sriram2520 multi-currency accounts aren't super common, but they do exist and have enough adoption for us to support them (think Wise, Revolut)—https://www.nerdwallet.com/article/banking/multi-currency-account

PR #504, opened 2024-03-01
Add missing classification local when using account_valuation_list view

jakubkottnauer:
There was a missing `locals` when referencing the `accounts/account_valuation_list` partial view, causing a crash whenever a new valuation history entry was created in the UI.


zachgoll (approved):
Good catch!

PR #520, opened 2024-03-06
Use our .ruby-version file in Bundler so it is managed in only 1 place

corsonknowles:
This provides a single source of truth and simplifies staying current on Ruby.

Background writeup: https://gorails.com/episodes/bundler-ruby-version-file
Original issue: https://github.com/rubygems/rubygems/issues/6742
Implementation: https://github.com/rubygems/rubygems/pull/6876

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby "3.3.0"
+ruby file: ".ruby-version"
 
 # Rails
 gem "rails", github: "rails/rails", branch: "main"


robzolkos:
This change will break the Dockerfile as the .ruby-version file isn't copied over when bundler is run.  (I think we reverted this change due to this a while back). Rails main has removed the ruby version from the Gemfile completely last I checked (a couple of weeks ago). Anyway you may get some issues raised with Docker no longer working for some peeps.

corsonknowles:
Thanks @robzolkos This is a widely used and much beloved new feature of Bundler. I was about to say I'll take a look at what is idiosyncratic about our Dockerfile tomorrow, but from the helpful description you provided, I think we can simply do this:
* https://github.com/maybe-finance/maybe/pull/523

PR #521, opened 2024-03-06
Dashboard View and Calculations

zachgoll:
A rough sketch of the views and backend queries for the main dashboard.

Adds:

- Net worth chart
- D3 trendline controller
- Assets trendline
- Liabilities trendline
- Accounts by group query + table
  - Accounts by group trend + allocation
  - Accounts within group trend + allocation


app/models/account.rb
```diff
@@ -11,27 +11,110 @@ class Account < ApplicationRecord
...
+                end_balance: account.end_balance,
+                allocation: (account.end_balance / total_liabilities * 100).round(2),
+                trend: Trend.new(current: account.end_balance, previous: account.start_balance, type: "liability")
+              }
+            end
+          }
+        end
+      }
+    }
   end
```
zachgoll, line 107: Plenty of refactoring / organizing to do here still—I decided to leave in its current state so that a separate PR can be opened to specifically address a refactor.

PR #523, opened 2024-03-07
Keep .ruby_version file in Docker

corsonknowles:
Addresses:
* https://github.com/maybe-finance/maybe/pull/520#issuecomment-1982466829


Dockerfile
```diff
@@ -25,7 +25,7 @@ FROM base as build
 RUN apt-get install --no-install-recommends -y build-essential git libpq-dev pkg-config
 
 # Install application gems
-COPY Gemfile Gemfile.lock ./
+COPY Gemfile Gemfile.lock ./ .ruby-version
```
zachgoll, line 28:
```suggestion
COPY .ruby-version Gemfile Gemfile.lock ./
```

zachgoll (changes requested):
Good catch @robzolkos, looks like this is the default for `rails new` now?

Dockerfile
zachgoll: The order of the `COPY` command matters here—all the source files must come before the destination, `./`

robzolkos:
> Good catch @robzolkos, looks like this is the default for `rails new` now?

No `rails new` will spit out unchanged [Dockerfile](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt#L45)

This functionality was [reverted](https://github.com/rails/rails/commit/707b5cb60616c86cd554ada2ef8754dd0ab5e72a) in Rails main.  I suspect largely due to it being a generator.

I think copying over the `.ruby-version` prior to the Gemfile should work fine.  Maybe is already generated so little risk in smoothing over these edges.

zachgoll:
> > Good catch @robzolkos, looks like this is the default for `rails new` now?
> 
> No `rails new` will spit out unchanged [Dockerfile](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt#L45)
> 
> This functionality was [reverted](https://github.com/rails/rails/commit/707b5cb60616c86cd554ada2ef8754dd0ab5e72a) in Rails main. I suspect largely due to it being a generator.
> 
> I think copying over the `.ruby-version` prior to the Gemfile should work fine. Maybe is already generated so little risk in smoothing over these edges.

Ahh, hadn't seen that commit.  Sounds good, and we can revisit this with more scrutiny later when implementing #295

PR #524, opened 2024-03-07
Add backend support for transaction categories

jakubkottnauer:
This PR adds basic support for transaction categories in the backend, PR for the UI coming up next. As we've agreed on previously:

- when a family is created, we seed its categories. Doing this allows for localization and user customization of all categories
- each category has an `internal_category` that in the future might be used by the rule engine to discern which category represents what
  - the `internal_category` field is automatically cleared when a user modifies a category's name as it's original purpose cannot be guaranteed anymore

First time I'm working with Ruby or Rails so please don't hesitate to let me know if there's anything odd 😄


app/controllers/registrations_controller.rb
```diff
@@ -13,6 +13,7 @@ def create
     @user.family = family
 
     if @user.save
+      Transaction::Category.create_default_categories(@user.family)
```
jakubkottnauer, line 16: I have considered moving this into an `after_create` callback in the Family model but thought it felt strange having the Family model create other model's instances as a side effect, so I made it explicit like this.

db/migrate/20240306082827_create_transaction_categories.rb
```diff
@@ -0,0 +1,15 @@
+class CreateTransactionCategories < ActiveRecord::Migration[7.2]
+  def change
+    create_table :transaction_categories, id: :uuid do |t|
+      t.string "name", null: false
+      t.string "icon", null: false
```
zachgoll, line 5: I just checked with our designer and found out we _won't_ be using icons for categories.  So we can remove this.

Here's what it will look like:

CleanShot 2024-03-07 at 10 18 37

db/migrate/20240306082827_create_transaction_categories.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string "color", null: false
```
zachgoll, line 6: Might be useful to default this to something like `#6172F3`?

(I didn't use gray because that will be used for "Uncategorized")

zachgoll (approved):
Nice work, changes here make sense to me!

Let me know what UI related changes you'll be making in the next PR so I can provide the relevant designs.

PR #540, opened 2024-03-11
Fix event handler removal on disconnect

jakubkottnauer:
Calling `.bind(this)` always returns a new function, meaning that event handlers registered in `connect` callbacks  were never properly removed in `disconnect` callbacks. Creating event handlers as arrow functions fixed this and also makes the code simpler.

### DIFF
--- a/app/javascript/controllers/account_collapse_controller.js
+++ b/app/javascript/controllers/account_collapse_controller.js
@@ -3,21 +3,19 @@ import { Controller } from "@hotwired/stimulus"
 // Connects to data-controller="account-collapse"
 export default class extends Controller {
   static values = { type: String }
-  boundOnToggle = null
   initialToggle = false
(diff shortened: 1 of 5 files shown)


zachgoll (commented):
The code does look nicer, but this introduces some bugs, likely related to Turbo as shown in the video below.  In order for tab content to show, a full browser refresh is required:

https://github.com/maybe-finance/maybe/assets/16676157/55c9aa27-74db-4ba9-ba7d-2a36201dcc19

jakubkottnauer:
@zachgoll Well spotted! I have pushed a fix, the issue was that the `updateClasses` method in `tab-controller` was receiving the turbo event as the first argument (but it expected the tab to select).

zachgoll:
@jakubkottnauer awesome, looks good now!

PR #541, opened 2024-03-11
Add inline category selection

### LINKED ISSUE #526: Add inline category selection
This issue will allow a user to select or change the category of a transaction, inline.

CleanShot 2024-03-10 at 17 44 07

## Scope

- User should see category in transactions view
- User should be able to click on a category to open a dropdown
- Selected category should have checkmark
- If transaction category is `nil`, should show "Uncategorized" 

### Not in scope

We can add a `cursor-not-allowed` for each of these:

- Search bar is a placeholder
- "Add new" is a placeholder
- "Edit" ellipses is a placeholder

Each of these will likely warrant an entire PR that I'll open up when this issue is completed.

### Design file

Here is the relevant frame for this issue:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-Design-(Community)?type=design&node-id=2003-697&mode=design&t=NUcLxlwf8nbQWWgh-0

jakubkottnauer:
Closes #526

This is what it looks like:

Screenshot 2024-03-14 at 10 04 23

Some things we have discussed and resolved in the comments below, keeping them around for context:

## Problem 1
~~Currently, the entire transaction row is wrapped inside an `<a>` which opens the edit sidebar. It's [against spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) for anchors to contain other interactive elements (such as the transaction category dropdown now). Not only does this cause issues with a11y but it also creates a bug where the anchor captures the click when you click empty space in the transaction category dropdown.~~

~~I've spend an unreasonable amount of time trying different `e.preventDefault` and `e.stopPropagation` tricks to work around this but it's been to no avail. Perhaps rethinking the UX here a little bit would be the best solution, but if anybody wants to try to fix the issue that would be great as well (honestly I'm out of ideas). It might be possible to move the category dropdown contents outside of the scope of the anchor but I haven't been able to do that due to how Stimulus works.~~

## Problem 2
~~By default Safari hides scrollbars unless you're actively scrolling, even when you've used `overflow-scroll` (this depends on your macOS settings). This can be seen in the screenshot I've attached where you can't really tell that the content of the dropdown is scrollable. Either we could create [a custom scrollbar](https://stackoverflow.com/questions/7492062/css-overflow-scroll-always-show-vertical-scroll-bar) or rethink the design to better indicate that the content is scrollable.~~


zachgoll (commented):
Great work here, overall code looks good!

To follow up on your comments:

1. I agree, we definitely shouldn't have interactive content inside an anchor.  Let me do some thinking on this and I'll post a follow-up in a bit.  The simple solution would be to add a little edit icon to the far right of the transaction and use that as the anchor that opens the transaction detail editor.  The more complex solution would be to write a `transaction_row_controller.js` Stimulus controller that listened for clicks on various areas of the row and determined which events to trigger or which links to programmatically click.
2. The designs do have a small custom scrollbar that we could implement (scoped to this specific dropdown) here, but I don't think this is a huge deal if you want to leave as-is.  Up to you.

zachgoll:
@jakubkottnauer I think the simplest solution we can do for now is move the anchor tag that opens the transactions detail view to the transaction title (rather than as the parent container).

We can use a `hover:underline hover:text-gray-800` (or something similar) 

That work for you?  Should remove a ton of complexity

jakubkottnauer:
Great, I agree that's easiest for now. I'll do that and add the custom sidebar tomorrow 👍

jakubkottnauer:
@zachgoll Done, ready for final review

zachgoll (approved):

PR #543, opened 2024-03-11
Multi-Currency Part 2

zachgoll:
## Overview

While the current app is built to support multi-currency, there are a few critical flows that still need to be implemented to support multi-currency.

This PR aims to introduce "v1" multi-currency support with:

- `Money` and `Money::Currency` classes (similar to `money` gem, but much simpler implementation tailored to our use case with major currency units and BigDecimal DB types)
- Support for all major currencies (see `Money::Currency.all`)
- Adds both a `EUR` and `USD/EUR` (multicurrency) account to both demo user and test fixtures
- Updates the account sync process to normalize historical account balances to a common currency defined by `Family.currency`
- Handles an "opt-in" exchange rates setup flow with Synth APIs for users who have accounts in multiple currencies
- Offers "on-demand" exchange rate syncs from Synth APIs

## Remaining tasks

- [x] Add exchange rates for fixtures
- [x] Update account sync to convert balances to common `Family.currency`
- [x] Update tests to expect multi-currency results
- [x] Update all places in UI where multi-currency applies


zachgoll:
Still a good number of follow-up PRs required beyond this one, but the core sync process for handling multi-currency accounts along with the associated tests are in.

zachgoll:
Will revisit this after #553 is merged (lots of code overlap)

PR #545, opened 2024-03-14
Add presence validations for required fields

corsonknowles:
In this PR:

* Adds presence validations for the non-nullable fields on models
* Delete the `credit` model which was identified by @tabertantrum as legacy cruft

Not in this PR:
* This PR does not validate the `family` which is non-nullable for `user`, because adding it breaks the basic test setup. This implies the test setup should be improved, which is out of scope for this PR.

Motivations:
1. Validating required fields before saving is more performant than raising database errors (in non-bulk operations). 
2. This is the first step in improving UX by raising proper errors for required fields. 
3. Combining this with surfacing the flash will provide UX guidance and prevent dead end errors in the app like this:

Screenshot 2024-03-15 at 10 21 21 AM

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -1,6 +1,8 @@
 class Account < ApplicationRecord
   include Syncable
 
+  validates :family_id, presence: true
+
   broadcasts_refreshes
   belongs_to :family
   has_many :balances, class_name: "AccountBalance"
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -1,7 +1,7 @@
 class Account < ApplicationRecord
   include Syncable
 
-  validates :family_id, presence: true
+  validates :family, presence: true
 
   broadcasts_refreshes
   belongs_to :family
--- a/app/models/account_balance.rb
+++ b/app/models/account_balance.rb
@@ -1,6 +1,8 @@
 class AccountBalance < ApplicationRecord
   belongs_to :account
 
+  validates :account_id, :date, :balance, presence: true
+
   scope :in_period, ->(period) { period.date_range.nil? ? all : where(date: period.date_range) }
 
   def self.to_series(account, period = Period.all)
--- a/app/models/account_balance.rb
+++ b/app/models/account_balance.rb
@@ -1,7 +1,7 @@
 class AccountBalance < ApplicationRecord
   belongs_to :account
 
-  validates :account_id, :date, :balance, presence: true
+  validates :account, :date, :balance, presence: true
 
   scope :in_period, ->(period) { period.date_range.nil? ? all : where(date: period.date_range) }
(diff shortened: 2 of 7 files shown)


app/models/credit.rb
tabertantrum, line 1: I believe this model can be removed. It appears to be an orphaned model created before Delegated Types were implemented on Account.

corsonknowles: Awesome, great find Ryan!

app/models/account_balance.rb
```diff
@@ -1,6 +1,8 @@
 class AccountBalance < ApplicationRecord
   belongs_to :account
 
+  validates :account_id, :date, :balance, presence: true
```
zachgoll, line 4: In this case, I think we should remove the `account_id` validation.  We'll often be doing bulk operations on this table and validating the association (`:account`) adds a DB query that will already be enforced at the DB level but will make these bulk operations less performant

app/models/account.rb
```diff
@@ -1,6 +1,8 @@
 class Account < ApplicationRecord
   include Syncable
 
+  validates :family_id, presence: true
```
zachgoll, line 4: When validating associations, you need to pass `validates :family, presence: true` so that not only does the validation ensure the presence of a foreign key, but also that the associated record exists in the database.

zachgoll (commented):
Good idea to add some validations to models here.

Just a few points of feedback on my end:

1. `.keep` files - I don't think deleting these is necessary as they allow the folders to be checked into source control.  We will want newcomers to the project to be able to see the default Rails directories even if there is nothing in them at the moment just from a convention standpoint.
2. Validating associations - left a few comments, but when validating associations, we'll want to validate more than just the ID being present, which introduces an additional DB query.  Not always a bad thing, but on certain models that we'll be doing lots of bulk operations on, it introduces some performance overhead.  These foreign keys are validated at the DB level, so we can be confident that they maintain referential integrity that way.

app/models/account_balance.rb
corsonknowles: In the case of bulk operations, I would recommend passing `validate: false` to the mutative db methods during save or update, which skips validations. Alternatively, and somewhat more dangerously you could use `upsert` or other direct `INSERT` methods which avoid validations and model instantiation entirely. 

But I'm happy to make this change to help make this mergable.

zachgoll: Yep, those are possible workarounds.  Up to you—if keeping it, just make sure to update to `:account` to get the full validation check.

corsonknowles: Thanks Zach! If you are happy either way on validating the `:account` for now, then I think all comments are addressed.

PR #546, opened 2024-03-15
Removes unused model and routes

tabertantrum:
The credit.rb model appears to have been created when delegated types were being introduced to Account. This PR removes this unused file, as well as some unnecessary routes that are now handled through delegated typing on the Account resource.


zachgoll (approved):
Nice catch, looks good

PR #548, opened 2024-03-18
Use before_create instead of before_validation in InviteCode model

yuler:
I think using `before_create` should be more appropriate here

I'm new to rails, if I missed something please tell me

### DIFF
--- a/app/models/invite_code.rb
+++ b/app/models/invite_code.rb
@@ -1,5 +1,5 @@
 class InviteCode < ApplicationRecord
-  before_validation :generate_token, on: :create
+  before_create :generate_token
 
   class << self
     def claim!(token)


Shpigford:
@yuler What's the problem this is solving? What makes it "more appropriate"?

yuler:
Hi, @Shpigford

I think

The `generate_token` method looks like it just generates a random token and doesn't need to be executed before validation.

Replace with `before_create` to make the code look concise and more easily understandable

henrikbjorn:
You can also just change it to `has_secure_token :token, length: 4` and then remove the most of the code in this file

yuler:
I find the MINIMUM_TOKEN_LENGTH = 24 

https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activerecord/lib/active_record/secure_token.rb#L7

henrikbjorn:
> I find the MINIMUM_TOKEN_LENGTH = 24
> 
> https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activerecord/lib/active_record/secure_token.rb#L7

The method have a options argument, where you can change the length https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activerecord/lib/active_record/secure_token.rb#L46

yuler:
Got it thanks

Shpigford:
I appreciate the input on this. Ultimately this particular feature (the invite code) is temporary and not something we're putting more time/effort in to at the moment, so going to close this for now.

yuler:
> > I find the MINIMUM_TOKEN_LENGTH = 24
> > https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activerecord/lib/active_record/secure_token.rb#L7
> 
> The method have a options argument, where you can change the length https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activerecord/lib/active_record/secure_token.rb#L46

Sorry to bother you

I think we got it wrong

The validate code is in

https://github.com/rails/rails/blob/6f0d1ad14b92b9f5906e44740fce8b4f1c7075dc/activerecord/lib/active_record/secure_token.rb#L47

I did a simple try

image

PR #555, opened 2024-03-19
feat: add crypto account type to demo data

jbrodriguez:
This adds a Crypto asset account type to the demo data, to more accurately resemble the main screenshot on the repo

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2024_03_09_180636) do
+ActiveRecord::Schema[7.2].define(version: 2024_03_19_154732) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
@@ -31,6 +31,11 @@ ActiveRecord::Schema[7.2].define(version: 2024_03_09_180636) do
   end
 
   create_table "account_credits", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  create_table "account_cryptos", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
   end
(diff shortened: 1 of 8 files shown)


lib/tasks/demo_data.rake
```diff
@@ -219,6 +219,23 @@ namespace :demo_data do
...
+      a.accountable = Account::Crypto.new
+      a.balance = 10000
+      end
+
+    crypto_valuations = [
+    { date: 1.year.ago.to_date, value: 9000 },
+    { date: 200.days.ago.to_date, value: 9500 },
+    { date: 100.days.ago.to_date, value: 9444.96 },
+    { date: 20.days.ago.to_date, value: 10000 }
+    ]
```
zachgoll, line 233: For a sample account like Bitcoin, we'll want to make sure and specify the currency, `BTC` and adjust these values to be smaller so that when it is converted to the family's currency (likely USD), it will represent a reasonable proportion of the portfolio.  I'd say somewhere around `0.1` is probably appropriate for a demo account (~$6,500 worth of BTC)

db/schema.rb
```diff
@@ -82,7 +87,7 @@
     t.string "currency", default: "USD"
     t.decimal "converted_balance", precision: 19, scale: 4, default: "0.0"
     t.string "converted_currency", default: "USD"
-    t.virtual "classification", type: :string, as: "\nCASE\n    WHEN ((accountable_type)::text = ANY ((ARRAY['Account::Loan'::character varying, 'Account::Credit'::character varying, 'Account::OtherLiability'::character varying])::text[])) THEN 'liability'::text\n    ELSE 'asset'::text\nEND", stored: true
+    t.virtual "classification", type: :string, as: "\nCASE\n    WHEN ((accountable_type)::text = ANY (ARRAY[('Account::Loan'::character varying)::text, ('Account::Credit'::character varying)::text, ('Account::OtherLiability'::character varying)::text])) THEN 'liability'::text\n    ELSE 'asset'::text\nEND", stored: true
```
zachgoll, line 90: Was this change a result of the migration?  Or a manual change?

test/fixtures/account/cryptos.yml
```diff
@@ -0,0 +1,11 @@
...
+
+# This model initially had no columns defined. If you add columns to the
+# model remove the "{}" from the fixture names and add the columns immediately
+# below each fixture, per the syntax in the comments below
+#
+one: {}
+# column: value
+#
+two: {}
+# column: value
```
zachgoll, line 11: These are the cause of the CI test failure.  Since an `Accountable` must always be associated with an `Account`, these create orphaned records that break the account rollups.

jbrodriguez:
@zachgoll i reverted the schema change, comment out the data for the test to run and changed the currency to BTC

however, the dashboard shows the account in USD, thought it had to do with multi-currency, so registered with synth, but there's no BTC "currency" there

Screenshot 2024-03-19 at 12 53 13

i'll dig a bit more ... the idea is to show each account in its base currency or convert everything to USD ?

zachgoll:
@jbrodriguez yep, that's expected right now.  I've got a bunch of local changes for #543 that will be pushed soon to fix that and make all of this work correctly though!

When that goes in, each account will display its own currency (in this case, number of bitcoins) and then in the sidebar and dashboard, the values will be exchanged (via Synth) and "rolled up" into the `Family`'s preferred currency.

PR #557, opened 2024-03-20
Fix ZeroDivisionError

JoshAntBrown:
It's been a little while since I've had time to check in on this project, I pulled the latest changes and ran the migration and ran into a ZeroDivisionError.

I didn't have any liabilities set so the sum was equal to 0.

This change fixes the error so that page can render:
Screenshot 2024-03-20 at 16 44 14

### DIFF
--- a/app/models/value_group.rb
+++ b/app/models/value_group.rb
@@ -40,7 +40,8 @@
     end
 
     def percent_of_total
-        return 100 if parent.nil?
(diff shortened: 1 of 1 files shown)


zachgoll:
@JoshAntBrown thanks for the fix, I'll go ahead and merge this in to get things working and I opened up #558 to map out a long-term solution.  If nobody grabs that, I'll circle back to it next week after I'm done with multi-currency stuff.

PR #561, opened 2024-03-24
Implement swappable provider infrastructure

josefarias:
This PR serves as a proposal for a generic data provider system that allows for swapping interchangeable providers.

It shouldn't be merged as-is; as it implements merchant and real estate valuation providers in an effort to show how this solution scales.

The above comment alone is smelly, as is justifiably called out in https://github.com/maybe-finance/maybe/pull/556#issuecomment-2016916740. But I'm presenting this solution anyway as a response to the request for comments on that PR and the accompanying Discord message. I think this direction is safe to pursue if the requirements are well known and understood internally — and we're definitely implementing multiple providers in the short term. 

Now, on to the PR...

### Mental model

Providers are implemented as a `Provider::<Name>` class. The following providers exist in this PR: `Provider::Base`, `Provider::Synth`, `Provider::Zillow`, `Provider::Null`.

`Provider::Base` should never be instantiated. It is meant as an abstract class, through which all provider classes respond to all provider methods. This is handy in case a user tries to use e.g. zillow as a merchant data provider. I think this is necessary because, while some providers are highly specialized, others provide a wide range of data. It's easy to lose track of which is which for the uninitiated.

`Provider::Null` is the provider you use when the user hasn't configured a provider. It's designed to blow up and communicate the service degradation to the user. I think this is better than implementing e.g. `Provider::Local`, since I don't think it's viable to mimic provider functionality offline.

Each provider class should be thought of as a bridge between Maybe and the providing service. Http calls made to each provider live inside these clases. Rich domain-agnostic objects are built and returned when the http calls conclude.

The domain-agnostic objects are then handed over to concrete domain-aware models implemented as `Provided::<Concept>` classes. The following provided concepts exist in this PR: `Provided::ExchangeRate`, `Provided::MerchantData`, `Provided::RealEstateValuation`.

These domain-aware models know how to transform the results of the http calls into domain objects. They also know when they should retry http calls. They're where the external service ends and Maybe begins.

These `Provided::<Concept>` classes are meant to be used wherever they're needed — class methods, instance methods, scripts, etc. The `Provided::` prefix communicates the external nature of their behavior.

### How it works

The chosen providers and their API keys are configured as environment variables, which are read from `data_providers.yml`. Implementing users only need to change their env vars to change providers.

Every `Provided::<Concept>` class implements a `Providable` concern — which knows which `Provider::<Name>` class to instantiate depending on `data_providers.yml`.

Finally, `Provided::<Concept>` classes are instantiated and called wherever needed — and their results are used and interpreted as necessary.

### Enforcing a contract between providers

As we add more providers, it becomes hard to test all possible combinations. If a common provider's interface drifts away from a less common provider's — it can be weeks until we get a report from customers saying their setups stopped working.

This is why this PR implements interface testing. By including an interface test in `Provider::<Name>Test`, you can guarantee that the provider returns an object of the expected shape. If it stops doing so, the test will fail — alerting the developer that their changes either need to be backwards-compatible, or they need to make the change for all providers of the same concept.

### Before we merge

The first step is deciding whether this is the path we want to go down. If it is, the next step is reverting the illustrative aspects of this PR — the bits that have to do with merchants and real estate valuations. Finally, I'd suggest merging the remaining code in a new PR, post-review.


app/models/exchange_rate.rb
```diff
@@ -2,41 +2,26 @@ class ExchangeRate < ApplicationRecord
   validates :base_currency, :converted_currency, presence: true
 
   class << self
-    def convert(from, to, amount)
```
josefarias, line 5: Seems like we're not using this

app/models/merchant.rb
```diff
@@ -0,0 +1,4 @@
+class Merchant
```
josefarias, line 1: Added so tests don't fail. This would be an ActiveRecord in practice.

test/lib/money_test.rb
```diff
@@ -92,8 +92,4 @@ class MoneyTest < ActiveSupport::TestCase
         er = exchange_rates(:today_usd_to_eur)
         assert_equal Money.new(1000).exchange_to(:eur), Money.new(1000 * er.rate, :eur)
     end
-
-    test "returns nil if exchange rate not available" do
```
josefarias, line 96: I think this test is stale

test/models/account_test.rb
```diff
@@ -103,11 +103,4 @@ def setup
     # Synced series will always have final balance equal to the current account balance
     assert_equal @account.balance_money, @account.series.last.value
   end
-
-  test "generates empty series for foreign currency if no exchange rate" do
```
josefarias, line 107: I think this test is stale, too

test/models/provider/null_test.rb
```diff
@@ -0,0 +1,11 @@
+require "test_helper"
+
+class Provider::NullTest < ActiveSupport::TestCase
+  include ExchangeRateProviderInterfaceTest
+  include MerchantDataProviderInterfaceTest
+  include RealEstateValuationsProviderInterfaceTest
```
josefarias, line 6: As much as the dynamically typed nature of ruby gains us, this is one area where type safety would've come in handy. These tests are a good middle ground I think.

I usually wouldn't implement them. But the project being open-source means many hands with varying degrees of familiarity will be touching this system. And I don't think providers will be added often after an initial flurry. So it's easy to forget things.

josefarias (commented):
This is ready for specific feedback.

app/models/provided/exchange_rate.rb
```diff
@@ -0,0 +1,22 @@
+class Provided::ExchangeRate
+  include Providable, Retryable
+
+  def initialize
+    @provider = exchange_rates_provider
+  end
+
+  def fetch(from:, to:, date:)
+    retrying Provider::Base.known_transient_errors do
```
josefarias, line 9: Haven't given any thought to how rate limiting comes into play. I think `Provided::<Concept>` classes can be aware of how many requests have been performed via `Providable` and modify their actions accordingly.

test/models/provider/null_test.rb
rangerscience: Have you played around much with the (newish) type systems in Ruby?

app/models/provided/exchange_rate.rb
```diff
@@ -0,0 +1,22 @@
+class Provided::ExchangeRate
+  include Providable, Retryable
+
+  def initialize
+    @provider = exchange_rates_provider
```
rangerscience, line 5: Hmm. If we unroll this - 

```
  def exchange_rates_provider
    self.class.exchange_rates_provider
  end
```

```
    def exchange_rates_provider
      provider_for :exchange_rates
    end
```

```
      def provider_for(concept)
        "Provider::#{provider_name(concept).camelize}"
          .constantize
          .new(provider_api_key(concept))
      end
```

```
      def provider_name(concept)
        providers_config[concept][:provider].presence_in(KNOWN_PROVIDERS) ||
          raise(ArgumentError, "Unknown provider for #{concept}")
      end
```

```
shared:
  exchange_rates:
    provider: <%= ENV.fetch "EXCHANGE_RATES_PROVIDER", "synth" %>
    key: <%= ENV.fetch "EXCHANGE_RATES_API_KEY", ENV["SYNTH_API_KEY"] %>
```

So all this... looks like it evaluates to `@provider = Providers::Synth.new`?

rangerscience (commented):
edit (complete comment rewrite): goddamnit nerd sniped myself on this :)

Okay. TL;DR, I think your approach here is overcomplicated things. Seems like you've got a few goals/needs/etc:

1. You want an implementation pattern that's easy for other people to follow to make new providers
2. You want a usage pattern so consumers of the code have a reliable, common abstraction, and don't have to worry about the provider's implementation details
3. You want easy configuration via environment variables so that it's trivial to pick which provider you're using

Here's some quick drafts:

```
#config/initializers/exchange_rates.rb

name = ENV.fetch("EXCHANGE_RATES_PROVIDER") || "synth"
Providers.exchange_rate = case 
when 'synth', 'zillow'
  "Providers::#{name.titleize}".constantize.new
else
  raise(ArgumentError, "Unknown provider for exchange rates")
end
```

Right? One place with a _very_ clear link between configuration value and results. Doesn't worry about how each provider is implemented, all it does it choose which one is in use.

```
class Provider::Synth
  def initialize(api_key=null)
    @api_key = api_key || ENV["SYNTH_API_KEY"]
  end

  def request url
    Faraday.get("#{base_url}/rates/historical") do |req|
        req.headers["Authorization"] = "Bearer #{api_key}"
        req.params = yield
     end
  end
end
```

Right? The provider is fully and solely responsible for knowing how to instantiate itself... but we can also override that behavior, if we want to, so we _can_ have different instances if we find ourselves wanting to. We can also pull out anything common to every request; this is just one of many ways to do that (and, honestly, one I'd like to explore further).

Now, I'm also getting overly fancy with the `yield`, and that might not work out, but it's fun for now so eh :)

For errors and retry logic, [Faraday has a bunch of things you can use](https://github.com/lostisland/awesome-faraday) :)

So now we get extremely concise API wrappers:

```
class Provider::Synth
  def request_exchange_rate(from:, to:, date:)
     request("#{base_url}/rates/historical") { { date: date.to_s, from: to: } }
   end
 end
```

But, this on it's own doesn't accomplish your #2. We can add a bit:

```
class Provider::Synth
  def request_exchange_rate(from:, to:, date:)
      request("#{base_url}/rates/historical") { { date: date.to_s, from: to: } }
  end
   
  def to_exchange_rate response, from:, to:, date:
      ExchangeRate.new \
        base_currency: from,
        converted_currency: to,
        rate: JSON.parse(response.body).dig("data", "rates", to),
        date: date
  end
     
  def get_exhange_rate(...)
    to_exchange_rate(request_exchange_rate(...))
  end
end
```

So now we've got parts. One part that knows how to get the data from Synth, and one part that knows how to turn the data we get from Synth into the data we want to use, and one that sticks them together. Plus, each part is testable independent of each other part, and, we've got a couple different levels we can 

PS - `...` is valid code; it's the ["pass-through"](https://sikac.hu/writing-a-pass-through-method-ruby-2-7-edition-a336239f2a61) language feature :)

Final note - I really don't mean to be "here's exactly how to do it", I mean it much more like "here's some inspiration for some very simple ways to approach this". The `yield` might not work out at all, and `to_exchange_rate` stuff might get wordy and overly boiler-plate. But, I think all this avoids locking future-you into any decisions but the critical ones (how it gets used). On the implementation side of things, all of this should be really flexible and easy to change.

test/test_helper.rb
```diff
@@ -1,6 +1,20 @@
...
+VCR.configure do |config|
+  config.cassette_library_dir = "test/vcr_cassettes"
+  config.hook_into :webmock
+  config.ignore_localhost = true
+
+  config.filter_sensitive_data("<SYNTH_API_KEY>") { ENV["SYNTH_API_KEY"] }
+  config.filter_sensitive_data("<EXCHANGE_RATES_API_KEY>") { ENV["EXCHANGE_RATES_API_KEY"] }
+  config.filter_sensitive_data("<MERCHANT_DATA_API_KEY>") { ENV["MERCHANT_DATA_API_KEY"] }
+  config.filter_sensitive_data("<REAL_ESTATE_VALUATIONS_API_KEY>") { ENV["REAL_ESTATE_VALUATIONS_API_KEY"] }
+end
```
rangerscience, line 17: VCR is amazing :) I've got a couple of utility functions you might find handy, let me get back to you tomorrow...

config/data_providers.yml
```diff
@@ -0,0 +1,10 @@
+shared:
+  exchange_rates:
+    provider: <%= ENV.fetch "EXCHANGE_RATES_PROVIDER", "synth" %>
+    key: <%= ENV.fetch "EXCHANGE_RATES_API_KEY", ENV["SYNTH_API_KEY"] %>
+  merchant_data:
+    provider: <%= ENV.fetch "MERCHANT_DATA_PROVIDER", "synth" %>
+    key: <%= ENV.fetch "MERCHANT_DATA_API_KEY", ENV["SYNTH_API_KEY"] %>
+  reale_state_valuations:
+    provider: <%= ENV.fetch "REAL_ESTATE_VALUATIONS_PROVIDER", "null" %>
+    key: <%= ENV["REAL_ESTATE_VALUATIONS_API_KEY"] %>
```
rangerscience, line 10: I think I'd recommend something like this, instead:

```
providers:
  exhange_rates: <%= ENV.fetch "EXCHANGE_RATES_PROVIDER", "synth" %>
synth:
  key: ENV["SYNTH_API_KEY"]
```

Mostly, this lets each provider define it's own instantiation, but still lets your user define which provider they want to use. 

For example, I think it's Google Cloud whose library wants a JSON file of credentials, instead of an API key.

Like - hmm. Your abstraction _isn't_ "all providers are the same", it's, "all providers can produce the same info". So you don't need the configuration of each provider to be identical.

lib/retryable.rb
```diff
@@ -0,0 +1,17 @@
+module Retryable
```
rangerscience, line 1: Faraday has a pretty sophisticated retry package: https://github.com/lostisland/faraday-retry

app/models/provider/synth.rb
```diff
@@ -0,0 +1,59 @@
...
+    @api_key = api_key
+  end
+
+  def fetch_exchange_rate(from:, to:, date:)
+    response = Faraday.get("#{base_url}/rates/historical") do |req|
+      req.headers["Authorization"] = "Bearer #{api_key}"
+      req.params["date"] = date.to_s
+      req.params["from"] = from
+      req.params["to"] = to
+    end
```
rangerscience, line 12: hmm I think you can have som fun like:

```
def request url
  response = Faraday.get("#{base_url}/rates/historical") do |req|
      req.headers["Authorization"] = "Bearer #{api_key}"
      req.params = yield
   end 
   
   # ...
end

def fetch_exchange_rate(from:, to:, date:)
  request("#{base_url}/rates/historical") do
    { date: date.to_s, from: to: }
  end
end
```

app/models/provider/synth.rb
```diff
@@ -0,0 +1,59 @@
...
+      ExchangeRateResponse.new \
+        rate: JSON.parse(response.body).dig("data", "rates", to),
+        raw_response: response
+    else
+      raise Provider::Base::ProviderError, <<~ERROR
+        Failed to fetch exchange rate from #{self.class}
+          Status: #{response.status}
+          Response: #{response.body.inspect}
+      ERROR
+    end
```
rangerscience, line 24: Why not just return the rate - what do you gain from returning a `ExchangeRateResponse`?

app/models/provided/exchange_rate.rb
josefarias: That eliminates the two outermost layers of abstraction only, which are the thinnest wrappers.

Those were implemented on purpose. I think it's better for the caller to only know the name of a method rather than the name pf a module, a method, and a slightly magical symbol argument.

Personally not worried about those two thin wrappers. They're meant to provide a purposeful public API.

config/data_providers.yml
josefarias: We don't need the configuration of each provider to be identical. But it's rather nice that it is.

I can see some value in adding the provider names as keys here, instead of generic "CONCEPT_PROVIDER_API_KEY". Kind of annoying that the example file would get bigger with each provider.

I originally had even removed SYNTH_API_KEY, but left it in as a fallback because CI would break otherwise. I think if we want to remove the synth one we'd have to add the others as secrets to GH cc @zachgoll up to you!

lib/retryable.rb
josefarias: Ah didn't know about this. I could go either way. This is a pretty simple implementation.

Unless that gem can effectively handle rate limiting out of the box the way we need it. In that case I'd probably go with the gem.

app/models/provider/synth.rb
josefarias: Yeah I've done this kinda thing in HTTP adapters before. It's dry-er. But I always ended up regretting it. It's easier to come back to all the request details being colocated.

josefarias: Until they get out of hand that is. This kind of cleanup is always a small effort away if needed.

josefarias:
Thanks for your comments @rangerscience! Responded to some of them.

This proposal prioritizes clear extension points, colocation of behavior, and minimal configuration.

I can appreciate the desire for less layers. But, as presented, I think that would be to the detriment of the colocation and clear delineation of responsibilities.

Less code is nice. But a clear mental model is nicer.

Debatable whether this mental model is actually clear. It is to me. But happy to have the maintainers disagree. All good!

test/models/provider/null_test.rb
josefarias: I haven't! Considered those for this. But I wouldn't recommend we adopt them for the whole codebase. Seems worse to use them for a subset of the app than to implement these tests which are a simple `include` away.

josefarias:
@zachgoll one thing we could do to avoid premature abstraction (which depending on internal conversations I think this PR might do) is remove everything swapping-related and just always instantiate `Provider::Synth` within `Providable`. I'd also remove the interface tests in that case. Would be left with a pretty lean PR and we can keep this as reference for when we do want to add other providers.

lib/retryable.rb
rangerscience: The less you do, the less you have to maintain :)

app/models/provider/synth.rb
rangerscience: Fair! I would at least reduce the boiler plate with:

```
req.params = {
   date: 
}
```

Just makes it a smidge easier to modify (and read, I'd say).

config/data_providers.yml
josefarias: I don't think we need to solve for this right now. Any non-standard behavior can be addressed in the concrete provider initializer. Also not sure data providers would require much more than an API key usually — I don't anticipate Google Cloud or similar being a data provider in this capacity.

Might push up a commit where the provider classes know about their provider-specific env vars. Would certainly be nice to be able to instantiate them without further config. Will revisit after work.

rangerscience:
> Finally, I don't think what you're proposing paints a full picture. 

Heh :) I mean, I nerd sniped myself at 3:00 AM, so yeah, doesn't paint a full picture :) 

I think the most critical bit I'm adding is pulling apart the code that chooses _which_ provider gets used, from the code that defines those providers.

> It doesn't consider implications of instantiating providers for more than one concept. 

I don't think I follow? If you have more concepts, you just have more `Providers.concept` initializers. If you have more than one provider for a given concept, you just wire them into those initializers; and, if you've got common code, pull it out into (as you propose) `Provider::Concept` base classes (or any of the other ways to handle that in Ruby). There just isn't any common to pull out yet.

I'd toss in: What happens when you have a provider that provides a unique mix of multiple concepts? Like, if Zillow does both exchange rates and housing prices, while Wall Street Bets does exchange rates and stock prices?

> And the obviation of the `Provided::<Concept>` classes will lead to each provider implementing the same logic to instantiate models outside their domain.

Don't need it yet, and you add it when you've got common code to start pulling out. If you want to do something resembling interface definitions, I'd do it in test code rather than inheritance stuff - Ruby's a duck-typed language, so it... "fits" a bit better to write up tests that say "in order to be an X, it has to pass these tests", than to define and interface you inherit from.

> There might be something to not needing to pass the API key to the provider class, though.

Yah! I mean, for most things that's what you want, but, how many times have you run across an API that tries to get weird with it (looking at you, GCP, with your credentials _file_)?

josefarias:
> I don't think I follow? 

We're on the same page. I meant your proposal didn't include that code where you add more concepts, ultimately calling for an abstraction. I just went directly to that abstraction because this PR is about sketching out a multi-concept, multi-provider, scaled, solution. That's what was asked for per the learnings in Maybe 1.0. This is not meant to be merged as-is — as articulated in the PR description.

> What happens when you have a provider that provides a unique mix of multiple concepts? 

I think that's very reasonably addressed in this PR. Synth, Null, and Zillow provide a mix of multiple concepts. Some shared, some not.

---

@zachgoll thanks for the comments! I'll look into those after work. That plan sounds perfect. I'll push up some final changes and post some notes for future consideration — will then close this and implement a pared down, only-what-we-need-right-now version in a separate PR as described in our comments above. Which I think is what @rangerscience is advocating for.

rangerscience:
Pleasure discussing all this with y'all, and thanks for my first real experience participating in open source :)

> separate PR

I might chuck in a link to these PRs in code comments, so that it's easy for someone to find all this additional context :)

app/models/provided/exchange_rate.rb
```diff
@@ -0,0 +1,22 @@
...
+    retrying Provider::Base.known_transient_errors do
+      response = provider.fetch_exchange_rate(from:, to:, date:)
+
+      ExchangeRate.new \
+        base_currency: from,
+        converted_currency: to,
+        rate: response.rate,
+        date: date
+    end
+  end
```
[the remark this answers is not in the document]
josefarias: You know? Come to think of it, I think this abstraction might be anemic.

I think we can just call the provider and instantiate a record from within the model. Would probably live within its own concern.

This would happen inside `ExchangeRate.fetch_from_provider` and `ExchangeRate.fetch_historical_from_provider` methods.

I'll spike on it tonight and see where it gets us. Will probably move the retryable stuff to the provider instead.

app/models/exchange_rate.rb
```diff
@@ -2,41 +2,26 @@ class ExchangeRate < ApplicationRecord
...
-    def get_rate(from, to, date)
-      _from = Money::Currency.new(from)
-      _to = Money::Currency.new(to)
-      find_by! base_currency: _from.iso_code, converted_currency: _to.iso_code, date: date
+    def find_rate_or_fetch(from:, to:, date:)
+      find_rate! from:, to:, date:
     rescue
-      logger.warn "Exchange rate not found for #{_from.iso_code} to #{_to.iso_code} on #{date}"
-      nil
+      fetch_rate_from_provider(from:, to:, date:).tap(&:save!)
```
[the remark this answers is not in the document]
josefarias: Yes. I think that makes perfect sense. Definitely smelly that we'd write to the database when fetching something. And if done in an HTTP GET request, that's even worse! Great callout.

config/data_providers.yml
josefarias: https://github.com/maybe-finance/maybe/pull/561/commits/ef7ced2e4c4fb752425e219a717ac90ae4424c1b

app/models/provider/synth.rb
josefarias: Moved some things around in https://github.com/maybe-finance/maybe/pull/561/commits/c49782ebfab9289bfd871c4616bd64e48fa063b8 and these classes ended up arguably messier.

I think that's fine. It's all self contained and easily refactored as patterns emerge within each provider.

Gut tells me to avoid sharing too much code between providers. Yes, they're all providers in the realm of Maybe. But they're each their own thing and have their own lifecycles. Probably best not to tie them together?

app/models/provided/exchange_rate.rb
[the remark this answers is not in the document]
josefarias: https://github.com/maybe-finance/maybe/pull/561/commits/c49782ebfab9289bfd871c4616bd64e48fa063b8

josefarias:
I've moved some things around. Left a summary in the commit message but re-sharing here for convenience:

```
Nix anemic abstractions

* The `Provided::<Concept>` classes were anemic. It's fine for the model to know
how to call a provider that provides it directly. It's also better to have the model instantiate
domain objects. And, finally, it makes more sense for the class making
the HTTP requests to be aware of retries.

* The base implementations for provider methods were moved from `Provider::Base` to
`Provider::Null`. They were part of the base so that every provider responds to every
providable method, which is unnecessary. Let's be less defensive about it.
Things will likely break with NoMethodError anyway in the edge case
this was trying to guard against.

* Previously, when HTTP providers like Synth failed, they would raise.
Now they return an inspectable response object of the same shape as when requests are successful,
but it returns false for `#success?`. This is handy for debugging when you'll likely
want to inspect the raw response. We can still raise on the Maybe side if needed.
But that's a Maybe concern not a provider concern.

* `ExchangeRate#find_rate` doesn't raise anymore in preparation for opting into fetching —
which should only happen on writes.
```

Will update the PR description tomorrow to match.

Ran out of time today to cherry pick just the parts we need. I'll do that tomorrow. Of course, the Maybe team should feel free to do it themselves if they're in a hurry.

Gotta say I really like where we're at with this. And I like the approach we took: take the learnings from a previously live product, validate the approach by scaling it artificially, implement a pared down version of the validated approach. Let's hope it pans out in the long run as well!

josefarias:
Closing before cherry-picking the immediately relevant bits in a separate PR.

This can be used as reference when we introduce new concepts and providers. Judgement should be used considering this PR was based on assumptions about how non-exchange rate concepts would work.

zachgoll:
@josefarias awesome, I agree that this is at a good spot.  The mental model is pretty clear and intuitive to me so hopefully that holds true for future contributors to the project as well and gives us a good launch point.  Will go check out the new PR!

josefarias:
New PR is up https://github.com/maybe-finance/maybe/pull/574 @zachgoll

PR #567, opened 2024-03-25
add uniq index for users email to ensure uniqueness

adrienpoly:
I noticed that the User model has a uniqueness validation for the email. It is usually recommended to back this uniqueness at the database level too. This PR adds this uniq index.

The default Rails rubocop settings does not provide this out of the box but there is a cop for that. It is personally one of the few extra rules I add as it is quit easy to forget

If needed, the rule is 

```yml
Rails/UniqueValidationWithoutIndex:
  Enabled: true
```


zachgoll:
@adrienpoly good catch, definitely want this enforced at the DB level

PR #569, opened 2024-03-26
to_series doesn't expect argument anymore

ThibautGrx:
I'm getting an error when deleting a valuation 
image


zachgoll:
@ThibautGrx nice catch 👍

PR #574, opened 2024-03-27
Implement Synth as an exchange rate provider

josefarias:
This PR is a pared down version of https://github.com/maybe-finance/maybe/pull/561, where we sketched out a multi-provider, multi-concept implementation. Please see the other PR for a detailed explanation.

### DIFF
--- a/app/models/provider/synth.rb
+++ b/app/models/provider/synth.rb
@@ -35,7 +35,7 @@ class Provider::Synth
   private
     attr_reader :api_key
 
-    ExchangeRateResponse = Struct.new(:rate, :success?, :error, :raw_response, keyword_init: true)
+    ExchangeRateResponse = Struct.new :rate, :success?, :error, :raw_response, keyword_init: true
 
     def base_url
       "https://api.synthfinance.com"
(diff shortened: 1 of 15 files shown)


app/models/provider/synth.rb
```diff
@@ -0,0 +1,51 @@
+class Provider::Synth
+  include Retryable
+
+  def initialize(api_key = ENV["SYNTH_API_KEY"])
+    @api_key = api_key || ENV["SYNTH_API_KEY"]
+  end
```
josefarias, line 6: This looks like we're doubly guarding for the same thing. We're not. But it's an admittedly slightly premature guard.

When we ultimately pass in `ENV["EXCHANGE_RATE_PROVIDER_API_KEY"]` — that value might be nil.

If we want to both be able to instantiate this without any args (`Provider::Synth.new`) and also be able to pass in a possibly `nil` key, I think this would be the way to support that.

zachgoll (approved):
This looks great and I think it's at a perfect starting level of abstraction.  Thanks for the collab on this one!

PR #575, opened 2024-03-27
Allow event propagation to fix turbo frame update

ThibautGrx:


ThibautGrx:
I'm new to turbo, but event probably has to bubble to parents and reach turbo frame.

I may miss something but I couldn't find why it was necessary to stop event propagation.

zachgoll:
@ThibautGrx at first glance I was thinking the same, but I think stopping event propagation actually may be useful and necessary if we were using this controller as a _true_ "dropdown" (where elements can be selected from the list).  In that case, we don't want the dropdown to close since the user is still selecting things.

In this case, I think our issue is we're using the wrong type of controller for the use case.  We should probably distinguish between:

- "Menu" - a "dropdown" that contains links and buttons (i.e. "actions")
- "Dropdown" - a "dropdown" that contains selectable items and behaves like [select](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select)

A "Menu" should _allow_ event propagation.  A "Dropdown" should stop it as this controller does.

Thoughts on that approach?

ThibautGrx:
@zachgoll ok I see.
I've made some changes adding a closeOnClick setting. 
For now, the two kind of dropdown behave almost the same way, so it was quicker to add a setting .  What do you think about it ?

zachgoll (approved):
Agreed, they do behave very similar right now, so I'm cool with this solution.  Thanks for tackling, this was a tricky one.

Issue #573, opened 2024-03-26
Allow user to search transaction categories list


zachgoll:
## Overview

Currently, there is a disabled UI placeholder for transaction category searching.

This issue will allow the user to type to search for a category:

![CleanShot 2024-03-26 at 14 46 05](https://github.com/maybe-finance/maybe/assets/16676157/16f835e5-ceea-46b2-b7a4-fdfc83a23490)

## Requirements

- Search should be debounced (see [this file](https://github.com/maybe-finance/maybe/blob/main/app/javascript/controllers/transactions_search_controller.js) for a good reference)
- Search should perform a full text search on the category name (we have `ransack` Gem setup, but may not be needed for this due to simplicity)
- Search can be cleared with browser native search controls
- When no results found, add a "No categories found" message in the UI

### Out of Scope

This issue will NOT cover:

- Category create/edit/delete actions (this is covered in #572)

## Designs

The designs are already implemented for this.  You can reference [this Figma file](https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2410-1384&mode=design&t=U1FgrK5nHhTYyk1W-0) for additional context.

mattia-malnis:
I'm working on this!

PR #577, opened 2024-03-27
Allow user to search transaction categories

mattia-malnis:
This closes [#573](https://github.com/maybe-finance/maybe/issues/573)

Here the result: 

https://github.com/maybe-finance/maybe/assets/15941778/d8351759-214e-4806-aca8-60942f52321a

I'm not sure about the Stimulus controller that auto submits the search input, I've created a new one that is similar to [transactions_search_controller.js](https://github.com/maybe-finance/maybe/blob/main/app/javascript/controllers/transactions_search_controller.js). Probably the best thing is to merge `auto_submit_form_controller.js`, `transactions_search_controller.js` and `transactions_category_search_controller.js` in a single controller.

### DIFF
--- a/app/javascript/controllers/list_filter_controller.js
+++ b/app/javascript/controllers/list_filter_controller.js
@@ -2,15 +2,29 @@ import { Controller } from "@hotwired/stimulus";
 
 // Basic functionality to filter a list based on a provided text attribute.
 export default class extends Controller {
-  static targets = ["input", "list"];
+  static targets = ["input", "list", "emptyMessage"];
 
   filter() {
     const filterValue = this.inputTarget.value.toLowerCase();
     const items = this.listTarget.querySelectorAll(".filterable-item");
+    let noMatchFound = true;
+
+    if (this.hasEmptyMessageTarget) {
+      this.emptyMessageTarget.classList.add("hidden");
+    }
 
     items.forEach((item) => {
       const text = item.getAttribute("data-filter-name").toLowerCase();
-      item.style.display = text.includes(filterValue) ? "" : "none";
+      const shouldDisplay = text.includes(filterValue);
+      item.style.display = shouldDisplay ? "" : "none";
+
+      if (shouldDisplay) {
+        noMatchFound = false;
+      }
     });
+
+    if (noMatchFound && this.hasEmptyMessageTarget) {
+      this.emptyMessageTarget.classList.remove("hidden");
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


zachgoll:
@mattia-malnis thanks for tackling this!  I do agree that there is some consolidation that could be done with the controllers, but I think we're okay to wait a little bit longer before trying to abstract things.

In regards to this solution, I'm wondering if we could avoid hitting the backend entirely since it's a simple filter operation.  By default, we show all possible categories in a scrollable container, and then attach a simple filter controller like this?

```js
import { Controller } from "@hotwired/stimulus";

// Basic functionality to filter a list based on a provided text attribute.
  static targets = ["input", "list"];

  filter() {
    const filterValue = this.inputTarget.value.toLowerCase();
    const items = this.listTarget.querySelectorAll(".filterable-item");

    items.forEach((item) => {
      const text = item.getAttribute("data-filter-name").toLowerCase();
      item.style.display = text.includes(filterValue) ? "" : "none";
    });
  }
}
```

Is there a specific reason we'd need to hit the backend here that I'm overlooking?  I think we'll need this in other areas of the app and it could be nice to have a generic controller to use.  For example, I've got this controller wired up locally for account filtering:

https://github.com/maybe-finance/maybe/assets/16676157/3683de2d-3c2c-4d8e-8759-7f167eea321d

mattia-malnis:
@zachgoll I thought that a backend search was required based on the issue description, but a client-side filter could work well in this case.

I see only a drawback using a client-side filtering: users may not get results for misspelled or slightly different terms, as the filter would be based on a string match. Using PG trigram we can match similar-sounding or slightly misspelled terms.

If this isn't a problem I can implement a JS filter!

zachgoll:
@mattia-malnis gotcha, yeah that's on me, will specify that kind of stuff better in future issues.

I think for simplicity, client-side would be a good starting point.  If you want to leave a comment with a link to this PR that may be a good way to reference the server-side implementation if requirements ever force us down that path in the future.

zachgoll:
@mattia-malnis if you're still working on this, here is a controller that will probably work:

https://github.com/maybe-finance/maybe/blob/main/app/javascript/controllers/list_filter_controller.js

In the [transaction filters PR](https://github.com/maybe-finance/maybe/pull/578), I consolidated some of the controllers as you were mentioning above (transaction search, auto submit, etc.)

mattia-malnis:
@zachgoll sorry I was a bit busy today. If it's not a problem I'd like to finish this between tonight and tomorrow morning!

zachgoll:
@mattia-malnis no worries, that's totally fine!  Just wanted to let you know of that change as it can probably be reused here in this PR.

mattia-malnis:
@zachgoll I've finished this PR, categories are now filtered client-side. 
I made a small change to the `list_filter_controller.js` controller. I've added an optional `emptyMessage` target that helps display the "No categories found" message when no categories match the filter criteria. This target is optional, so it won't break the existing transactions filtering implementation.

https://github.com/maybe-finance/maybe/assets/15941778/b280eeed-5f0e-4a6a-9f66-1616daef529d

zachgoll (approved):
Looks good, nice idea adding the empty message target!

PR #580, opened 2024-03-28
Add Custom Turbo Confirm Modal

### LINKED ISSUE #570: Build custom `data-turbo-confirm` modal
This issue is for building a custom modal that replaces the [browser's default confirm behavior](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm).

We can do this globally via the `Turbo.setConfirmMethod`:

```js
Turbo.setConfirmMethod(() => {
  return new Promise((resolve, reject) => {
    // show modal
    // on confirm, resolve(true)
    // on cancel, resolve(false)
  })
})
```

## Requirements

CleanShot 2024-03-26 at 11 57 59

- Each usage of `data-turbo-confirm` should be customizable (with sensible default values)
  - Title can be customized
  - Description can be customized with any valid HTML
  - Confirm button text can be customized
- Has cancel button (X icon in top right corner)
- Has confirm button (main button at bottom)

## Design

Please visit the following link for the design:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2404-108&mode=design&t=U1FgrK5nHhTYyk1W-0

## Misc

- [Here's a good video](https://www.youtube.com/watch?v=jn22KyMEdwg) to get started with.
- The codebase already has one usage instance of this here:

https://github.com/maybe-finance/maybe/blob/d274d1a39447d552ecbb7e951d0df15a3a508d3f/app/views/accounts/_account_valuation_list.html.erb#L36

crolla97:
Closes #570 

## Solution
Implemented turbo dialog with customisable messages.
Screenshot 2024-03-28 at 22 36 20

To customise the dialog message you pass can pass an object with 3 options, title, body and accept (button)

Example
```
<%= link_to valuation_path(valuation.original),  data: { turbo_method: :delete, turbo_confirm: { title: t('turbo_confirm.history.title'), body: t('turbo_confirm.history.body'), accept: t('turbo_confirm.history.accept') } } do %>
```

## Notes

I've added fallbacks to handle scenarios where just a string or empty value is passed through the turbo confirm resulting in the default values being shown. Also, if you only wanted to change a single text value, for instance the button, you can do that by passing  `turbo_confirm: {accept: "button"}`.

Unsure if this is the best way of handling fallbacks so open to feedback.


app/javascript/controllers/application.js
```diff
@@ -6,4 +6,31 @@ const application = Application.start()
...
+    msg = {
+      title: "Are you sure?",
+      body: "You will not be able to undo this decision",
+      accept: "Confirm"
+    }
+  }
+
+  dialog.querySelector("[id='turbo-confirm-title']").innerHTML = msg.title ? msg.title : "Are you sure?";
+  dialog.querySelector("[id='turbo-confirm-body']").innerHTML = msg.body ? msg.body : "You will not be able to undo this decision";
+  dialog.querySelector("[id='turbo-confirm-accept']").innerHTML = msg.accept ? msg.accept : "Confirm";
```
adrienpoly, line 25: The default messages are defined in the modal template and they are translated (coming from I18n keys). With this logic we are losing the ability to get a default message translated.

I would suggest to conditionally  replace the text if they are provided otherwise keep the default text from the template

zachgoll: What about a simple check for the value?  I think the benefit of something like this would be:

- We keep i18n intact
- Developer can "opt-in" to how much customization they need
  - If only title needed, just pass a string: `turbo_confirm: "Are you sure?"`
  - Or... `turbo_confirm: { title: "Are you sure?" }`
  - Or... `turbo_confirm: { title: "...", body: "...", accept: "..." }`

```js
Turbo.setConfirmMethod((message, element) => {
  const dialog = document.getElementById("turbo-confirm");

  try {
    const { title, body, accept } = JSON.parse(message);

    if (title) {
      document.getElementById("turbo-confirm-title").innerHTML = title;
    }

    if (body) {
      document.getElementById("turbo-confirm-body").innerHTML = body;
    }

    if (accept) {
      document.getElementById("turbo-confirm-accept").innerHTML = accept;
    }
  } catch (e) {
    document.getElementById("turbo-confirm-title").innerText = message;
  }

  dialog.showModal();

  return new Promise((resolve, reject) => {
    dialog.addEventListener(
      "close",
      () => {
        resolve(dialog.returnValue == "confirm");
      },
      { once: true }
    );
  });
});
```

app/views/shared/_custom_confirm_modal.html.erb
```diff
@@ -0,0 +1,16 @@
...
+        <h3 id="turbo-confirm-title" class="font-medium text-md"><%= t('turbo_confirm.default.title') %></h3>
+        <button value="cancel">
+          <%= lucide_icon("x", class: "w-5 h-5 shrink-0 text-gray-500") %>
+        </button>
+      </div>
+      <p id="turbo-confirm-body" class="text-gray-500 text-sm">
+        <%= t('turbo_confirm.default.body') %>
+      </p>
+    </div>
+    <button id="turbo-confirm-accept" class="w-full text-red-600 rounded-xl text-center p-[10px] border" value="confirm"><%= t('turbo_confirm.default.title') %></button>
```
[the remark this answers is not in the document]
crolla97: yes good catch

config/locales/views/turbo_confirm/en.yml
```diff
@@ -0,0 +1,10 @@
+en:
+  turbo_confirm:
+    default:
+      title: Are you sure?
+      body: "<p>You will not be able to undo this decision</p>"
```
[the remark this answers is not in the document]
crolla97: Didn't know about safe html translations that's good to know.

Yes I like this change, I wasn't thinking about the other shared partials that will need translations eventually

app/javascript/controllers/application.js
crolla97: @adrienpoly @zachgoll Thanks that's pretty much what I was looking to achieve, don't know why I thought overriding the default locales to static strings was a good idea.

zachgoll (approved):
Awesome, looks good now!  Thanks for tackling this.

Issue #579, opened 2024-03-28
Add transaction category filtering


zachgoll:
## Overview

Allow the user to select 1 or more categories to filter transactions by:

![CleanShot 2024-03-28 at 14 03 32](https://github.com/maybe-finance/maybe/assets/16676157/75d73839-21e1-4724-993f-bb4c40bedd40)

## Requirements

- Use the existing family categories (no new categories needed here)
- User can select 1 or more categories
- Selected categories will be rendered in the filters list above the results
  - Filter will show category name and color
- Form auto submits when category is selected (see "accounts" filter in #578 for example)

## Designs

You can find the design spec for this at this link:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2443-295&mode=design&t=U1FgrK5nHhTYyk1W-0

PR #581, opened 2024-03-29
Filter transactions by category

ThibautGrx:
closes https://github.com/maybe-finance/maybe/issues/579

### DIFF
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -34,6 +34,10 @@ class Transaction < ApplicationRecord
           value.each do |account_id|
             filters << { type: "account", value: family.accounts.find(account_id), original: { key: key, value: account_id } }
           end
+        when "category_id_in"
+          value.each do |category_id|
+            filters << { type: "category", value: family.transaction_categories.find(category_id), original: { key: key, value: category_id } }
+          end
         when "category_name_or_account_name_or_name_cont"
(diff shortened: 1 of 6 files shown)


ThibautGrx:
I've tried to fix some small UI dysfunctions at the same time.
- Tick/untick the checkbox when clicking on the label
- Keep the dropdown opened when clicking on tabs

However, the dropdown closes when selecting a category (or account) because turbo stream replace the whole turbo frame "transactions_search_form". Thoughts on how we could fix it ?

zachgoll (commented):
Looks good!

In regards to the dropdown closing when a category is selected, we've got updated designs coming soon that will introduce an "apply filters" button on each of these tabs so the user can select multiple options and then explicitly reload the results.

Until those are ready, I think this behavior is okay for the moment.

zachgoll:
@ThibautGrx was doing one last check on this prior to merging and noticed that when the category filter is applied, the "badge" does not show up.

Can you update this so those are added?  Here's where the core of that logic is happening, and you can use #578 as a reference as well.

https://github.com/maybe-finance/maybe/blob/b1bfdef8ffa6ef3b43a7b7cc12f898d2f4771365/app/models/transaction.rb#L23-L42

ThibautGrx:
You're right , I went a little fast... here's the update

zachgoll (approved):
Awesome, thanks for the update, looks good!

Issue #571, opened 2024-03-26
Add ability to delete `Account`


zachgoll:
## Overview

Currently, a user has the ability to toggle the `is_active` status of an `Account`, but cannot fully delete an account.

This issue will implement full deletes with correct cascade behavior for associated models like `Transaction` and `Valuation`.

## Requirements

- User can delete account from account page
- When `Account` deleted, all of the following associated records should cascade delete:
  - `Transaction`
  - `Valuation`
  - `Account::Balance`
- When `Account` delete is triggered, a "confirm" modal should be triggered.
  - **NOTE:** for this PR, please check the status of #570 .  If it is not completed, this PR will use the browser's default confirm modal.

## Design

See the following design file:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2406-463&mode=design&t=U1FgrK5nHhTYyk1W-0

## Future Improvements

This issue is _likely_ and intermediate step.  In the future, we _may_ need to implement "soft deletes" on the `Account` model for either compliance reasons, to prevent accidental deletions, or both.

mattia-malnis:
Working on this!

zachgoll:
Completed by #582

PR #582, opened 2024-03-29
Add ability to delete Account

mattia-malnis:
Issue [#571](https://github.com/maybe-finance/maybe/issues/571)

https://github.com/maybe-finance/maybe/assets/15941778/46ae4228-e795-4fb4-b727-4d38f47b38b5

### DIFF
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -2,7 +2,7 @@ module Account::Syncable
     extend ActiveSupport::Concern
 
     def sync_later
-        AccountSyncJob.perform_later self
+        AccountSyncJob.perform_later(id)
     end
 
     def sync
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -2,7 +2,7 @@ module Account::Syncable
     extend ActiveSupport::Concern
 
     def sync_later
-        AccountSyncJob.perform_later(id)
+        AccountSyncJob.perform_later self
     end
 
     def sync
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   scan_ruby:
     runs-on: ubuntu-latest
-
+    timeout-minutes: 10
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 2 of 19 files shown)


mattia-malnis:
I needed to change `AccountSyncJob`. When an account is deleted, we get the `ActiveJob::DeserializationError` because the job was enqueued using an account instance that could be deleted. 
By enqueuing the job with only the Account ID, the job can retrieve the Account record from the database when it's executed, avoiding deserialization errors caused by deleted records.

mattia-malnis:
@zachgoll I'm trying to figure out how to fix the error "No live threads left. Deadlock?" and I noticed [your PR](https://github.com/maybe-finance/maybe/pull/583). 
When you'll close the PR 583 I'll remove my commit `c0e4ae8` as I didn't know how to stop this loop.

zachgoll:
@mattia-malnis yep, I saw your PR was getting stuck on the CI tests, so opened up #583 to unblock this one. 

I _think_ you should be all set on this PR now.  Let me know if you run into issues.

mattia-malnis:
@zachgoll thanks a million!

I should've finished with this, the only missing part is to add the confirm modal of the [PR 580](https://github.com/maybe-finance/maybe/pull/580).

Furthermore, I was thinking about on how many balances/valuations/transactions will be destroyed whan an account is deleted. Do you think it would be better to handle the destruction of these dependencies in the background?

mattia-malnis:
@zachgoll now you can find the custom confirm modal!


https://github.com/maybe-finance/maybe/assets/15941778/83bac6df-209b-449c-b9f7-02ebd06c99ba

app/models/account/syncable.rb
```diff
@@ -2,7 +2,7 @@ module Account::Syncable
     extend ActiveSupport::Concern
 
     def sync_later
-        AccountSyncJob.perform_later self
+        AccountSyncJob.perform_later(id)
```
zachgoll, line 5: Was this a leftover commit?  Or is this still required?

mattia-malnis: Now we have removed `after_commit :sync_account` from Transaction and Valuation so it's not necessary, I reverted the commit!

zachgoll: Awesome, thanks!

PR #583, opened 2024-03-29
Isolate infinite loop bug, add timeout to actions

zachgoll:
The following causes an infinite loop on cascade deletes:

1. Account gets deleted
2. Dependent valuation/transaction gets deleted, triggering `sync_account`
3. `sync_account` interacts with `Account`, causes deadlock

```
# account.rb

has_many :valuations, dependent: :destroy
has_many :transactions, dependent: :destroy
```

```
# valuation.rb

after_commit :sync_account
```

```
# transaction.rb

after_commit :sync_account
```

## Solution 

- Add timeout to Github Actions (to protect against this on future PRs)
- Trigger sync process explicitly at the controller (rather than model) layer.  The sync process will continue to increase in complexity and this gives more flexibility and explicitness for triggering syncs.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -6,9 +6,10 @@ class Account < ApplicationRecord
 
   broadcasts_refreshes
   belongs_to :family
-  has_many :balances
-  has_many :valuations
-  has_many :transactions
+
+  has_many :balances, dependent: :destroy
+  has_many :valuations, dependent: :destroy
+  has_many :transactions, dependent: :destroy
 
   monetize :balance
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   scan_ruby:
     runs-on: ubuntu-latest
-
+    timeout-minutes: 10
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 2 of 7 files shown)


app/models/account.rb
```diff
@@ -6,9 +6,10 @@ class Account < ApplicationRecord
 
   broadcasts_refreshes
   belongs_to :family
-  has_many :balances
-  has_many :valuations
-  has_many :transactions
+
+  has_many :balances, dependent: :destroy
+  has_many :valuations, dependent: :destroy
+  has_many :transactions, dependent: :destroy
```
zachgoll, line 13: This is correct behavior and matches the behavior at the database level.  The previous version was an oversight.

zachgoll:
Here is the intentionally forced loop error, cancelled after 5 minutes due to GH Action timeout:

CleanShot 2024-03-29 at 12 05 06

See: 

https://github.com/maybe-finance/maybe/actions/runs/8482867745/job/23242934822?pr=583

zachgoll:
This PR is blocking several others, so merging it now.

That said, the strategy for triggering syncs will be an important concept for this app, so feel free to comment here on this PR after it's closed, open a discussion, or separate issues if there's a better way to do this.

PR #598, opened 2024-04-03
Centralize authentication

zachgoll:
By default, every page in the app should require authentication except registration/login/password resets.

Centralizing auth allows us to [implement actions that run around the authentication](https://github.com/maybe-finance/maybe/pull/594#discussion_r1549657005) lifecycle easier.

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -2,7 +2,13 @@ module Authentication
   extend ActiveSupport::Concern
 
   included do
-    helper_method :user_signed_in?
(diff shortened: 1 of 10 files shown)


app/controllers/concerns/authentication.rb
```diff
@@ -15,10 +21,6 @@ def authenticate_user!
     end
   end
 
-  def user_signed_in?
-    Current.user.present?
-  end
```
zachgoll, line 20: Not currently in use.  Removing until there is a need.

PR #599, opened 2024-04-03
Simplify setup using Docker Compose

Quintasan:
This is a barebones configuration for Docker Compose.

1. `docker compose build`
2. `docker compose up -d`

Application will be available on `http://localhost:3000`.

Actual production deployments will require a reverse proxy in front of the `app` container.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,7 @@
+# You can safely generate one with
+# cat /dev/urandom | head -n 10 | sha256sum
+SECRET_KEY_BASE=
+
 # Exchange Rate API
 # This is used to convert between different currencies in the app. We use Synth, which is a Maybe product. You can sign up for a free account at synthfinance.com.
 SYNTH_API_KEY=
@@ -12,10 +16,10 @@ SMTP_PASSWORD=
 TLS=true
 
 # Database Configuration
-DB_HOST=localhost # May need to be changed to `DB_HOST=db` if using devcontainer
+DB_HOST=db
 DB_PORT=5432
-POSTGRES_PASSWORD=postgres
-POSTGRES_USER=postgres
+POSTGRES_USER=maybe
+POSTGRES_PASSWORD=changeme
 
 # App Domain
 # This is the domain that your Maybe instance will be hosted at. It is used to generate links in emails and other places.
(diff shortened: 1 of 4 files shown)


zachgoll:
@Quintasan the ENV changes and sample file make sense here, but could you explain the overall usage of this in a bit more detail?  I'll be doing a bunch of work this week around Docker and self-hosting, so just want to make sure I understand the purpose here.

I'm guessing this is mainly a setup that allows self-hosters to specify their own custom Docker configs?

Quintasan:
> @Quintasan the ENV changes and sample file make sense here, but could you explain the overall usage of this in a bit more detail? I'll be doing a bunch of work this week around Docker and self-hosting, so just want to make sure I understand the purpose here.
> 
> I'm guessing this is mainly a setup that allows self-hosters to specify their own custom Docker compose configs?

@zachgoll The intent here is to have a minimal hassle, non-opinionated way to deploy Maybe using Docker Compose. We could make it the **default** (because right now you have to `cp compose.example.yml compose.yml`) and instruct users that they can use `compose.override.yml` to change any setting they don't like, since it's sourced by Docker Compose if present.

zachgoll:
@Quintasan gotcha, so you're looking to self-host on a VPS and use a `docker-compose pull && docker-compose up` workflow for keeping it on the latest release, correct?

Quintasan:
> @Quintasan gotcha, so you're looking to self-host on a VPS and use a `docker-compose pull && docker-compose up` workflow for keeping it on the latest release, correct?

Ideally yes, but since you don't actually publish an image anywhere you need to `docker compose build` before.

Quintasan:
> @Quintasan got it, I'm thinking we'll want to provide an _opinionated_ compose file in source control along with publishing an official image on each release per #295. I've got a bit of local work around this right now, so I think at this moment if we could pare this PR down to just the ENV cleanup that would be great. I'll have a PR opened this week for the deployment scenario you're describing, so if you have specific requirements that you'd be looking for, let me know.

@zachgoll  Makes sense. If you're willing to push up your changes up to a visible branch then I'd be happy to collaborate on the compose file

zachgoll:
@Quintasan sounds great, hoping to get something opened in the next day or two.

zachgoll:
@Quintasan okay so we now have a much clearer path for Docker and self-hosting.  I've opened up #627 with detailed requirements and to act as a centralized collaboration spot.

PR #603, opened 2024-04-08
Fix error when users try to logout

portovep:
**Steps to reproduce**
On the dashboard page, after logging it, when the user tries to logouts an internal error is shown.

image

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -12,6 +12,9 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
       fill_in "Password", with: "password"
       click_button "Log in"
     end
-    assert_text "Dashboard"
+    assert_text "Dashboard", wait: 5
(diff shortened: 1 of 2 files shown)


test/application_system_test_case.rb
```diff
@@ -12,6 +12,9 @@ def sign_in(user)
       fill_in "Password", with: "password"
       click_button "Log in"
     end
-    assert_text "Dashboard"
+    assert_text "Dashboard", wait: 5
```
portovep, line 15: While adding a test to cover the broken functionality, I ran into some flakiness running the test. The dashboard page was taking a bit longer to load and the default 2 seconds waiting time was not enough.

I increased the default Capybara waiting time from up to 2 seconds to up to 5 seconds to make the test more stable.

zachgoll (approved):
Thanks for the fix, looks good 👍

PR #609, opened 2024-04-08
Added lints for ERB templates

igor-alexandrov:
This PR adds lints to the ERB templates. Yes, `erblint` is not perfect, however currently it is the only solution that exists and works.

I fixed existing errors and added `erblint` step to the CI.


Gemfile
```diff
@@ -43,6 +43,7 @@ group :development, :test do
   gem "dotenv-rails"
   gem "letter_opener"
   gem "i18n-tasks"
+  gem "erb_lint"
```
zachgoll, line 46: What do you think of adding a `.erb-lint.yml` file for a bit more explicitness and integrating with our Rubocop setup?

```yml
EnableDefaultLinters: true
linters:
  Rubocop:
    enabled: true
    rubocop_config:
      inherit_from:
        - .rubocop.yml
```

zachgoll (commented):
Thanks for the cleanup!  Just left a few potential suggestions.

.github/workflows/ci.yml
```diff
@@ -54,6 +54,22 @@ jobs:
...
+        uses: actions/checkout@v4
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: .ruby-version
+          bundler-cache: true
+
+      - name: Lint templates for consistent style
+        run: ./bin/erblint ./app/**/*.erb
```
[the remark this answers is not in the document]
igor-alexandrov: @zachgoll I merged two tasks into one. I prefer calling `./bin/erblint ./app/**/*.erb` because of explicitness, but `--lint-all` also makes sense, updated.

Gemfile
igor-alexandrov: @zachgoll Yes, I forget about the config for the `erblint`. At the same time, I am completely against enabling all Rubocop rules for `erblint`. I tried doing this locally, and I got about 3500 errors with only 3000 auto-correctable. Besides this, some HTML tags started to look really strange after the automatic correction.

igor-alexandrov: I added `.erb-lint.yml`, please check.

zachgoll: Sounds good, and very understandable on the rubocop integration.  I'm all for auto checks, but not if it's going to slow us down too much.

zachgoll (approved):
Looks good! Thanks again for the cleanup 👍

igor-alexandrov:
`erblint` can be tricky. I prefer to have a very explicit call that will only lint files inside of an `./app` folder.

adrienpoly:
Very nice addition 

with the linters growing in the CI (in the future we could have i18n-tasks, eslint etc) I usually add to my project a small utility to run them all in a single command

something like that in `bin/lint`

```bash
#!/usr/bin/env bash

# Exit when any command fails
set -e

# Check Ruby code formatting
echo "Checking Ruby code formatting..."
bundle exec rubocop -A

# Check erb file formatting
echo "Checking erb file formatting..."
bundle exec erblint --lint-all --autocorrect
```

igor-alexandrov:
@adrienpoly I will open a new PR for this. Thank you for this suggestion.

PR #610, opened 2024-04-08
Enhance currency selector to list 'All Others' after 'Popular'

mikekosulin:
This PR enhances the currency selector by ensuring that not only popular currencies are displayed, but it also categorizes other available currencies within optgroups.

money currency selector w/ optgroups


zachgoll (approved):
Nice improvements here!

Issue #597, opened 2024-04-03
Extend `shared/notification` partial to allow persistent notifications with action buttons


zachgoll:
As part of #596, we will need to create notifications that are _persistent_ until the account sync completes:
![CleanShot 2024-04-02 at 21 15 52](https://github.com/maybe-finance/maybe/assets/16676157/eacee004-7249-4280-985c-92b5a2d33444)

## Requirements

- `shared/notification` should accept `options[:auto_dismiss]`, which should default to true.  When false, notification is persistent until explicitly removed or a page refresh.
- `shared/notification` should be able to accept arbitrary content that provides a "dismiss" or custom action button:

![CleanShot 2024-04-03 at 10 10 27](https://github.com/maybe-finance/maybe/assets/16676157/6a4019b2-5dc3-4f50-a17c-67c40f67174b)

## Designs

See the link below:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2568-2224&mode=design&t=XdQIQAEd4nZtUHfU-0

ThibautGrx:
I'm working on this!

ThibautGrx:
We keep the spinner ? When we click on it, it closes the notification. Do we keep this behavior ?
![image](https://github.com/maybe-finance/maybe/assets/35230028/5f090e1a-dbed-41a3-af9a-015eeef8cf29)

zachgoll:
@ThibautGrx I think we can remove that behavior and replace with explicit "dismiss" buttons.

PR #611, opened 2024-04-10
Add persistent notification with dismiss and action button 

ThibautGrx:
closes https://github.com/maybe-finance/maybe/issues/597


ThibautGrx:
Maybe notification partial is becoming a little complex ... Do you want me to rework code ? create partials  helpers ?

zachgoll (commented):
Looking good, thanks for tackling this one! 

Just a few things that I think we need to address before merging: 

## Styling

There are a few inconsistencies between this and the design file (also check spelling):

CleanShot 2024-04-15 at 11 55 48

vs.

CleanShot 2024-04-15 at 11 57 20

Here's the official spec for this one: https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?node-id=2568%3A2765&mode=dev

## I18n

For default messages like "Dismiss" in the `_notification` partial, I think we should add to a localization file.

ThibautGrx:
Thanks for the review. I've tried to reproduce the spec, I hope it will be better.

zachgoll (approved):
Nice, looks good now!

PR #614, opened 2024-04-12
Allow 2 decimals when adding a balance

### LINKED ISSUE #613: New account is unable to have cents
I've been playing with the self-hosting branch because I'm excited about maybe and this was the first time I got to play with it. 
When I went to add a new bank account (manual entry) I noticed that I wasn't allowed to have any cents as the starting value. I know its early days but thought I'd report the bug anyway! I also tried to enter it register-style as `100011` but that turned into `$100,011`

image

pieterbeulque:
Fixes #613 

This should fix money fields throughout the app.

One thing I'm not sure of is that not all currencies allow decimals, so this should maybe be dynamic based on the currency this input is in? Maybe that's a property on that `Money.new(0, money&.currency || Money.default_currency)` line too?

### DIFF
--- a/app/helpers/application_form_builder.rb
+++ b/app/helpers/application_form_builder.rb
@@ -36,7 +36,8 @@ class ApplicationFormBuilder < ActionView::Helpers::FormBuilder
     default_options = {
       class: "form-field__input",
       value: money&.amount,
-      placeholder: Money.new(0, money&.currency || Money.default_currency).format
+      placeholder: Money.new(0, money&.currency || Money.default_currency).format,
+      step: "0.01" # Not all currencies have 2 decimal places
     }
 
     merged_options = default_options.merge(options)


app/helpers/application_form_builder.rb
```diff
@@ -36,7 +36,8 @@ def money_field(method, options = {})
     default_options = {
       class: "form-field__input",
       value: money&.amount,
-      placeholder: Money.new(0, money&.currency || Money.default_currency).format
+      placeholder: Money.new(0, money&.currency || Money.default_currency).format,
+      step: "0.01" # Not all currencies have 2 decimal places
```
zachgoll, line 40: RE dynamically setting based on currency:

That's a good point and I'm wondering if we add an instance method in `Money::Currency`:

```
def step
    return 1 if @minor_unit_conversion.nil? || @minor_unit_conversion == 1
    1.0 / @minor_unit_conversion.to_f
end
```

And then use it here:

```
step: (money&.currency || Money.default_currency).step
```

pieterbeulque: I'm just curious how that will behave when you change the dropdown value? If the default currency is EUR or USD it will be set to `0.01`, but if you change that dropdown value to JPY, it won't auto-update the money field, right? Do we need to add in some JS to handle that on change?

zachgoll: Good point, wasn't thinking about that.  I think your solution works for the time being, so I'll merge this and I just opened up #625 to capture this additional requirement.

pieterbeulque: Sweet, cheers. Curious to learn what'd be the modern Rails way to do this.

zachgoll: Same, adding a JS controller just for this seems overkill, but I can't think of any other solution myself

PR #616, opened 2024-04-13
Add Catalan localization

aryiu:


zachgoll:
@aryiu really appreciate the contribution but I think at the moment we should hold off on adding new translations.

We are pushing new features on a daily basis that require additional translations, so these will quickly fall out of sync.

Long term we will be prioritizing I18n but due to the global nature of this repo we will likely need to build an automated, AI-assisted process to keep up with the tens of languages we need to translate for.

PR #622, opened 2024-04-15
Add german translations

alex289:
Hey, I would like to add german translations to this project.

In german there are two ways to say "you". One is more formal and one is more personal. For financing most platforms choose the formal one so I did too. I can change it to the personal one if you want

### DIFF
--- a/config/locales/views/passwords/de.yml
+++ b/config/locales/views/passwords/de.yml
@@ -1,0 +1,10 @@
+---
+de:
+  passwords:
+    edit:
+      password: Neues Passwort
+      password_challenge: Derzeitiges Passwort
+      submit: Password zurücksetzen
+      title: Passwort aktualisieren
+    update:
+      success: Ihr Passwort wurde zurückgesetzt.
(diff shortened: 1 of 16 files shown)


config/locales/views/passwords/de.yml
```diff
@@ -0,0 +1,10 @@
+---
+de:
+  passwords:
+    edit:
+      password: Neues Passwort
+      password_challenge: Derzeitiges Passwort
+      submit: Password zurücksetzen
```
lewis-wf, line 7: Password or passwor**t** here?

alex289: Good eye, thank u

zachgoll:
@alex289 really appreciate the contribution but I think at the moment we should hold off on adding new translations.

We are pushing new features on a daily basis that require additional translations, so these will quickly fall out of sync.

Long term we will be prioritizing I18n but due to the global nature of this repo we will likely need to build an automated, AI-assisted process to keep up with the tens of languages we need to translate for.

alex289:
> @alex289 really appreciate the contribution but I think at the moment we should hold off on adding new translations.
> 
> 
> 
> We are pushing new features on a daily basis that require additional translations, so these will quickly fall out of sync.
> 
> 
> 
> Long term we will be prioritizing I18n but due to the global nature of this repo we will likely need to build an automated, AI-assisted process to keep up with the tens of languages we need to translate for.

No problem 😄. If you still need some help with the german translations I would love t help.

Issue #629, opened 2024-04-15
Feature: Add transaction modal flow


zachgoll:
## Overview

This feature enables a user to add new transactions to an account, individually:

![CleanShot 2024-04-15 at 13 41 51](https://github.com/maybe-finance/maybe/assets/16676157/8fe25663-7f5d-4998-8e07-8bb1522096cd)

## Requirements

- [ ] User can click button on transaction page to open a modal
- [ ] User can click on button on account page
  - [ ] Account dropdown is pre-filled in this scenario
- [ ] Modal has a radio group of `Expense`, `Income`, `Transfer`
  - [ ] Transfers are NOT part of this issue.  A `cursor-not-allowed` class should be applied to the "Transfer" button and implementation should be _skipped_
- [ ] If "Expense", amount value is saved as a _positive_ value in the DB.  If "Income", amount is saved as _negative_ value in DB.
- [ ] Form uses native HTML input elements (no custom select or datepickers needed here)
- [ ] Form submit should be handled correctly (all fields should be required):
  - [ ] If missing fields or validation errors, modal stays open, no alerts
  - [ ] If successful submit, app shows success notification
  - [ ] If error, app shows error notification
- [ ] I18n translations are provided

## Design

You can find all the design specs below:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2664-225&mode=design&t=p9lOop61MfLabHUV-0

## Bounty
There is a $250 bounty on this. To participate in the bounty:

1. Comment `/attempt #629` to indicate you plan on working on this.
2. Submit a draft PR once you start and include `/claim #629` along with your general game plan for how you intent to implement this.
3. Change out of draft mode on the PR once you're ready to submit your attempt.
4. Include a short video showing functionality in your PR (drastically speeds up us approving).

In the event of multiple attempts, we generally pick the one that most closes matches the scope with the cleanest code and test coverage (when appropriate).

mehulmathur16:
Hi @zachgoll, can I be assigned ?

Shpigford:
> Hi @zachgoll, can I be assigned ?

Hi @mehulmathur16! Given the speed at which we're moving on the codebase, we don't assign issues or "give" issues to anyone at the moment.

Ekep-Obasi:
I'd like to attempt this
 /attempt #629

JoshAntBrown:
[...]

PR #633, opened 2024-04-16
Add transaction modal flow

josefarias:
This PR adds a manual transaction form inside a modal as described in https://github.com/maybe-finance/maybe/issues/629

## Preview
A preview of what this feature looks like in this PR

### Toggling through the tab radio buttons (10s):
GIF showing what it looks like to toggle through the tabbed design

### Full demo (20s):
https://github.com/maybe-finance/maybe/assets/31393016/5e529deb-62e3-47be-91e7-ba4c1d5b9173

## Requirements

* [x]  User can click button on transaction page to open a modal
* [x]  User can click on button on account page
  * [x]  Account dropdown is pre-filled in this scenario
* [x]  Modal has a radio group of `Expense`, `Income`, `Transfer`
  * [x]  Transfers are NOT part of this issue.  A `cursor-not-allowed` class should be applied to the "Transfer" button and implementation should be _skipped_
* [x]  If "Expense", amount value is saved as a _positive_ value in the DB.  If "Income", amount is saved as _negative_ value in DB.
* [x]  Form uses native HTML input elements (no custom select or datepickers needed here)
* [x]  Form submit should be handled correctly (all fields should be required):
  * [x]  If missing fields or validation errors, modal stays open, no alerts
  * [x]  If successful submit, app shows success notification
  * [ ]  ~~If error, app shows error notification~~ _(see * below)_
* [x]  I18n translations are provided

_* I think we'll need further considerations for this. All form fields are required, which means errors indicate foul play. We should raise in those cases, not return a friendly error._

/claim #629

### DIFF
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -49,16 +49,20 @@ class TransactionsController < ApplicationController
   end
 
   def new
-    @transaction = Transaction.new
+    @transaction = Transaction.new.tap do |txn|
+      if params[:account_id]
+        txn.account = Current.family.accounts.find(params[:account_id])
+      end
+    end
   end
 
   def edit
   end
 
   def create
-    account = Current.family.accounts.find(params[:transaction][:account_id])
-
-    @transaction = account.transactions.build(transaction_params)
+    @transaction = Current.family.accounts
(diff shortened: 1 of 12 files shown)


zachgoll (approved):
Nice work on this!

Issue #623, opened 2024-04-15
Multiple accounts share the same data


trfhgx:
I've noticed that when I have created a new account all the assests I've created in the original account are also displayed in the new account but not accessible they should be unique to each account
<img width="1336" alt="Screenshot 2024-04-15 at 11 08 22" src="https://github.com/maybe-finance/maybe/assets/42844286/e4630adf-f474-4228-a206-6a8c04e000e0">

PR #636, opened 2024-04-16
Fixed Data Access Restriction

Harry-kp:
Fixes https://github.com/maybe-finance/maybe/issues/623

This should prevent the data from being shared between two accounts.


zachgoll (approved):
Awesome, appreciate the fix!

Issue #627, opened 2024-04-15
Implement Self-Hosted Docker Compose Setup


zachgoll:
## Overview

Self hosting with Docker compose has been a highly requested feature and we plan to support it.  But due to feature prioritization, we are making this a community-led effort for the time being so we can keep pushing new features to host :).  We currently offer [one-click self hosting on Render](https://github.com/maybe-finance/maybe/blob/main/docs/self-hosting/render.md) which is the most "beginner friendly" way to self host and our recommended method at the time of writing.

#612 laid the groundwork for self hosting, so adding Docker compose should now have a much clearer path forward.

## Requirements

_Please note, these requirements are a "best guess".  If you find a better way to do something or additional requirements, leave a comment on this issue so they can be added to the list below._

### Tasks

- [ ] `Dockerfile` should be in working state for the Rails app
- [ ] A Github action that publishes `Dockerfile` to GHCR on every commit to `main`
- [ ] A `compose.yml` file that is [production-ready](https://docs.docker.com/compose/production/)
  - [ ] Compose file should reference the published Dockerfile

## Possible future improvements

While some users will be installing Maybe on a self-hosted instance with many other Docker images, other users will be installing it to a fresh VPS.

It _may_ be worth creating an `install.sh` or `prepare.sh` bash script that acts as a "wizard" to configure the machine's environment, things including (but not limited to):

- Identify distro
- Install Docker / Docker compose
- Configure ports, volumes, environment variables
- Let's Encrypt certs
- Spin up container

e-renna:
@zachgoll Might want to reconsider the 2nd end goal. The "docker way" of upgrading an application is to redeploy the container with an updated image. E.g. for a docker stack you would do:
```
docker compose pull
docker compose up -d
```

This would pull the latest docker image from GHCR, and then redeploy the docker stack with it. Therefore, the 2nd goal is not required, but also not recommended.

The folders Maybe Finance saves data to (such as user-uploaded content), should be mapped to the host running the docker daemon (so that they are retained), while the source code of the app and all the dependency should not. 

This means that every time a new version of Maybe Finance is released, the new docker image ships the updated source code along with all the required dependencies. 

This does not only improve stability as the software is shipped with the right dependencies, but also would allow you to reproduce the bug easily as it would allow you to run the same environment the user encountered the bug in. 

If something is not clear, just tag me and I'll be happy to give a better explanation

e-renna:
[...]

PR #640, opened 2024-04-17
Self Hosting: Docker Compose setup and flow

Radu-C-Martin:
My attempt at implementing Issue #627.

I'm for sure missing some stuff, so please don't hesitate to mention it.

The `force_ssl` environment configs are based on [this](https://github.com/maybe-finance/maybe/issues/308) discussion.


zachgoll:
@Radu-C-Martin nice start here!

I think the goal for this issue will be to merge an "all in one" solution that is well tested and validated across the entire self-hosting flow, which will include:

- Successful runs of the Github action + publishing to GHCR (I'm guessing a forked repo is probably the best route here, but open for suggestions)
- Some documentation that walks a self-hoster start to finish using this solution to successfully deploy and access their Maybe instance
  - We'll need a `docker.md` document here - https://github.com/maybe-finance/maybe/tree/main/docs/self-hosting

Any thoughts on how we can test and validate amongst the community?  I'd imagine we'll have quite a few different types of setups to run through.

zachgoll:
I'm going to mark this pull request as a draft and would encourage any other self-hosters to collaborate on this branch to get this to completion.  Would love to hear more thoughts around:

- The best way to test and collaborate on a flow like this
- What self-hosted Docker flows do we need to prioritize and support? (i.e. fresh VPS setup, integrating with HomeLab setups, etc.)
- What is our long-term method for ensuring this flow stays up-to-date as the app evolves

Radu-C-Martin:
I played around with this some more.

I adjusted the deployment workflow to tag not only the complete version, but also a major.minor tag and a major-only tag (this probably would only make sense if breaking changes only happen in major versions?).

I also added a step to prune the untagged images when deploying, since creating a new image for every commit will generate quite a few images :)

Additionally, I updated the documentation a bit to reflect the docker deployment. From commercial VPS I could only test Linode, but it works well there. 

I left the references to my images in the docker-compose file for now, so that people can play around with that and suggest changes. 

One thing to note is that the `latest` tag follows the last image that commit that was tagged, so for example if you first tag a `v.2.1.2`, and then a fix for `v.1.5.6` or something, the latest tag will be on the `v.1.5.6`. But in that case we would probably need more logic in the workflow anyway, so tell me if that's something you want me to consider :)

zachgoll:
Coming along nicely!  Would love to hear from additional self-hosters on their own use-cases and review of this flow.

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+      DB_HOST: "postgres"
+      HOSTING_PLATFORM: "localhost"
+      DISABLE_SSL: "true"
+    depends_on:
+      redis:
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
+
+  redis:
```
DorianMazur, line 21: One question, why do we need redis here?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,121 @@
...
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          flavor: latest=auto
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}.{{minor}}.{{patch}}
+            type=ref,event=branch
+
+      - name: Build and push Docker image
```
DorianMazur, line 104: It would be nice to build docker image for arm64

Radu-C-Martin: That's probably more a question to the core maintainers since that could bring some additional overhead :D

docker-compose.yml
Radu-C-Martin: I based the docker-compose on the devcontainer one, which has redis as well, but as far as I can see it's not actually used?

zachgoll: We won't need Redis for this first pass.  It is not being used within the app currently.

docs/self-hosting.md
```diff
@@ -13,10 +13,66 @@ The fastest way to get your own version of Maybe running is a "one-click deploy"
...
+`DB_PORT`, `POSTGRES_PASSWORD`, `POSTGRES_USER` variables as well
+
+4. Run the `maybe` docker container
+```bash
+docker run -d \
+    --name app \
+    -p 3000:3000 \
+    --restart unless-stopped \
+    --env-file .env \
+    -e RUN_DB_MIGRATIONS_IN_BUILD_STEP=false \
```
zachgoll, line 56: The `RUN_DB_MIGRATIONS_IN_BUILD_STEP` is specific to the Render deployments so I believe it can safely be removed here.

.github/workflows/publish.yml
DorianMazur: It's working perfectly fine -> https://github.com/DorianMazur/maybe/actions/runs/8822599388/job/24221266397

zachgoll: @DorianMazur couple questions:

1. What device do you need this architecture for?
2. Any way we could reduce the build time on this?

DorianMazur: @zachgoll Maybe we could reduce build time by using `macos-latest` runners (arm64 architecture). Because currently in my workflow I am using ubuntu-latest that is x86_64. 

I am working on arm64, I have multiple homeLab servers (each of them is using 5W of power) on arm64 and my Macbook is also arm64. That's why I need this, probably not only me :)

DorianMazur:
I just tested docker compose and it's not working for me :/ 
```
ActiveRecord::ConnectionNotEstablished: connection to server at "172.29.0.2", port 5432 failed: fe_sendauth: no password supplied (ActiveRecord::ConnectionNotEstablished)
Caused by:
PG::ConnectionBad: connection to server at "172.29.0.2", port 5432 failed: fe_sendauth: no password supplied (PG::ConnectionBad)
Tasks: TOP => db:prepare
```

.github/workflows/publish.yml
zachgoll: Gotcha, had forgotten about the M1 chip and Docker compatibility.  I'm guessing we'll have a decent number of self-hosters running this on a Mac, so probably worth supporting.

That said, I think we either need to get that build time significantly reduced _or_ we could change the frequency in which we're running these builds.

I was originally thinking we'd build and publish on every commit to `main`, but it may be more practical to only build and publish for each new release (which we're eventually going to hit a weekly/bi-weekly cycle on).

zachgoll: Not worried about cost—mostly want to be conscious of the overall feedback loop of making modifications to this flow in the future.

If we add this architecture, looks like we can add `runs-on: ${{ matrix.os }}` syntax ([reference](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-a-multi-dimension-matrix)) to speed things up.

.env.example
```diff
@@ -36,6 +36,10 @@ SELF_HOSTING_ENABLED=false
 # `localhost` (or unset) is used for local development and testing
 HOSTING_PLATFORM=localhost
 
+# WIP: Secret key used for stuff?
```
gantoine, line 39: Used by rails directly https://api.rubyonrails.org/v7.1.3.2/classes/Rails/Application.html#method-i-secret_key_base

docker-compose.yml
```diff
@@ -0,0 +1,34 @@
+services:
+
+  app:
+    image: ghcr.io/radu-c-martin/maybe:latest
```
gantoine, line 4: i assume this will change to whoever owns the image?

docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+      POSTGRES_DB: ${POSTGRES_DB:-postgres}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:?}
+    healthcheck:
+      test: [ "CMD-SHELL", "pg_isready -U $$POSTGRES_USER" ]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+volumes:
+  postgres-data:
```
gantoine, line 34: unless you're expecting devs/contributors to use this file to run it locally, i'd rename it to something like `docker-compose.example.yml`

docs/self-hosting.md
```diff
@@ -13,10 +13,61 @@ The fastest way to get your own version of Maybe running is a "one-click deploy"
...
+
+3. Copy the `.env.example` file from the root of the repository and rename it
+   to `.env`
+
+4. Edit the `.env` file and adjust the `SELF_HOSTING_ENABLED` and
+   `SECRET_KEY_BASE` variables
+
+4. Run `docker-compose up -d` to bring up the stack
+
+5. Navigate to http://localhost:3000 in your browser
```
gantoine, line 35: this might be a little _too_ prescriptive, though anyone who know's what they're doing will likely just copy the contents on docker-compose.yml and adapt it

gantoine (commented):
looks good, can't wait to see this merged! just a couples questions/comments

docker-compose.yml
zachgoll: That's a good point—we will _not_ be supporting a dev-optimized compose file since that is already done in `.devcontainer`.  It should be made clear that this should not be used for development.

We should also throw a noticeable comment at the top of the file with some links to the documentation on how to get it running for self-hosting

.env.example
zachgoll: For some additional context, this was originally introduced as an _alternative to_ using the `credentials:edit` for "one-click" Render deploys.  My thinking was that generating this key and saving it in an env var would be _easier_ for self-hosters than going through the process of securing `config/master.key`.

This may or may not be the best way to handle this Docker self hosting flow.  Suggestions around this are welcome!

https://github.com/maybe-finance/maybe/blob/36e8b144868c9b68b23d61c30ac9a04e29f23318/render.yaml#L29-L34

docker-compose.yml
zachgoll: Unless I'm not thinking about this correctly, we'd set this to Maybe's latest published image.

@Radu-C-Martin I'm guessing this was just here so you could test it on your fork correct?

Radu-C-Martin: yes, originally I even pre-setup it with the maybe-finance image name, but then you said you'd like people to test it first, which is a fair point 🤷

.env.example
gantoine: > My thinking was that generating this key and saving it in an env var would be easier for self-hosters

Yes IMO is absolutely is 👍🏼

docs/self-hosting.md
gantoine: i'm a big fan of one-liner scripts (over prescriptive setup instructions) for those less savvy users who will just copy the instructions anyways. makes it less likely they'll mess something up when copying instructions.

> successfully get the project running on a fresh VPS

i wonder what the split between "savvy users who understand docker and compose" and "users running on a cheap vps/rpi" 🤔

maddox:
Am I missing something, or is there not really a docker image actually available yet?

    ERROR: Head "https://ghcr.io/v2/maybe-finance/maybe/manifests/latest": denied

Radu-C-Martin:
> Am I missing something, or is there not really a docker image actually available yet?
> 
> ```
> ERROR: Head "https://ghcr.io/v2/maybe-finance/maybe/manifests/latest": denied
> ```

Not available yet, it will be available once this gets merged. This PR also includes the publish workflow 😅

maddox:
Gotcha, thanks!

zachgoll:
@Radu-C-Martin I'll do some local testing on all of this today/tomorrow and we'll get it merged!

.github/workflows/publish.yml
```diff
@@ -0,0 +1,121 @@
...
+        uses: docker/build-push-action@v5
+        id: build
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Prune all untagged docker images
```
zachgoll, line 113: Could be wrong, but I don't think Github charges anything for _storage_ on GHCR for public repos?

Either way, I think we can remove this final step here for simplicity.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,121 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Context for Buildx
+        run: docker context create builders
+        shell: bash
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          endpoint: builders
```
zachgoll, line 82: Do we need these two steps?  It looks like the `ubuntu-latest` runner already has `Docker-Buildx 0.14.0` installed: 

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#tools

zachgoll (changes requested):
@Radu-C-Martin had some time this afternoon to thoroughly test this.  I was able to get it all running on a Digital Ocean droplet, so I think the sample compose file looks good.

Just left some comments around config and documentation that I think we should address before merging this in.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,121 @@
...
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          flavor: latest=auto
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}.{{minor}}.{{patch}}
+            type=ref,event=branch
```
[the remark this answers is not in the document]
Radu-C-Martin: In my experience, the `latest` tag always points to the latest release. Do you plan for the main branch to always be production-ready?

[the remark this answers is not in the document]
zachgoll: In theory (since we're using trunk-based development), yes, `main` should always be production ready.  If it's not, the checks should fail and the Docker image should not be published.

I would expect `latest` to represent the _most recently published image_, which in our case, would always be the latest commit.

But there probably is some value in providing an alias for self-hosters who want to use the "latest release".  What do you think of using the following tag scheme?

- `latest` - the latest published image (i.e. `main`)
- `stable` - the latest published release
- `[release_semver_tag]` - the semver tag of the release
- `[commit_sha]` - the commit sha of the docker image so self-hoster can lock into a certain version

So something like:

```yml
tags:
  # Explicit commit sha and version tags
  type=sha
  type=semver,pattern={{version}}

  # Aliases
  type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
  type=raw,value=stable,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```

docker-compose.example.yml
```diff
@@ -0,0 +1,34 @@
+services:
+
+  app:
+    image: ghcr.io/maybe-finance/maybe:latest
+    ports:
+      - 3000:3000
```
Quintasan, line 6: I believe this should be:
```suggestion
      - "127.0.0.1:3000:3000"
```
so the Rails backend is not exposed to the internet without a reverse proxy in front of it

config/environments/production.rb
```diff
@@ -40,10 +40,9 @@
 
   # Assume all access to the app is happening through a SSL-terminating reverse proxy.
   # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
-  # config.assume_ssl = true
-
+  config.assume_ssl = ENV["DISABLE_SSL"].blank?
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  config.force_ssl  = ENV["DISABLE_SSL"].blank?
```
Quintasan, line 45:
```suggestion
  config.force_ssl  = ENV.include?("FORCE_SSL")
```

I think the default assumption is that you're using a reverse proxy to terminate incoming HTTPS traffic.

config/environments/production.rb
```diff
@@ -40,10 +40,9 @@
 
   # Assume all access to the app is happening through a SSL-terminating reverse proxy.
   # Can be used together with config.force_ssl for Strict-Transport-Security and secure cookies.
-  # config.assume_ssl = true
-
+  config.assume_ssl = ENV["DISABLE_SSL"].blank?
```
Quintasan, line 43:
```suggestion
  config.assume_ssl = ENV.include?["ASSUME_SSL"]
```

docker-compose.example.yml
gantoine: I don't think that's needed, and even then if you want to stop it being exposed to the host (not internet, since that would require port-forwarding) just don't bind the ports and use a reverse proxy container.

config/environments/production.rb
zachgoll: We've been back and forth on this one quite a bit, but as I was going back through prior discussions, found this comment, which IMO is the clearest solution offered to this:

https://github.com/maybe-finance/maybe/issues/308#issuecomment-2081522347

config/environments/production.rb
zachgoll: Resolving this so we consolidate into one comment.

docker-compose.example.yml
Quintasan: >being exposed to the host (not internet, since that would require port-forwarding)

@gantoine I think your vision of how this works is incorrect, especially if you're talking about servers that are exposed to the internet directly.

`3000:3000` is equivalent to `0.0.0.0:3000:3000` which means `listen on port 3000 on ALL interfaces` so if your server has an interface that's exposed to the internet (literally any VPS) then it will accept incoming connections on port 3000 from virtually anyone. That's not a safe/sane default.

> just don't bind the ports and use a reverse proxy container

That's another assumption that's going to make it harder for anyone interested in self-hosting, since your reverse proxy container will listen on ports 80 and 443, and you will have to put any/all subsequent containers in the same Docker network and configure the reverse proxy serving your Maybe instance to serve your other containers or use some convoluted setup involving Traefik and labeling stuff.

I, for example, have a Caddy on the host doing all the reverse proxying to all containers I have on my server like Immich, Maybe, Nextcloud and a ton of other stuff.

zachgoll: Just to confirm—you're optimizing for a self-hosting scenario here where you're running this app on your _local_ machine and _never_ want it exposed to the internet?

Quintasan: @zachgoll Me? No, I'm optimizing for any sane self-hosting solution - there's a reverse proxy in front of the Rails application. If we put a reverse proxy in the Docker Compose configuration and the self-hosting user has some other services on the server, then they have two options.

1. Modify the provided `docker-compose.yml` so the reverse proxy doesn't listen on ports 80 and 443 and point their preferred reverse proxy to... reverse proxy to our reverse proxy
2. Dump all their other containerized services into the same Docker network and reconfigure our reverse proxy to work with their services.

Maybe there's some convoluted setup leveraging Treafik, but that's not we are aiming for

Quintasan: What we *could* do is have an additional `docker-compose.override.yml` sample which has a reverse proxy configured inside. Then we can point everyone who doesn't particularly care about those details to use alongside the default `docker-compose.yml`

WDYT?

gantoine: @Quintasan Your comment is perfectly reasonable, it's just that I've never seen an example docker-compose bind the network interface (in like, any self-hosted software I use or have tried).

> literally any VPS

I'm not thinking about the VPS case, I'm more-so thinking about the user who's first getting started with self hosting on a small server/old laptop/RPI at home.

> you will have to put any/all subsequent containers in the same Docker network and configure the reverse proxy serving your Maybe instance to serve your other containers

That's how I have it setup at home, and it works well pretty well 🤷🏼 

> you're optimizing for a self-hosting scenario here where you're running this app on your local machine and never want it exposed to the internet?

A machine running in my home, accessible via tailscale or cloudflare tunnel.

zachgoll: @Quintasan @gantoine thanks for this additional detail, this is helpful.

While I think both of you are probably stronger than me on the config details here, here's the general direction I was thinking:

- `docker-compose.example.yml` - this should represent a setup that intersects the 1) most common and 2) most "beginner friendly" setup.  It should _nearly_ be an "out of the box" copy/paste solution, or as close as we can get.
- `docker-compose.[specific-setup].example.yml` - we could introduce variations of the default as sample files OR add instructions to our documentation

config/environments/production.rb
Quintasan: > it's just that I've never seen an example docker-compose bind the network interface (in like, any self-hosted software I use or have tried).

@gantoine Fair enough. I'm not really going to say that `everyone does it so it's a good idea` is ever going to convince me, especially once you realize how much automated traffic scanning is going on nowadays. Exposing a web application directly is usually a bad idea considering that webservers like Puma, Webrick etc. are not well-equipped to serve static assets.

@zachgoll I think we should provide a `docker-compose.yml` in the repository and a `docker-compose.override.yml` example in the documentation. By default, Docker Compose uses both `docker-compose.yml` and `docker-compose.override.yml` when you invoke `docker compose up -d` so we can support two scenarios out of the box with no convoluted tinkering required:

1. Bring your own reverse proxy - in this case the self-hoster has just to use the `docker-compose.yml`  in the repository and configure the reverse proxy on their own
2. `I don't care just let me run this` - in this case, we would ask the self-hoster to additionally create `docker-compose.override.yml` in the repository which would set up the reverse proxy for them, so literally no work would be required. Just copy both files and run `docker compose up -d` and you're ready to go.

In both scenarios, binding the backend port to the loopback interface (`127.0.0.1`) only prevents external connections from accessing the backend directly on port 3000.

If someone **actually** wants to expose the backend directly, then they can use `docker-compose.override.yml` to override our default binding.

docker-compose.example.yml
csmith1210: I’m just a self-hosting bystander here, but I would tend to agree with @gantoine. I’ve never seen Docker compose set ups specifically bind the network interface. And, even on a VPS set up, you would need to specifically allow that port in the firewall; at least that’s how it has been for any fresh box I have installed.

I also think if most people are using a Docker set up, there’s no real need to “dumb it down”. They should already understand the risks with simple networking concepts like ports and how to securely access your apps over the internet. I think the Render deploy option serves that purpose well.

Quintasan: A very, very simplified example:

```
# docker-compose.yml
# Ours will be slightly more verbose
services:
  backend:
    image: traefik/whoami
    command:
      - --port=6666
    ports:
      - "127.0.0.1:6666:6666"
```

```
# docker-compose.override.yml
services:
  caddy:
    image: caddy:alpine
    command: caddy reverse-proxy --from <domain to serve on> --to backend:6666
    ports:
      - 80:80
      - 443:443
```

If you `docker-compose up -d` with those two files, then you end up with a backend container and a reverse proxy for it with automatic HTTPS. We can supply a custom Caddyfile OR use Traefik as the reverse proxy if we need a more complicated setup

Quintasan: >I’ve never seen Docker compose set ups specifically bind the network interface. And, even on a VPS set up, you would need to specifically allow that port in the firewall; at least that’s how it has been for any fresh box I have installed.

All OVH and Hetzner boxes I have had so far accept all connections on all ports, unless you explicitly configure them not to do so. Assuming your boxes don't (which is a good thing), binding to `127.0.0.1:<port>` changes exactly nothing in your case, while being a sane default for all boxes which don't have a firewall. I remain unconvinced that binding to `0.0.0.0` and exposing the backend container directly to the internet is a good idea.

> I also think if most people are using a Docker set up, there’s no real need to “dumb it down”.

I don't think we're dumbing anything down here. We're trying to provide sane and secure defaults while giving people option to `just run it` without any tinkering. If someone wants to tinker with this setup, then they have all the escape hatches (`docker-compose.override.yml` and the ability to bring their own reverse proxy)

gantoine: > binding to 127.0.0.1:<port> changes exactly nothing in your case, while being a sane default for all boxes which don't have a firewall.

This is a fair point, and as it personally wouldn't affect me (as I know how to edit a compose file) I think it's fine to set it that way in the default. 👍🏼

.github/workflows/publish.yml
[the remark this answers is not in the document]
adyanth: As long as the tags are semvar compatible, the existing tag scheme is very nice to be able to select an upgrade schedule using auto image pulls. One can have `maybe:latest` which always auto upgrades (and should not be done), while one can do `maybe:2` or `maybe:2.3` and get the patch upgrades.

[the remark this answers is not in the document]
Radu-C-Martin: That's how I've seen it done in other projects. It's intuitive, and also works well with renovate/dependabot. But I guess that would impose stricter adherence to semver versioning to be totally useful, so it's more for the core maintainers to decide. It can always be added later :D

docker-compose.example.yml
zachgoll: I think @Quintasan's proposal here makes sense overall:

1. Give a sane default that "just works"
2. Provide escape hatches for those with the knowledge and/or motivation to customize

zachgoll:
@Radu-C-Martin this looks good, thanks for the changes!  The documentation and overall structure is good with me.

If I'm not mistaken, the only pending item we have left is the reverse proxy config and the "override" file mentioned by @Quintasan.  I think this may be a good follow-up PR.

Unless there are objections, I think we can merge this as-is so we can start getting some images published to GHCR for self-hosters to start playing around with.

zachgoll (approved):

Radu-C-Martin:
Hey, Zach! Thanks for merging my PR 🙂 I tried accessing the image and apparently I'm not authorized. I think there could be some settings missing in the organization settings? (this is what I found, but could be out of date: https://github.com/orgs/community/discussions/26014, I've never had to deal this yet). I see that the CI passes correctly, so I assume for you the image works at least?

zachgoll:
@Radu-C-Martin thanks for the heads up, we'll get the org setting updated shortly.

zachgoll:
@Radu-C-Martin should be public now, are you able to access?

Radu-C-Martin:
All good now!

Issue #643, opened 2024-04-18
Improvement: D3 Line Chart Styles and Formatting


zachgoll:
![CleanShot 2024-04-18 at 11 18 01](https://github.com/maybe-finance/maybe/assets/16676157/7c8c2196-5a19-4660-ac79-7282a43a2dbc)

## Designs

Below are the designs you should follow while implementing this:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2770-226&mode=design&t=Y0hF2UeUXaOhxGL4-0

## Overview

The current D3 line chart (`line_chart_controller.js`) we are using is lacking several design aspects and features.  This issue is dedicated to getting this chart to a near-"final" state.

## Requirements

_If there is a missing / incorrect requirement, please leave a comment before starting work on this._

### Functional Requirements

- [ ] Currently, we have a `trendline_controller.js` and a `line_chart_controller.js`.  We should consolidate these into a single `line_chart_controller` that is configurable.  Options will include:
  - [ ] Hide or display x-axis labels ("trendlines" have no axis) - show by default
  - [ ] Enable or disable tooltips ("trendlines" have no tooltip) - enable by default
- [ ] The chart controller will always receive a `TimeSeries.new([]).to_json` value, which can be an instance of `Money` or a regular integer (this is already implemented).  As part of this issue, we need to make sure the chart can handle both `Money` values and regular integers (including percentages) in the tooltip.
- [ ] When the chart is supplied with a **single value** or an "empty" Time Series (i.e. `TimeSeries.new([])`), it should display the "empty" state design.

### Design Requirements

- [ ] Implement gradient underneath the line (very subtle, see designs)
- [ ] Ensure padding and spacing matches designs
- [ ] When the _overall_ trend of the series that is passed to the line chart is "bad" (decreasing for an asset, increasing for a liability), the chart should go from a green color to a red color.
- [ ] When hovering over a data point in the chart, the line should highlight for dates prior to it, and the line after it should be grayed out

## Implementation Suggestions

_Below are some ideas for implementation to get you started.  Use your best judgment here—if there's a better way to do things, go for it!_

- Be sure to read through [time series code](https://github.com/maybe-finance/maybe/tree/main/app/models/time_series) as it will help conceptualize things for this issue
[...]

PR #651, opened 2024-04-19
Refactor `TimeSeries` artifacts

josefarias:
I did this refactoring as an exercise for getting oriented before tackling https://github.com/maybe-finance/maybe/issues/643 and, well, for practice.

I think the resulting code is more robust and easier to follow. Here's what I did:

1. Detangle responsibilities. All classes are still there, but they're less aware of each other's concerns, even if not perfect yet. For example, each class knows how to turn itself into json, and `TimeSeries::Value` knows how to calculate its own trend, rather than `TimeSeries` orchestrating all of it.
2. Make use of Ruby and Rails idioms to make the code a bit more terse and easier to follow to someone who would recognize these patterns.
3. Reindent a couple of files to 2 spaces, as is the Ruby convention.
4. Replace the notion of a TimeSeries `type` with `favorable_direction`. This new name is longer, but communicates exactly what it's for as opposed to "type" which could mean anything. This already uncovered a bug in the `#trend_styles` helper!
5. Remove some unused codepaths, like `TimeSeries::Value` knowing how to destructure an object when we're only ever passing hashes to it.
6. Removing duplicate knowledge like what the `favorable_direction` is. As far as I can tell, trends and values can only exist within a series — and it's the nature of the series which determines what the `favorable_direction` is. So trends now delegate that knowledge to its containing series. This is easily changed if we ever want the two to diverge.

I'm mostly submitting this PR because I did the work recreationally and I'm presenting it for what it's worth. There's more work to be done here (two big smells are how often we're calling `#is_a?` and how often we explicitly need to handle `nil`). But I think this is a step in the right direction and I had to draw the line somewhere.

I think this would be an improvement that makes the campsite cleaner. But the value of a refactor is subjective and I'm happy to close this if this is not something that warrants focus right now. Just didn't want to throw away the work when it might (or might not) be valuable.

All tests are passing. If this breaks anything I'm sure it'd be an easy fix. We're at a stage where we can make this kind of change, I think.


zachgoll:
@josefarias thanks for the improvements here, I think these are certainly worth merging into `main`.

While I know this doesn't change the public interface, I'll wait until the bounty is over to merge these changes so as not to disrupt any others working with this code right now.  If you end up submitting a bounty solution, you're more than welcome to include these changes if it makes things easier.  Up to you.

RE the changes: Overall, I agree with the changes.  I have a few ideas around the naming of `Trend.direction` and copious `is_a?` calls and `nil` checks.  Would you be okay if I opened a PR against your branch with some ideas?

zachgoll:
@josefarias to avoid blocking the bounty PR and to align with your suggestions, I'm going to go ahead and merge this as-is.  I think these changes are good improvements to the existing design, so no reason to block a merge on them.

I will tag you in any follow-up PRs that I open in relation to this.

PR #653, opened 2024-04-19
Add partial account sync support

### LINKED ISSUE #584: Implement "Partial Account Sync" Logic
## Overview 

Currently, the app can call `@account.sync_later` ([see this file](https://github.com/maybe-finance/maybe/blob/7fa77b4fd7ecf9b51f16f335ee7c67dd7f2b0282/app/models/account/syncable.rb#L8-L21)), which will do the following:

1. Fetch and save exchange rates needed to sync the account
2. Search through the account's valuations and transactions and _accumulate_ daily balances
3. Save balances in both the account and family currency (if needed)

This sync process is triggered on every Create/Update/Delete of both `Transaction` and `Valuation` right now, which means even if a `Valuation` dated _yesterday_ is updated, the job will sync _all_ historical balances.

The goal of this issue is to implement "partial" syncs given a `start_date` so we can only calculate and update balances that were affected by the change (i.e. if I change the balance of a `Transaction` yesterday, we only need to sync the last 2 days balances)

## Requirements

- [Sync](https://github.com/maybe-finance/maybe/blob/7fa77b4fd7ecf9b51f16f335ee7c67dd7f2b0282/app/models/account/syncable.rb#L8-L21) should take a `start_date` param
- On `CREATE` and `UPDATE` of both `Valuation` and `Transaction`, the `start_date` is equal to the date of the `Valuation` or `Transaction`
- On `DELETE` of `Valuation`/`Transaction`, the `start_date` will be equal to the _prior_ record's date

jakubkottnauer:
Closes #584

### DIFF
--- a/test/controllers/valuations_controller_test.rb
+++ b/test/controllers/valuations_controller_test.rb
@@ -21,7 +21,7 @@ class ValuationsControllerTest < ActionDispatch::IntegrationTest
   test "create should sync account with correct start date" do
     date = Date.current - 1.day
 
-    assert_enqueued_with(job: AccountSyncJob, args: [@account, date]) do
+    assert_enqueued_with(job: AccountSyncJob, args: [ @account, date ]) do
       post account_valuations_url(@account), params: { valuation: { value: 2, date:, type: "Appraisal" } }
     end
   end
@@ -34,12 +34,12 @@ class ValuationsControllerTest < ActionDispatch::IntegrationTest
 
   test "update should sync account with correct start date" do
     new_date = @valuation.date - 1.day
-    assert_enqueued_with(job: AccountSyncJob, args: [@account, new_date]) do
+    assert_enqueued_with(job: AccountSyncJob, args: [ @account, new_date ]) do
       patch valuation_url(@valuation), params: { valuation: { account_id: @valuation.account_id, value: @valuation.value, date: new_date, type: "Appraisal" } }
     end
 
     new_date = @valuation.reload.date + 1.day
-    assert_enqueued_with(job: AccountSyncJob, args: [@account, @valuation.date]) do
+    assert_enqueued_with(job: AccountSyncJob, args: [ @account, @valuation.date ]) do
       patch valuation_url(@valuation), params: { valuation: { account_id: @valuation.account_id, value: @valuation.value, date: new_date, type: "Appraisal" } }
     end
   end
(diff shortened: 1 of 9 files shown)


jakubkottnauer:
@zachgoll Ready for review

test/controllers/valuations_controller_test.rb
```diff
@@ -3,17 +3,107 @@
...
+    date = @valuation.date
+
+    account.sync
+    account.balances.where(date: date - 10.day).update!(balance: 200)
+
+    assert_no_changes("account.balance_on(date - 10.day)") do
+      delete valuation_url(@valuation)
+      perform_enqueued_jobs
+    end
+  end
```
zachgoll, line 108: So looking through the tests outlined in `valuations_controller_test` and `transactions_controller_test`, I think the test cases themselves are useful to have, but they probably don't belong in a controller test.  It looks like the questions we're asking here are:

1. Did the job happen?
2. Did the job happen with correct start date?
3. Did running the job result in the correct end state?

I think the first and the second assertions are correct to be here in the controllers, but the third question should probably be tested in `syncable_test` as it asserts against the logic that is handled by the `Account` model.

We could probably consolidate all the controller tests so that for each CRUD controller action, we assert that:

1. The resource was edited/created/deleted 
2. The sync job was enqueued with the correct params (we can use [assert_enqueued_with](https://guides.rubyonrails.org/testing.html#testing-jobs-in-context) for this).

That way, we're only testing the responsibility of the controller, which is to ensure the job is successfully enqueued with the correct args.

test/models/account/syncable_test.rb
```diff
@@ -1,38 +1,81 @@
...
+  end
+
+  test "balance on the sync date is updated after syncing" do
+    account = accounts(:savings_with_valuation_overrides)
+    balance_date = 5.days.ago
+    account.balances.create!(date: balance_date, balance: 1000)
+    account.sync balance_date.to_date
+
+    assert_equal 19700, account.balances.find_by(date: balance_date)[:balance]
+  end
```
zachgoll, line 58: In reference to my comment above about the controller tests, I think what you have here is good and covers the core logic that we're trying to validate.

zachgoll (commented):
Thanks for tackling this one!  Definitely a tricky flow.  

I think the code looks good overall, just added a few comments around the location and content of the tests.

test/controllers/valuations_controller_test.rb
jakubkottnauer: Thanks for the great feedback, I agree!

jakubkottnauer: I've updated the tests

zachgoll (approved):
Awesome, thanks for the updates.

Issue #625, opened 2024-04-15
Shared Money input should respond to change in currency


zachgoll:
## Requirements

- When currency changed, placeholders and other currency-specific information should update in the input (will need JS controller)
- `Money::Currency` has methods like `step` for easier calculation of placeholder values

Please see [this comment thread](https://github.com/maybe-finance/maybe/pull/614#discussion_r1565245260) for more info.

PR #654, opened 2024-04-19
Shared money input should respond to change in currency

ThibautGrx:
closes https://github.com/maybe-finance/maybe/issues/625


ThibautGrx:
This is my proposal for this issue. It's maybe a bit overkill and too specific. Any though on that ?

zachgoll (approved):
Nice work on this!  I'll leave this open for a few more hours for others to take a look at, but personally, I can't think of a better way to do this than what you've outlined in this PR.

Our currencies list is so large I think your idea of introducing a simple getter endpoint works well 👍

PR #658, opened 2024-04-20
Add transaction destroy balance sync test

jakubkottnauer:
I think I may have found a bug in the `Account::Balance::Calculator::implied_start_balance` method while working on #584. This PR contains a fix and a test that passes after the fix has been implemented. Feel free to close this if I got things wrong 😄 

At the moment, to get the start balance in a situation where there aren't any valuations, we sum amounts of transactions occurring  _after_ the `calc_start_date`. Shouldn't that be the other way round?

For example, I would expect that if we're computing balances between 2024-04-01 and today (and we don't have any valuations entries), the balance on the first day (2024-04-01) should be equal to `@account.balance + sum(transactions before 2024-04-01)`, with `@account.balance` being the "initial" balance of the account (set when the account was created).

There are now a bunch of other tests failing that depend on the `implied_start_balance` method, I will fix those after I have received confirmation that this really is a bug.

### DIFF
--- a/app/models/account/balance/calculator.rb
+++ b/app/models/account/balance/calculator.rb
@@ -100,7 +100,8 @@ class Account::Balance::Calculator
           oldest_valuation = normalized_valuations.find { |v| v["date"] == oldest_valuation_date }
           oldest_valuation["value"].to_d
         else
-          net_transaction_flows = normalized_transactions.sum { |t| t["amount"].to_d }
+          prior_transactions = normalize_entries_to_account_currency(@account.transactions.where("date < ?", @calc_start_date).select(:date, :amount, :currency), :amount)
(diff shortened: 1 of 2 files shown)


zachgoll:
@jakubkottnauer I think your understanding is near correct here except for the assumption that `account.balance` represents the "origin" balance on the account.

Instead, `account.balance` should always represent the "current" balance on the account, which aligns with how many external data providers supply account data (i.e. Plaid).

To create a "historical" account balance, one would create a `Valuation` on a certain date.

#501 is a good resource for seeing how the expected results were derived, and I'd also look through the expected values in [this Google Sheet](https://docs.google.com/spreadsheets/d/18LN5N-VLq4b49Mq1fNwF7_eBiHSQB46qQduRtdAEN98/edit#gid=0) which provides a more visual representation of our expected results.

PR #668, opened 2024-04-23
Add income and spending insight cards to dashboard

### LINKED ISSUE #645: Feature: Implement income and spending insight cards for dashboard
CleanShot 2024-04-18 at 11 37 41

## Feature Overview

This issue is for completing the "income" and "spending" insight cards that are located on the dashboard.

## Designs

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2770-1101&mode=design&t=Y0hF2UeUXaOhxGL4-0

## Requirements

- [ ] Should generate a time series of income and spending for a `Family`, scoped to `Period.last_30_days` (see "suggestions" section below for more detail)
  - [ ] Each data point in the series represents a "Trailing 30 day sum" of income (negative transaction values)
  - [ ] The "headline" number shows the last value in the series
  - [ ] The "trend" value is the overall time series trend
- [ ] Should find the top 3 accounts with the _most_ income or spending within the last 30 days and display the sum
  - [ ] Each account is an anchor that goes to that account page
  - [ ] Bank icons are not in-scope (just use a generic letter as shown in designs)
  - [ ] A "more" counter should show if there are more than 3 accounts
- [ ] Display a trendline of series.  No tooltips, no axes.
- [ ] These insights should be included in the dashboard - https://github.com/maybe-finance/maybe/blob/f5f624881f3b8900fa56d655ec040b5ee5d8a0af/app/views/pages/dashboard.html.erb#L35-L68

## Implementation Suggestions

- Be sure to read through [time series code](https://github.com/maybe-finance/maybe/tree/main/app/models/time_series) as it will help conceptualize things for this issue
- **Trailing 30 day calculation** - to further expand on this calculation, here are two examples of what this represents.  For these examples, assume the current date is `Apr 18 2024` and we are calculating the "Income" card:
  - Last value in graph (and "headline" number) represents the total income earned from `Mar 19 2024` to `Apr 19 2024` 
  - First value in graph (e.g. the "vs last month" trend comparison) represents total income earned from `Feb 18 2024` to `Mar 19 2024`

## Reminders

## Bounty
There is a $100 bounty on this. To participate in the bounty:

1. Comment `/attempt #645` to indicate you plan on working on this.
2. Submit a draft PR once you start and include `/claim #645` along with your general game plan for how you intent to implement this.

JoshAntBrown:
This PR implements the income and spending insight cards for the dashboard.

Screenshot 2024-04-23 at 14 50 25

## Requirements
* [x]  Should generate a time series of income and spending for a `Family`, scoped to `Period.last_30_days`
  * [x]  Each data point in the series represents a "Trailing 30 day sum" of income (negative transaction values)
  * [x]  The "headline" number shows the last value in the series
  * [x]  The "trend" value is the overall time series trend
* [x]  Should find the top 3 accounts with the _most_ income or spending within the last 30 days and display the sum
  * [x]  Each account is an anchor that goes to that account page
  * [x]  Bank icons are not in-scope (just use a generic letter as shown in designs)
  * [x]  A "more" counter should show if there are more than 3 accounts
* [x]  Display a trendline of series.  No tooltips, no axes.
* [x]  These insights should be included in the dashboard

/claim #645
Resolves #645


app/models/family.rb
```diff
@@ -28,6 +28,78 @@ def snapshot(period = Period.all)
...
+        date: r.date,
+        value: Money.new(r.rolling_income, r.currency)
+      }
+    end
+
+    {
+      income_series: TimeSeries.new(income, favorable_direction: "up"),
+      spending_series: TimeSeries.new(spending, favorable_direction: "down")
+    }
+  end
```
zachgoll, line 101: I like the overall direction of this query.  Definitely a complex one (that will likely need to be modified as we start adding various transaction types), so I'm thinking we may need a few test cases for documentation purposes.  As a quick summary of this comment:

- Series length should always be 30 days, regardless of how many currencies a `Family` has
- We'll need to handle exchange rate conversions for this rollup
- Consider using a CTE for clarity of the query? - Rails 7 has the `with` method that _may_ make this a little easier to scan top-to-bottom?

Looking at the output for the demo account (`rake demo_data:reset`):

CleanShot 2024-04-23 at 10 53 17

In the screenshot above, I would expect these insights to _always_ show in the `Family.currency` (the demo account is `USD`, but one of the top moving accounts is `EUR`).  Also, it appears that both currency series from the query are being included, causing the count to be `62` rather than the intended `31`.

To handle multiple currencies, I'm thinking we'll need an _additional_ step that will normalize everything into the `Family.currency` prior to summing it all up
  - Since we cannot _depend on_ exchange rates being available, we'll either need to _omit_ unconverted amounts from the aggregation or omit accounts that contain "foreign" currencies from the calculation entirely.
  - My guess is that it would be easiest to perform a join on `exchange_rates` table somewhere within this query and do the filtering there (rather than trying to call `exchange_to` on each money instance after-the-fact).  In this strategy, we'd basically be saying, "we'll try to convert, but if the app doesn't already have the needed exchange rates cached to the DB, we can't calculate this and will omit it"
  - We could still technically show a "top mover" in a native foreign currency (since we have a series for all currencies), but series that cannot be converted via the query should never make it into that headline number

I've been using [this spreadsheet](https://docs.google.com/spreadsheets/d/18LN5N-VLq4b49Mq1fNwF7_eBiHSQB46qQduRtdAEN98/edit#gid=0) to visualize the expected results and conversions a little bit better.  Just throwing it out there in case it helps with this one.

As this is a tricky one, I'm open to ideas!  All the comments above are just my best guess at how we could approach it.

----------

Here's a _sketch_ of what I was thinking for the exchange rates (haven't tested thoroughly, using some hardcoded placeholders):

```sql
WITH date_series AS (
	SELECT * 
	FROM 
		generate_series('2024-02-23'::date, '2024-04-23'::date, interval '1 day') AS gs(date) 
		CROSS JOIN (SELECT DISTINCT currency FROM transactions) as c
), normalized AS (
	SELECT
        gs.date,
        gs.currency,
        COALESCE(SUM(t.amount) FILTER (WHERE t.amount > 0), 0) AS spending,
        COALESCE(SUM(-t.amount) FILTER (WHERE t.amount < 0), 0) AS income
    FROM
        date_series gs
        LEFT JOIN transactions t ON t.date = gs.date AND t.currency = gs.currency
    GROUP BY gs.date, gs.currency
), rolling AS (
	SELECT 
		nq.*,
		SUM(spending) OVER (PARTITION BY currency ORDER BY date RANGE BETWEEN INTERVAL '30 days' PRECEDING and CURRENT ROW) as rolling_spend,
		SUM(income) OVER (PARTITION BY currency ORDER BY date RANGE BETWEEN INTERVAL '30 days' PRECEDING and CURRENT ROW) as rolling_income
	FROM normalized nq
	ORDER BY nq.date
), converted_rolling AS (
	SELECT
        r.date,
        COALESCE(r.rolling_spend * er.rate, 0) AS converted_rolling_spend,
        COALESCE(r.rolling_income * er.rate, 0) AS converted_rolling_income
    FROM
        rolling r
        LEFT JOIN exchange_rates er ON r.currency = er.base_currency AND er.converted_currency = 'USD'
    GROUP BY r.date, er.rate, r.rolling_spend, r.rolling_income
) SELECT * FROM converted_rolling;
```

IMO, these CTEs read a little easier but I know it completely deviates from ActiveRecord convention.  Not sure what the best approach is?

zachgoll (commented):
The structure of this looks good to me, just added some notes regarding tests / exchange rate conversions for handling multi-currency.

app/models/family.rb
```diff
@@ -28,6 +28,77 @@ def snapshot(period = Period.all)
...
+
+    {
+      top_spenders: results.sort_by { |a| a.spending }.filter { |a| a.spending > 0 }.reverse,
+      top_earners: results.sort_by { |a| a.income }.filter { |a| a.income > 0 }.reverse
+    }
+  end
+
+  def snapshot_transactions
+    period = Period.last_30_days
+    days_rolling = 30
```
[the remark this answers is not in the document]
JoshAntBrown: Yep makes sense to me, we can just count the days in period we're using and then use that to build the query. I'll get that updated.

app/models/family.rb
JoshAntBrown: Thanks! Yeah it's been a tricky one, that makes sense regarding the currencies I was mirroring what was going on with the net worth chart but this makes more sense.

I'll take a look at working the currency conversion into things, and while I do that I'll see if I can do anything to improve the readability.

zachgoll: Sounds good.  Yeah, the NW chart benefits from having the `account_balances` table to read from which is "normalized" (which includes currency conversion) during the account sync process, which makes it easier to just filter by the family's currency (and assume any account balances that aren't in this subset couldn't be converted).

This one is definitely tricky since we have to normalize and convert on-the-fly

zachgoll: @JoshAntBrown I had previously experimented around with a DB table called `family_snapshots` that would essentially act as a daily "cache" table for these sorts of calculations (similar to `account_balances`):

```
date, net_worth, assets, liabilities, rolling_30_day_income, rolling_30_day_expense
```

We'd essentially be doing the same calculations here but rather than on-the-fly we'd have a daily cron running to calculate them and then here we'd just read the values.

Not sure if that would really make any of this easier but thought I'd float the idea in case it helped with this one at all.

JoshAntBrown:
I've had a bit of a refactor and added a test in a similar fashion to the other account snapshot.

Hopefully this is a bit more manageable now without straying too far from Active Record conventions.

Here's my copy of the test cases, updated to include spending and income on a rolling basis:
https://docs.google.com/spreadsheets/d/112tRiy0l8WWngEGMdEX2JprYJmYWUW--B2UDUiW9nOE/edit

zachgoll (approved):
Looks great, thanks for adding those test cases!  I have copy/pasted the expected results into that main spreadsheet for now and we'll eventually get all those into an easier format to collaborate on.

PR #673, opened 2024-04-25
Refactor placeholder logo into common controller

JoshAntBrown:
We're starting to see a lot of duplication around account logo placeholders, and there's already a couple of inconsistencies.

This PR adds a way of rendering an SVG placeholder logo on a per account basis so that we can simply just render the SVG.

This can be updated in the future to redirect/pull from third party sources in future, with the option of always falling back to the placeholder if there are any failures.

### DIFF
--- a/app/controllers/accounts/logos_controller.rb
+++ b/app/controllers/accounts/logos_controller.rb
@@ -1,0 +1,10 @@
+class Accounts::LogosController < ApplicationController
+  def show
+    @account = Current.family.accounts.find(params[:account_id])
+    render_placeholder
+  end
+
+  def render_placeholder
+    render formats: :svg
+  end
+end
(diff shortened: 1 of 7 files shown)


zachgoll:
@JoshAntBrown I wasn't saying we'd be storing a placeholder in ActiveStorage, we'd just do something like this:

`views/accounts/_logo.html.erb`:

```erb
<%# locals: (account:, size:) %>

<% if account.logo.attached? %>
  <%= image_tag account.logo, size: size, alt: "#{account.name} Logo" %>
<% else %>
  Custom placeholder SVG
<% endif %>
```

The Campfire approach makes sense to me for user-uploaded content like avatars where you'd be getting various sizes and cache lifetimes, I guess I was just questioning whether we'd need all that fine-grained control and whether we'd be okay with the number of network requests this introduces (which may be fine).

Either way, my main concern is making sure we can easily keep this aligned with our "provided" 3rd party data strategy, which looks like we can.

PR #682, opened 2024-04-26
Allow a self-hosted user to configure their SMTP settings directly from within the UI

### LINKED ISSUE #678: Feature: Configure SMTP for both Self Hosted and Hosted Versions of Maybe
CleanShot 2024-04-25 at 14 56 04

## Feature Overview

The goal of this feature is to:

- Allow a self-hosted user to configure their SMTP settings directly from within the UI to maintain our "zero-config" self hosting setup.
- Allow the _hosted_ version of Maybe to send emails

## Requirements

- [ ] A self-hosted user can configure their SMTP settings
- [ ] A self-hosted user can send a test email to verify their settings
- [ ] Maybe Finance (the company) can still configure SMTP settings through environment variables for our hosted version of the app
- [ ] Tests are written where appropriate

## Implementation Suggestions

- In order to show the self host settings, make sure that you set `SELF_HOSTING_ENABLED=true` in your local `.env` file
  - `/settings/hosting` is where the placeholder form is located
- The Rails Settings Cached gem gives [an example of SMTP settings here ](https://github.com/huacnlee/rails-settings-cached?tab=readme-ov-file#readonly-field)and a [suggested implementation](https://github.com/ruby-china/homeland/blob/main/app/mailers/application_mailer.rb#L19) (to override the default ActionMailer)
  - This _may not_ be the _best_ way to do things.  Other ideas are welcomed.
- The existing placeholder fields in the self host settings are a _starting point_.  We _may_ need to add/remove fields to get this working.

## Designs

The designs for this have already been implemented.  This issue should solely focus on _functionality_.

## Reminders

## Bounty
There is a $200 bounty on this. To participate in the bounty:

1. Comment `/attempt #678` to indicate you plan on working on this.
2. Submit a draft PR once you start and include `/claim #678` along with your general game plan for how you intent to implement this.

ThibautGrx:
This PR enables SMTP settings configuration directly from within the UI.

https://github.com/maybe-finance/maybe/assets/35230028/70a2d39d-db15-4924-a9d5-6b70709acc30

Dynamic SMTP settings is inspired from exemples given in the issue , and from https://guides.rubyonrails.org/action_mailer_basics.html#sending-emails-with-dynamic-delivery-options

I've added `email_sender` setting to define email sender ( even if it's not a smtp settings)
The test email is very simple, but it can be changed if you have any suggestions. It's sent to `Current.user.email`.

"Coming soon" displayed in the video has been removed.

/claim #678
Resolves #678


app/controllers/settings/hostings_controller.rb
```diff
@@ -17,6 +17,17 @@ def update
     end
   end
 
+  def send_test_email
+    current_user_email = Current.user.email
+    begin
+      NotificationMailer.test_email(current_user_email).deliver_now
```
zachgoll, line 23: For consistency with the [Rails convention](https://guides.rubyonrails.org/action_mailer_basics.html#mailers-are-similar-to-controllers), can we invoke this as `NotificationMailer.with(user: Current.user).test_email` and then obtain the email later with params[:user].email?

db/schema.rb
```diff
@@ -86,7 +86,7 @@
     t.uuid "accountable_id"
     t.decimal "balance", precision: 19, scale: 4, default: "0.0"
     t.string "currency", default: "USD"
-    t.virtual "classification", type: :string, as: "\nCASE\n    WHEN ((accountable_type)::text = ANY (ARRAY[('Account::Loan'::character varying)::text, ('Account::Credit'::character varying)::text, ('Account::OtherLiability'::character varying)::text])) THEN 'liability'::text\n    ELSE 'asset'::text\nEND", stored: true
+    t.virtual "classification", type: :string, as: "\nCASE\n    WHEN ((accountable_type)::text = ANY ((ARRAY['Account::Loan'::character varying, 'Account::Credit'::character varying, 'Account::OtherLiability'::character varying])::text[])) THEN 'liability'::text\n    ELSE 'asset'::text\nEND", stored: true
     t.boolean "is_active", default: true, null: false
```
zachgoll, line 90: Which Postgres version are you running locally?  Think the schema dump may have just interpreted this differently

zachgoll (commented):
Nice work here.  I left some specific comments within the code.

In addition to those comments, I think we need to add some unit and functional tests.  Off the top of my head, I think the important things we need to verify are:

- The endpoint sends an email if all SMTP settings are populated
- The test email endpoint fails if user is missing SMTP settings
- The mailer is passed the correct SMTP settings 
- Depending on if app is self-hosted or not, SMTP settings come from different places (see my other comment inline)

db/schema.rb
ThibautGrx: v 13.13. I could not find why it was interpreted differently.

ThibautGrx:
Thanks @zachgoll for the review. Here is the updated version. I hope it meets your expectations. 
Sorry in advance if you find some misspellings

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,23 @@
...
+  private
+    def smpt_settings
+      {
+        address:   Setting.smtp_host,
+        port:      Setting.smtp_port,
+        user_name: Setting.smtp_username,
+        password:  Setting.smtp_password,
+        tls:       ENV["TLS"] == "true"
+        }
+    end
```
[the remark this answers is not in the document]
ThibautGrx: I had to use after_action to override delivery method settings  : https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-callbacks

db/schema.rb
zachgoll: Gotcha.  I could be wrong, but I _think_ if you update to PG 14+ you won't run into this

test/mailers/application_mailer_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    assert_equal ENV["EMAIL_SENDER"], email.from.first
+  end
+
+  test "should return setting config when self hosting is enabled" do
+    ENV["SELF_HOSTING_ENABLED"] = "true"
+    email = ApplicationMailer.test_email
+    assert_equal @smtp_settings_from_settings, email.delivery_method.settings.slice(:address, :port, :user_name, :password, :tls)
+    assert_equal Setting.email_sender, email.from.first
+  end
+end
```
zachgoll, line 61: This is a tough one.  I think you're testing the _right_ things here, but I'm a little worried that we're diving too deep into some of the Rails internals here.  What about something like this?

```rb
require "test_helper"

class ApplicationMailerTest < ActionMailer::TestCase
  setup do
    class TestMailer < ApplicationMailer
      def test_email
        mail(to: "zach.gollwitzer@gmail.com", from: "test@email.com", subject: "Test email subject", body: "Test email body")
      end
    end
  end

  test "should use self host settings when self host enabled" do
    ENV["SELF_HOSTING_ENABLED"] = "true"

    Setting.stubs(:smtp_host).returns("self-hosted-smtp.example.com")

    TestMailer.test_email.deliver_now

    assert_emails 1
    assert_equal "self-hosted-smtp.example.com", ActionMailer::Base.deliveries.first.delivery_method.settings[:address]
  end

  test "should use regular env settings when self host disabled" do
    ENV["SELF_HOSTING_ENABLED"] = "false"

    TestMailer.test_email.deliver_now

    assert_emails 1
    assert_nil ActionMailer::Base.deliveries.first.delivery_method.settings[:address]
  end
end

```

zachgoll (commented):
Looking good here!  Just a few more little suggestions and then I think we're good to merge this through.

db/schema.rb
ThibautGrx: Thanks. I'll try to update pg.

test/mailers/application_mailer_test.rb
ThibautGrx: It's seems to me much better and simpler... I had to change mailer after action condition, and stop checking that delivery method == :smtp.

zachgoll (approved):
Looks good, thanks for the revisions on this!

ppfeister:
Apologies for the dead-thread revival ---

Did y'all test this dockerized as well or just in a dev env/one of the other self hosted envs?
Having some issues with SMTP and would like to know if it's localized or not.

zachgoll:
@ppfeister we're actually going to be removing this settings option (temporarily) and re-introducing it at a later date.  I will have it removed today.

This PR originally went in because we had anticipated a feature to be implemented that uses it, but we ended up not implementing that feature.

villancikos:
Let me know if I can at least send the reset password email 👀 through a shell command or so!

PR #683, opened 2024-04-26
add basic support for deleting account

### LINKED ISSUE #675: Feature: Allow user to delete their Maybe account and all data
CleanShot 2024-04-25 at 11 20 40

## Feature Overview

In an effort to give users full control over their data, this issue allows them to delete their entire Maybe account and associated data.

## Requirements

- [ ] User can delete their user account
  - [ ] If user is `admin` and there are no other users in the family, delete the user, family, and all associated data
  - [ ] If user is `member`, only delete the user and the associated _user_ data 
  - [ ] In all other scenarios, trigger a failure
    - For example: An `admin` tries to delete their account while the family has other members.  In a scenario like this, we should error and say something like, "Admin cannot delete account while other users are present.  Please delete all members first."
- [ ] User must confirm deletion via a dialog
- [ ] After deletion, user is redirected to the sign up page with success notification
- [ ] Deletions should be atomic.  A delete action should _never_ result in invalid states or partially deleted data.
- [ ] Appropriate tests written
- [ ] i18n translations provided for `en.yml`

## Implementation Suggestions

- Be sure to familiarize yourself with the `Family` and `User` models.  Currently, the app stores quite a bit of data on `Family`, hence some of the conditional deletion requirements listed above.
- Use your best judgment on this issue.  There are likely edge cases not covered here.  The overall goal at this stage is to get a _basic_ and _simple_ version of this working that can be improved upon in the future
- We _may_ want to put the deletion process in a background job, but only if it has a clear benefit

## Designs

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=3176-910&mode=design&t=4eaJRTzT6AYU5CNv-0

## Reminders

## Bounty
There is a $150 bounty on this. To participate in the bounty:

1. Comment `/attempt #675` to indicate you plan on working on this.
2. Submit a draft PR once you start and include `/claim #675` along with your general game plan for how you intent to implement this.

jemiluv8:
fixes #675 
/claim #675

### DEMO 1 - admin is not allowed to delete account until there are no members
https://github.com/maybe-finance/maybe/assets/119384208/3cf8f17c-460e-49b1-a0f8-030ecc297d28

### DEMO 2 - admin account deletion after all members are deleted

https://github.com/maybe-finance/maybe/assets/119384208/86202816-9012-472e-bd7e-941cd63244b5

This draft PR demonstrates a naive implementation of user account deletion that I'll build on over the next few hours.
This implementation deletes the user and attempts to delete other related data in the same request "cycle".

The problem identified while deleting the demo account is that the delete operation can really take a long time depending on the number of accounts associated with the user's family. I have thus decided to do the cleanup in a background job.

**UI Requirements**
- [X]  User must confirm deletion via a dialog
- [X]  After deletion, user is redirected to the sign up page with success notification
- [X]  Deletions should be atomic. A delete action should never result in invalid states or partially deleted data.
- [X]  i18n translations provided for en.yml

BACKEND Requirements
- [X]  Appropriate tests written
- [X] Introduce background worker to do cleanup according to logic relating to Family, Accounts, admin and member User types
- [X] Write tests for endpoint and background worker
- [ ] Final review and testing

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -7,7 +7,7 @@ class SessionsController < ApplicationController
   end
 
   def create
-    if user = User.authenticate_by(email: params[:email], password: params[:password])
+    if user = User.authenticate_by(email: params[:email], password: params[:password], marked_for_deletion: false)
       login user
       redirect_to root_path
     else
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -19,6 +19,7 @@ class Settings::ProfilesController < ApplicationController
   
   def destroy
     begin
+      Rails.logger.info "DESTROYING"
       delete_user
       logout
       redirect_to new_registration_path, notice: t(".success")
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -19,7 +19,6 @@ class Settings::ProfilesController < ApplicationController
   
   def destroy
     begin
-      Rails.logger.info "DESTROYING"
       delete_user
       logout
       redirect_to new_registration_path, notice: t(".success")
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -16,11 +16,42 @@ class Settings::ProfilesController < ApplicationController
       render :edit, status: :unprocessable_entity
     end
   end
+  
+  def destroy
+    begin
+      delete_user
+      logout
+      redirect_to new_session_path, notice: t(".success")
+    rescue ActiveRecord::RecordNotDestroyed => e
+      Rails.logger.error "Error deleting account: #{e.message}"
+      return redirect_to settings_profile_path, alert: t(".account_deletion_failed")
+    rescue StandardError => e
+      Rails.logger.error "An unexpected error occurred: #{e.message}"
+      redirect_to settings_profile_path, alert: t(".account_deletion_failed")
+    end
+  end
 
   private
 
   def user_params
     params.require(:user).permit(:first_name, :last_name,
                                  family_attributes: [ :name, :id ])
+  end
+
+  def delete_user
+    # raise StandardError.new "Unexpected error"
+    if Current.user.isMember
+      Rails.logger.info "Is member"
+      other_family_users = User.where(family_id: Current.user.family_id).where.not(id: Current.user.id).count
+      ActiveRecord::Base.transaction do
+        if other_family_users == 0
+          # this is true for our demo user but should not normally happen
+          Current.family.destroy # this takes care of deleting all related accounts
+        end
+        Current.user.destroy
+      end
+    ## TODO: work on other edge cases involving admin and others
+    ## TODO: handle errors properly if the transaction fails
+    end
   end
 end
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -19,12 +19,16 @@ class Settings::ProfilesController < ApplicationController
   
   def destroy
     begin
-      delete_user
-      logout
+      if Current.user.admin?
+        other_admins_count = Current.family.admins.where.not(id: Current.user.id).count
+        members_count = Current.family.members.count
+
+        if other_admins_count === 0 && members_count > 0
+          return redirect_to settings_profile_path, alert: t(".cannot_delete_admin_account_warning")
+        end
+      end
+      mark_user_for_deletion_and_logout
       redirect_to new_registration_path, notice: t(".success")
-    rescue ActiveRecord::RecordNotDestroyed => e
-      Rails.logger.error "Error deleting account: #{e.message}"
-      return redirect_to settings_profile_path, alert: t(".account_deletion_failed")
     rescue StandardError => e
       Rails.logger.error "An unexpected error occurred: #{e.message}"
       redirect_to settings_profile_path, alert: t(".account_deletion_failed")
@@ -38,10 +42,9 @@ class Settings::ProfilesController < ApplicationController
                                  family_attributes: [ :name, :id ])
   end
 
-  def delete_user
-    attributes = Current.user.attributes
-    Current.user.destroy!
-
-    DeleteUserJob.perform_later(attributes)
+  def mark_user_for_deletion_and_logout
+    Current.user.update!(marked_for_deletion: true)
+    DeleteUserJob.perform_later(Current.user)
+    logout
   end
 end
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -16,7 +16,7 @@ class Settings::ProfilesController < ApplicationController
       render :edit, status: :unprocessable_entity
     end
   end
-  
+
   def destroy
     begin
       if Current.user.admin?
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -21,7 +21,7 @@ class Settings::ProfilesController < ApplicationController
     begin
       delete_user
       logout
-      redirect_to new_session_path, notice: t(".success")
+      redirect_to new_registration_path, notice: t(".success")
     rescue ActiveRecord::RecordNotDestroyed => e
       Rails.logger.error "Error deleting account: #{e.message}"
       return redirect_to settings_profile_path, alert: t(".account_deletion_failed")
@@ -39,19 +39,9 @@ class Settings::ProfilesController < ApplicationController
   end
 
   def delete_user
-    # raise StandardError.new "Unexpected error"
-    if Current.user.isMember
-      Rails.logger.info "Is member"
-      other_family_users = User.where(family_id: Current.user.family_id).where.not(id: Current.user.id).count
-      ActiveRecord::Base.transaction do
-        if other_family_users == 0
-          # this is true for our demo user but should not normally happen
-          Current.family.destroy # this takes care of deleting all related accounts
-        end
-        Current.user.destroy
-      end
-    ## TODO: work on other edge cases involving admin and others
-    ## TODO: handle errors properly if the transaction fails
-    end
+    attributes = Current.user.attributes
+    Current.user.destroy!
+
+    DeleteUserJob.perform_later(attributes)
   end
 end
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -44,7 +44,7 @@ class Settings::ProfilesController < ApplicationController
 
   def mark_user_for_deletion_and_logout
     Current.user.update!(marked_for_deletion: true)
-    DeleteUserJob.perform_later(Current.user)
+    Current.user.purge_data_later
     logout
   end
 end
--- a/app/javascript/controllers/application.js
+++ b/app/javascript/controllers/application.js
@@ -29,9 +29,13 @@ Turbo.setConfirmMethod((message) => {
     }
 
     if (reject) {
-      document.getElementById("turbo-confirm-reject").innerHTML = reject;
-    } else {
-      document.getElementById("turbo-confirm-reject").remove()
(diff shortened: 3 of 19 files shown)


app/controllers/settings/profiles_controller.rb
```diff
@@ -16,11 +16,32 @@ def update
       render :edit, status: :unprocessable_entity
     end
   end
+  
+  def destroy
```
jemiluv8, line 20: the primary controller that handles deletion of the user account
It simply deletes the user account (which is the most basic requirement)
and schedules a background job to do required cleanup

app/javascript/controllers/application.js
```diff
@@ -20,9 +20,24 @@ Turbo.setConfirmMethod((message) => {
...
 
+    if (acceptClass) {
+      document.getElementById("turbo-confirm-accept").className += acceptClass
+    }
+
     if (accept) {
       document.getElementById("turbo-confirm-accept").innerHTML = accept;
     }
+
+    if (reject) {
```
jemiluv8, line 31: Hack really to allow us too add a reject button down there

app/javascript/controllers/application.js
```diff
@@ -20,9 +20,24 @@ Turbo.setConfirmMethod((message) => {
       document.getElementById("turbo-confirm-body").innerHTML = body;
     }
 
+    if (acceptClass) {
```
jemiluv8, line 23: Allows us to customize the accept button

app/jobs/delete_user_job.rb
```diff
@@ -0,0 +1,37 @@
...
+        end
+      end
+    elsif user.role == "admin"
+      other_admins = User.where(family_id: user.family_id, role: "admin").where.not(id: user.id).count
+
+      if other_admins == 0
+        ActiveRecord::Base.transaction do
+          
+          if family != nil
+            family.destroy
```
jemiluv8, line 26: this deletes accounts and related data

app/jobs/delete_user_job.rb
```diff
@@ -0,0 +1,37 @@
...
+          end
+          # since other users are now invalid because of the orphaned family -> delete users as well
+          users = User.where(family_id: user.family_id)
+          users.destroy_all
```
jemiluv8, line 30: This is my addition thus far.
I figured since family_id is required on all user models,
any existing user with a non-existent family is "invalid"?

test/fixtures/users.yml
```diff
@@ -4,6 +4,7 @@ family_admin:
   last_name: Dylan
   email: bob@bobdylan.com
   password_digest: <%= BCrypt::Password.create('password') %>
+  role: admin
```
jemiluv8, line 7: I just wonder why whoever created this test fixture never bothered to specify the admin role?

test/jobs/delete_user_job_test.rb
```diff
@@ -0,0 +1,41 @@
...
+      DeleteUserJob.perform_later(@user_admin.attributes)
+    end
+  end
+
+  test "deleting a member with an admin will not delete the family" do
+    DeleteUserJob.perform_now(@user_member.attributes)
+    Family.find(@user_member.family_id)
+  end
+
+  test "family, related accounts and other members are deleted for the last remaining admin" do
```
jemiluv8, line 27: We could have more edge cases to test but these are it for now.

app/javascript/controllers/application.js
```diff
@@ -20,9 +20,24 @@ Turbo.setConfirmMethod((message) => {
...
     }
+
+    if (reject) {
+      const button = document.createElement("button")
+      button.setAttribute("id", "turbo-confirm-reject")
+      button.setAttribute("class", "w-full text-red-600 rounded-xl text-center p-[10px] border mt-2")
+      button.setAttribute("value", "reject")
+      button.innerHTML = reject
+
+      document.getElementById("turbo-confirm-dialog-form").appendChild(button)
```
zachgoll, line 38: I think we could probably move this to the `_confirm_modal` partial and then toggle classes as needed here.

app/models/user.rb
```diff
@@ -13,6 +13,14 @@ class User < ApplicationRecord
     password_salt&.last(10)
   end
 
+  def isAdmin
+    role.downcase === "admin"
+  end
+
+  def isMember
+    role.downcase === "member"
+  end
```
zachgoll, line 22: These are not necessary since we've got an enum.  You should be able to already call `User.first.member?` and `User.first.admin?`.

app/controllers/settings/profiles_controller.rb
```diff
@@ -16,11 +16,32 @@ def update
...
   def user_params
     params.require(:user).permit(:first_name, :last_name,
                                  family_attributes: [ :name, :id ])
   end
+
+  def delete_user
+    attributes = Current.user.attributes
+    Current.user.destroy!
+
+    DeleteUserJob.perform_later(attributes)
```
zachgoll, line 45: While I think you're right that we should be doing this in a background job, I think a potentially more stable strategy would be something like this:

```rb
def mark_user_for_deletion_and_logout
  Current.user.update!(marked_for_deletion: true)
  DeleteUserJob.perform_later(Current.user)
  logout
end
```

And then on login, we could check `marked_for_deletion` and fail early if it is marked `true`.

This would allow us to work directly with the `User` record all in one spot and avoid some of the error handling here.

app/jobs/delete_user_job.rb
```diff
@@ -0,0 +1,35 @@
...
```
zachgoll, line 30: While I think my comment above (passing in the user instance) will simplify all of this logic a bit, we should move a lot of this method into the models themselves.  As an example:

```rb
other_admins = User.where(family_id: user.family_id, role: "admin").where.not(id: user.id).count
```

This should likely be a scope or method on `Family` so that we can call:

```rb
other_admins_count = Current.family.admins.where.not(id: user.id).count
```

zachgoll (commented):
I like the test cases here and overall strategy.  Just left a few suggestions on how we could simplify some of the logic.

app/controllers/settings/profiles_controller.rb
zachgoll: FYI - https://github.com/maybe-finance/maybe/issues/675#issuecomment-2081051730

app/jobs/delete_user_job.rb
zachgoll: FYI - https://github.com/maybe-finance/maybe/issues/675#issuecomment-2081051730

app/models/user.rb
jemiluv8: Gotcha

app/javascript/controllers/application.js
jemiluv8: I just didn't want to modify the _confirm_modal partial. I'll do that now instead. So we just toggle visibility based on args.

app/jobs/delete_user_job.rb
jemiluv8: Gotcha.

jemiluv8:
@zachgoll , I'll work on the changes and revert.

app/controllers/sessions_controller.rb
```diff
@@ -7,7 +7,7 @@ def new
   end
 
   def create
-    if user = User.authenticate_by(email: params[:email], password: params[:password])
+    if user = User.authenticate_by(email: params[:email], password: params[:password], marked_for_deletion: false)
```
jemiluv8, line 10: prevents users marked_for_deletion from logging in.

app/controllers/settings/profiles_controller.rb
```diff
@@ -16,11 +16,35 @@ def update
       render :edit, status: :unprocessable_entity
     end
   end
+  
+  def destroy
+    begin
+      if Current.user.admin?
```
jemiluv8, line 22: This is much simpler I hope.
We just refuse to delete last admin with members - otherwise all goes through

app/controllers/settings/profiles_controller.rb
```diff
@@ -16,11 +16,35 @@ def update
...
+  end
 
   private
 
   def user_params
     params.require(:user).permit(:first_name, :last_name,
                                  family_attributes: [ :name, :id ])
   end
+
+  def mark_user_for_deletion_and_logout
```
jemiluv8, line 45: It read so well. I couldn't help using it. Thanks for the snippet.

app/jobs/delete_user_job.rb
```diff
@@ -0,0 +1,17 @@
+class DeleteUserJob < ApplicationJob
+  queue_as :default
+
+  def perform(user)
```
jemiluv8, line 4: I got rid of ActiveRecord transactions here because destroys are always put within a transaction - from the docs.
And we only ever destroy here.

So far, regular users don't have anything besides access to family accounts so other cleanup might be a future thing.

Admin deletion is currently in two
- the last admin and we just get rid of the family which deletes the admin user
- not the last admin and we just delete the user only - and related data.

jemiluv8:
@zachgoll , I've made changes based on review. Let me know how the click-testing goes.

app/jobs/delete_user_job.rb
```diff
@@ -0,0 +1,24 @@
...
+        else
+          user.destroy
+        end
+      end
+    end
+  rescue => e
+    Rails.logger.error "An error occurred during DeleteUserJob: #{e.message} for the user #{user.id}"
+    raise e # just propagating
+  end
+end
```
zachgoll, line 24: The logic here looks better, but I think we should be handling all of this within `user.rb` and `family.rb`.  Something along the lines of:

```rb
# delete_user_job.rb

def perform(user)
  user.purge_data
end
```

```rb
# user.rb

def purge_data_later
  DeleteUserJob.perform_later(self)
end

def purge_data
    if family.can_be_destroyed_by?(self)
      family.destroy! # with correct cascade behavior, this should destroy the user too
    else
      destroy!
    end
end
```

```rb
# family.rb

def can_be_destroyed_by?(user)
  users.count == 1 && user.admin?
end
```

```rb
# profiles_controller.rb
def mark_user_for_deletion_and_logout
    Current.user.update!(marked_for_deletion: true)
    Current.user.purge_data_later
    logout
end
```

app/views/shared/_confirm_modal.html.erb
```diff
@@ -1,5 +1,5 @@
 <dialog id="turbo-confirm" class="max-w-[420px] w-full rounded-xl">
-  <form method="dialog" class="p-4">
+  <form method="dialog" class="p-4" id="turbo-confirm-dialog-form">
```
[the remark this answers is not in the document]
jemiluv8: It is not being used. Will clean it up.

app/models/family.rb
```diff
@@ -4,6 +4,14 @@ class Family < ApplicationRecord
   has_many :transactions, through: :accounts
   has_many :transaction_categories, dependent: :destroy, class_name: "Transaction::Category"
 
+  def admins
+    users.where(users: { role: "admin", marked_for_deletion: false })
+  end
+
+  def members
+    users.where(users: { role: "member", marked_for_deletion: false })
+  end
```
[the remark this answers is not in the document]
jemiluv8: updated the method to not user the marked_for_deletion query.
Had challenges using scope here though so maintained the method.

zachgoll:
@jemiluv8 thanks for the updates here.  I think this PR is at a final state now, although in #698 there is a "deactivate" strategy implemented which I think offers a more concise and intuitive way to handle what we had come up here with the `marked_for_deletion` flag on the `User` model.

Since there's a lot of good work here that allowed us to see the requirements for this issue more clearly and come to a solid solution, we'll award a split bounty on this.

I'll close this PR out, but there will be a payment awarded after #698 is merged.  Thanks for all the work on this!

jemiluv8:
That's great @zachgoll, I learned a lot and been intrigued by the deactivation strategy.
It looked solid.

zachgoll:
@jemiluv8 just an FYI, the bounty split didn't work properly when the PR was merged, but we're working to get that resolved right now so you get the appropriate payout.

PR #684, opened 2024-04-26
Fix account page on fresh install

rmarescu:
### What

- [x] Fix `/settings/profile` on a fresh install
- [x] Consolidate logic around displaying name on the UI
- [x] Consolidate logic around user initial on the UI

### Why

I did a fresh install on Render (it worked great!) and I got an error when I tried accessing the Account page.

CleanShot 2024-04-26 at 14 26 31@2x

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,10 +7,20 @@ class User < ApplicationRecord
   validates :email, presence: true, uniqueness: true
   normalizes :email, with: ->(email) { email.strip.downcase }
 
+  normalizes :first_name, :last_name, with: ->(value) { value.strip.presence }
+
   enum :role, { member: "member", admin: "admin" }, validate: true
 
   generates_token_for :password_reset, expires_in: 15.minutes do
     password_salt&.last(10)
+  end
+
+  def display_name
+    [ first_name, last_name ].compact.join(" ").presence || email
+  end
+
+  def initial
+    (display_name&.first || email.first).upcase
   end
 
   def acknowledge_upgrade_prompt(commit_sha)
--- a/app/views/layouts/_sidebar.html.erb
+++ b/app/views/layouts/_sidebar.html.erb
@@ -4,14 +4,16 @@
   <% end %>
   <div id="user-menu" data-controller="menu">
     <button data-menu-target="button">
-      <div class="text-white w-9 h-9 bg-gray-400 rounded-full flex items-center justify-center text-lg uppercase"><%= Current.user.email.first %></div>
+      <div class="text-white w-9 h-9 bg-gray-400 rounded-full flex items-center justify-center text-lg uppercase"><%= Current.user.initial %></div>
     </button>
     <div data-menu-target="content" class="hidden absolute w-[240px] z-10 top-10 left-[255px] top-[72px] bg-white rounded-sm shadow-xs border border-alpha-black-25">
       <div class="p-3 flex items-center gap-3">
-        <div class="text-white shrink-0 w-9 h-9 bg-gray-400 rounded-full flex items-center justify-center text-lg uppercase"><%= Current.user.email.first %></div>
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -7,12 +7,22 @@ class User < ApplicationRecord
   validates :email, presence: true, uniqueness: true
   normalizes :email, with: ->(email) { email.strip.downcase }
 
+  normalizes :first_name, :last_name, with: ->(value) { value.strip.presence }
```
rmarescu, line 10: By default, the columns have a `nil` value. When the form is saved on the UI (without values), then empty strings are saved in the database.
Generally, I think it's best to normalize `String` values to have sanitized data (i.e. `nil`s instead of empty strings).

app/models/user.rb
```diff
@@ -7,12 +7,22 @@ class User < ApplicationRecord
...
+
   enum :role, { member: "member", admin: "admin" }, validate: true
 
   generates_token_for :password_reset, expires_in: 15.minutes do
     password_salt&.last(10)
   end
 
+  def display_name
+    [ first_name, last_name ].compact.join(" ").presence || email
```
rmarescu, line 19: This falls back to email because is also used on the list of members, and it looks odd without some sort of identifier. I've added a screenshot below for reference.

app/views/layouts/_sidebar.html.erb
```diff
@@ -4,14 +4,16 @@
...
     </button>
     <div data-menu-target="content" class="hidden absolute w-[240px] z-10 top-10 left-[255px] top-[72px] bg-white rounded-sm shadow-xs border border-alpha-black-25">
       <div class="p-3 flex items-center gap-3">
-        <div class="text-white shrink-0 w-9 h-9 bg-gray-400 rounded-full flex items-center justify-center text-lg uppercase"><%= Current.user.email.first %></div>
+        <div class="text-white shrink-0 w-9 h-9 bg-gray-400 rounded-full flex items-center justify-center text-lg uppercase"><%= Current.user.initial %></div>
         <div>
-          <span class="text-gray-900 font-medium text-sm"><%= Current.user.first_name %> <%= Current.user.last_name %></span>
-          <span class="text-gray-500 text-sm"><%= Current.user.email %></span>
+          <span class="text-gray-900 font-medium text-sm"><%= Current.user.display_name %></span>
+          <% if Current.user.display_name != Current.user.email %>
```
rmarescu, line 14: Until at least one name is provided, there is no point rendering the email twice.

app/views/settings/profiles/show.html.erb
```diff
@@ -39,9 +39,9 @@
           </div>
           <div class="flex gap-2 items-center bg-white p-4 border border-alpha-black-25 rounded-lg">
             <div class="mr-1 flex justify-center items-center bg-gray-50 w-8 h-8 rounded-full border border-alpha-black-25">
-              <p class="uppercase text-xs text-gray-500"><%= Current.user.first_name.first %></p>
```
rmarescu, line 42: This was causing the page crash, when `first_name` was `nil` on fresh install.

app/views/settings/profiles/show.html.erb
```diff
@@ -39,9 +39,9 @@
...
+              <p class="uppercase text-xs text-gray-500"><%= Current.user.initial %></p>
             </div>
-            <p class="text-gray-900 font-medium text-sm"><%= Current.user.first_name %> <%= Current.user.last_name %></p>
```
rmarescu, line 44: Before was like this - 

CleanShot 2024-04-26 at 14 48 01@2x

rmarescu (commented):
Here is an attempt to resolve an issue I encountered after doing a fresh installation.

zachgoll (approved):
This is great, thanks for the fixes and comments on this!

Issue #674, opened 2024-04-25
Feature: Configure ActiveStorage and implement profile picture uploads


zachgoll:
![CleanShot 2024-04-25 at 10 55 04](https://github.com/maybe-finance/maybe/assets/16676157/332cb5c7-0ef3-4cfa-823f-3223fb588341)

## Feature Overview

We currently have a _placeholder_ in the profile settings for profile picture uploads.  This issue will enable a user to upload their own profile picture to Maybe and will enable ActiveStorage.

## Requirements

_If there is a missing / incorrect requirement, please leave a comment before starting work on this._

### ActiveStorage Setup

Since the Maybe app will be self-hosted, we should allow a user to choose their preferred object storage and configure it 100% with environment variables (credentials helper may still be used, but we cannot _rely_ on it as some self-hosted setups like Render need to be configured 100% by environment variables).

- [ ] Default storage will be Disk storage (to allow self-hosters to remain "zero config")
- [ ] Other storage options will be AWS, GCP, and Azure and can be specified via `ACTIVE_STORAGE_SERVICE=`
- [ ] Platform specific storage should have a prefix `SERVICE_` for each required config.  For example, `S3_ACCESS_KEY_ID` for AWS S3 or `GCP_ACCESS_KEY_ID` for Google Cloud
- [ ] Can be configured 100% through environment variables

### Profile picture uploads

- [ ] User can upload a new profile picture
  - Suggested constraints (use your best judgment): 5MB max, JPG/PNG/GIF format?
- [ ] User can remove an existing profile picture
- [ ] Tests where appropriate

## Designs

Below are the designs you should follow while implementing this:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=3176-87&mode=design&t=4eaJRTzT6AYU5CNv-0

## Reminders

- Make sure to review our [contributing guidelines](https://github.com/maybe-finance/maybe/blob/main/CONTRIBUTING.md) before starting on an issue
- We do our best to define a clear spec for new features and fixes, but think of them as "suggestions", not "hard requirements".  We welcome ideas and suggestions!
  - If you see missing requirements to this issue, please leave a comment below explaining what is missing and why it is important.
  - If you see a requirement that you think is _incorrect_ or _not optimal_, please leave a comment explaining what you think needs to change below.

## Bounty
There is a $200 bounty on this. To participate in the bounty:

[...]

PR #685, opened 2024-04-27
Activestorage profile picture uploads

claudey:
Attempt for #674 

- ActiveStorage had no version for 7.0.2.alpha so I had to update to the latest rails
- Parser gem would not work with 3.3.0 so .ruby-version was updated to 3.3.1

- Successfully installed activestorage after those changes
- Attached an avatar field to User model
- Currently saving successfully
- Shows in preview after saving.

WIP:
- Not completed purging user avatar

/claim #674

https://github.com/maybe-finance/maybe/assets/1866816/5c372cb0-3e56-4dc5-9800-8cb22f88b932


zachgoll:
@claudey thanks for the work on this!  I'm going to close it out now as we've merged a completed solution in #687.

claudey:
Thanks for the reply, Zach. I'll keep trying at future issues.

PR #686, opened 2024-04-27
Add Transaction Merchant management

### LINKED ISSUE #681: Feature: Add Transaction Merchant management
CleanShot 2024-04-25 at 16 42 27

## Feature Overview

The goal of this feature is to introduce a "Merchant" model that a user can manage.  A user will be able to set a merchant for a transaction and will be able to manage all of their family-owned merchants in settings.

## Requirements

- [ ] Should be implemented at the route `/transactions/merchants` (in settings)
- [ ] User can create a new merchant (choose a color and name)
- [ ] User can edit existing merchant
- [ ] User can delete merchant
- [ ] When user deletes merchant, app asks them to confirm before deleting
- [ ] i18n translations provided
- [ ] Tests written where appropriate

## Designs

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=3194-441&mode=design&t=gI2CS4JU0eghOMVN-0

## Reminders

## Bounty
There is a $100 bounty on this. To participate in the bounty:

1. Comment `/attempt #681` to indicate you plan on working on this.
2. Submit a draft PR once you start and include `/claim #681` along with your general game plan for how you intent to implement this.

jakubkottnauer:
Closes #681

In addition to the issue requirements, I have also hooked up the merchants to transaction filtering.

Video:
Kapture 2024-04-28 at 12 54 23

/claim #681


jakubkottnauer:
@zachgoll ready for review 😄

app/javascript/controllers/merchant_avatar_controller.js
```diff
@@ -0,0 +1,32 @@
...
+    this.avatarTarget.textContent = (e.currentTarget.value?.[0] || "?").toUpperCase();
+  }
+
+  handleColorChange = (e) => {
+    const color = e.currentTarget.value;
+    this.avatarTarget.style.backgroundColor = `color-mix(in srgb, ${color} 5%, white)`;
+    this.avatarTarget.style.borderColor = `color-mix(in srgb, ${color} 10%, white)`;
+    this.avatarTarget.style.color = color;
+  }
+}
```
zachgoll, line 32: Nice idea for this 👍

app/models/transaction.rb
```diff
@@ -3,6 +3,7 @@ class Transaction < ApplicationRecord
 
   belongs_to :account
   belongs_to :category, optional: true
+  belongs_to :merchant, optional: true
```
zachgoll, line 6: We should probably update these lines to:

```
belongs_to :category, optional: true, dependent: :nullify
belongs_to :merchant, optional: true, dependent: :nullify
```

Currently, there is a database constraint for categories that nullifies the `category` field on all transactions when it is deleted introduced by [this migration](https://github.com/maybe-finance/maybe/blob/main/db/migrate/20240404112829_change_transaction_category_delete_behavior.rb).

I'm still not 100% sure whether we should be doing this at the model level, DB level, or _both_.  In the future, we may have other APIs modifying data (companion apps, etc.) which would lean towards handling it at the DB level.  But for now, I think we can just do it at the model level here.

zachgoll (approved):
Awesome work on this.  Thanks for adding this to the filters as well!  One small comment about the delete behavior, but otherwise, looks like this is good to go!

app/models/transaction.rb
jakubkottnauer: Doesn't the `has_many :transactions, dependent: :nullify` on the `Transaction::Category`/ `Transaction::Merchant` models already take care of this?

zachgoll: Yep, you're 100% correct.  I was thinking about that backwards!

PR #687, opened 2024-04-27
Feature/profile image uploads

crobbo:
/claim #674

[Screencast from 2024-04-27 10-13-24.webm](https://github.com/maybe-finance/maybe/assets/47796704/7e2fc296-ddf8-48af-a2f3-437b410e5fab)


config/storage.yml
```diff
@@ -1,34 +1,27 @@
...
 
-# mirror:
-#   service: Mirror
-#   primary: local
-#   mirrors: [ amazon, google, microsoft ]
+microsoft:
+  service: AzureStorage
+  storage_account_name: ENV["AZURE_STORAGE_ACCOUNT_NAME"] %>
+  storage_access_key: ENV["AZURE_STORAGE_ACCESS_KEY"] %>
+  container: ENV["AZURE_STORAGE_CONTAINER"] %>
```
zachgoll, line 27: ERB syntax error here

config/storage.yml
```diff
@@ -1,34 +1,27 @@
...
-# microsoft:
-#   service: AzureStorage
-#   storage_account_name: your_account_name
-#   storage_access_key: <%= Rails.application.credentials.dig(:azure_storage, :storage_access_key) %>
-#   container: your_container_name-<%= Rails.env %>
+# Replace credentials with your own GCS keyfile
+google:
+  service: GCS
+  project: your_project
+  credentials: <%= Rails.root.join("path/to/gcs.keyfile") %>
```
zachgoll, line 20: We should probably set a more descriptive keyfile name such as `gcp-storage-keyfile.json` and then add that to the .gitignore here

Gemfile
```diff
@@ -36,6 +36,12 @@ gem "sentry-rails"
 gem "rails-settings-cached"
 gem "octokit"
 
+# Active Storage
+gem 'aws-sdk-s3', '~> 1'
+gem 'azure-storage-blob', '~> 1.0', '>= 1.0.1'
```
zachgoll, line 41: Is there a specific reason for using a different versions from the Rails guides for AWS and Azure?

```
gem 'aws-sdk-s3', require: false
gem 'azure-storage-blob', "~> 2.0", require: false
```

zachgoll (commented):
The upload and preview flow looks good.  Left some more specific comments on improvements we could make.

In addition to those comments, can we also add a section in `.env.example` that explains how to configure this and what env vars are available for use?  (similar to what we have for other self-host configs)

Gemfile
crobbo: Nope. I missed that in the guide.

app/views/settings/profiles/show.html.erb
```diff
@@ -5,26 +5,40 @@
...
+              <% else %>
+                <%= lucide_icon "image-plus", class: "w-6 h-6 text-gray-500" %>
+              <% end %>
+            </div>
+            <%= lucide_icon "image-plus", class: "hidden w-6 h-6 text-gray-500", data: { profile_image_preview_target: "template" } %>
+            <div data-profile-image-preview-target="clearBtn" 
+                data-action="click->profile-image-preview#clear" 
+                class="<%= profile_image_attached ? '' : 'hidden' %> cursor-pointer absolute bottom-0 right-0 w-8 h-8 bg-gray-50 rounded-full flex justify-center items-center border border-white border-2">
+              <%= lucide_icon "x", class: "w-4 h-4 text-gray-500" %>
+            </div>
```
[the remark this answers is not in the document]
crobbo: Amended this so it deletes the image. 

It always irritates me when an app lets you upload an image but not remove it, so you have to upload a new image to replace it.

app/models/user.rb
```diff
@@ -9,6 +9,9 @@ class User < ApplicationRecord
 
   enum :role, { member: "member", admin: "admin" }, validate: true
 
+  has_one_attached :profile_image
```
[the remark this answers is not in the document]
crobbo: That seems a sensible thing to include.

crobbo:
@zachgoll thanks for the review, it was all valid feedback. I've made those amends and requested a re-review 👍🏻

.env.example
```diff
@@ -56,3 +56,33 @@ UPGRADES_TARGET=release # `release` or `commit`
...
+# ACTIVE_STORAGE_SERVICE=google
+# GCS_PROJECT=
+# GCS_BUCKET=
+
+# Microsoft Azure Storage
+# ========================
+# ACTIVE_STORAGE_SERVICE=azure
+# AZURE_STORAGE_ACCOUNT_NAME=
+# AZURE_STORAGE_ACCESS_KEY=
+# AZURE_STORAGE_CONTAINER=
```
zachgoll, line 88: Perfect, looks good 👍

app/controllers/settings/profiles_controller.rb
```diff
@@ -5,6 +5,10 @@ def show
   def update
     user_params_with_family = user_params
 
+    if params[:user][:profile_image].blank?
+      Current.user.profile_image.purge
+    end
```
zachgoll, line 10: Since this form is tied to the update for a user's first and last name, the profile picture will be cleared out on every save to the form.

Repro:

1. Add picture
2. Hit save
3. Hit save again

app/models/user.rb
```diff
@@ -11,10 +11,17 @@ class User < ApplicationRecord
...
+  validate :profile_image_size
+
   generates_token_for :password_reset, expires_in: 15.minutes do
     password_salt&.last(10)
   end
 
+  def profile_image_thumbnail
+    profile_image.variant(resize_to_limit: [150, 150])
```
zachgoll, line 22: 2 notes here:

1. Can we get this added to the sidebar and within settings menu?

CleanShot 2024-04-29 at 09 42 45


3. We could probably add `preprocessed: true` here since this shows in sidebar and we're confident it will always need to load

zachgoll (commented):
Thanks for the changes here, looks like there are just a few remaining tasks here before merging:

1. Make sure lint / test all passes (i'd run these locally with `bin/rails test:all` and `bin/erblint ./app/**/*.erb` since first-time contributors have to wait on approval for GH workflows to run)
2. Fix bug where profile deletes on second save of profile
3. Add profile image to sidebar

config/environments/production.rb
```diff
@@ -31,7 +31,7 @@
   # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :local
+  config.active_storage.service = ENV["ACTIVE_STORAGE_SERVICE"].to_sym || :local
```
zachgoll, line 34: I _think_ this will fail entirely if `ACTIVE_STORAGE_SERVICE` is not defined.  Probably better to use `ENV.fetch`

app/controllers/settings/profiles_controller.rb
crobbo: Oh good spot.

app/models/user.rb
crobbo: Done this but not sure what where I would add `preprocessed: true`.

How about adding `.processed` to the model method?

```ruby
  def profile_image_thumbnail
    profile_image.variant(resize_to_limit: [150, 150]).processed
  end
  ```

Comment (zachgoll) [reply app/models/user.rb:22]: I was just referring to the Rails guide example here:

https://guides.rubyonrails.org/active_storage_overview.html#:~:text=If%20you%20know%20in%20advance%20that%20your%20variants%20will%20be%20accessed%2C%20you%20can%20specify%20that%20Rails%20should%20generate%20them%20ahead%20of%20time%3A

File: app/views/shared/_user_profile_image.html.erb [hunk 10]
```diff
@@ -0,0 +1 @@
+<%= image_tag user.profile_image_thumbnail, class: "rounded-full" %>
```
zachgoll, line 1: I think we'll need a `h-full w-full` here:

CleanShot 2024-04-29 at 17 04 29

crobbo: Good spot I was testing with a perfectly square image.  Added `object-cover` too to prevent the image looking warped. Also form validations for image format which were missed!

app/models/user.rb
crobbo: This is a better way to do it. Nice one 👍🏻

zachgoll:
@crobbo I think all the code looks good now.  Not entirely sure what's going on with the CI checks—you _may_ need to revert back to the lockfile on `main` and call `bundle install` again since we've merged some dependabot changes recently

crobbo:
@zachgoll possibly a bad merge somewhere. Think that last commit should fix the CI issue.

zachgoll:
@crobbo checks are working now.  Just need to resolve lint errors and we'll be set.

crobbo:
@zachgoll oh there's a rubocop linter too, done 👍🏻

zachgoll:
@crobbo I merged another PR not thinking about the merge conflicts that would be introduced here.  Sorry about that, would you mind resolving those conflicts real quick and then I'll get this in?

crobbo:
@zachgoll no problem, have done this 👍🏻

zachgoll (approved):
Looks good, thanks for the work on this!  Will get it merged and the bounty should be paid out automatically.

Issue #679, opened 2024-04-25
Feature: Add transaction category management


zachgoll:
![CleanShot 2024-04-25 at 15 22 53](https://github.com/maybe-finance/maybe/assets/16676157/583cfdf8-dcf0-4a61-8ca1-8538396d88e2)

## Feature Overview

The goal of this feature is to enable a user to manage all of their transaction categories from settings and inline.

## Requirements

_If there is a missing / incorrect requirement, please leave a comment before starting work on this._

- [ ] User can view all of their categories in settings
- [ ] User can delete a category
  - [ ] from settings page
  - [ ] from inline transaction view
- [ ] When a user deletes a category, they can optionally select a new category to re-assign transactions to
- [ ] User can edit the name and color of an existing category
- [ ] User can create a category
- [ ] Tests written for all CRUD controller actions and wherever appropriate
- [ ] i18n translations provided where necessary for `en.yml`
- [ ] Ensure that both settings view (`/transactions/categories`) and inline (`/transactions`) work well together.  There are already a few partials created for editing/creating that _might_ be helpful for this issue

## Implementation Suggestions

_Below are some ideas for implementation to get you started.  Use your best judgment here—if there's a better way to do things, go for it!_

- If you're not seeing any categories in your local app, here's where those should be auto-generated - https://github.com/maybe-finance/maybe/blob/f5f624881f3b8900fa56d655ec040b5ee5d8a0af/app/controllers/registrations_controller.rb#L18
- #659 outlines a bug that happens when editing a category _inline_, which will likely affect this issue and need to be fixed

## Designs

Below are the designs you should follow while implementing this:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=3188-1313&mode=design&t=4eaJRTzT6AYU5CNv-0

## Reminders

- Make sure to review our [contributing guidelines](https://github.com/maybe-finance/maybe/blob/main/CONTRIBUTING.md) before starting on an issue
- We do our best to define a clear spec for new features and fixes, but think of them as "suggestions", not "hard requirements".  We welcome ideas and suggestions!
  - If you see missing requirements to this issue, please leave a comment below explaining what is missing and why it is important.
  - If you see a requirement that you think is _incorrect_ or _not optimal_, please leave a comment explaining what you think needs to change below.

## Bounty
There is a $250 bounty on this. To participate in the bounty:

1. Comment `/attempt #679` to indicate you plan on working on this.
[...]

PR #688, opened 2024-04-27
Implement transaction category management

josefarias:
closes https://github.com/maybe-finance/maybe/issues/679
/claim #679

This PR adds all relevant CRUD operations for transaction category management.

There are some intentional differences with the provided designs. I'll list them here together with my reasoning:

1. No double-popover design

CleanShot 2024-04-30 at 21 50 09

The modal flow works everywhere, so let's use it everywhere. It's too early to maintain two separate flows for doing the same thing. Additionally, double-popovers are bound to run into challenges with stacking contexts and alignment on smaller screens. Using the modal everywhere sidesteps these complications.

2. No double CTA on delete modal

CleanShot 2024-04-30 at 21 51 25

Instead, I've only implemented a single CTA which changes depending on whether or not a replacement category is selected. No replacement chosen -> nullify transactions, then delete. Replacement chosen -> reassign transactions, then delete.

3. I've added a way to remove a category from a transaction

CleanShot 2024-04-30 at 22 19 54

Not sure if this was the intended purpose of the inline delete action from the designs. But I thought this'd be handy.

## Demo
https://github.com/maybe-finance/maybe/assets/31393016/60a9fcde-df0c-4354-9813-8d78cdeb8ade

## Requirements
* [x] User can view all of their categories in settings
* [x] User can delete a category
  * [x] from settings page
  * [x] from inline transaction view
* [x] When a user deletes a category, they can optionally select a new category to re-assign transactions to
* [x] User can edit the name and color of an existing category
* [x] User can create a category
* [x] Tests written for all CRUD controller actions and wherever appropriate
* [x] i18n translations provided where necessary for `en.yml`
* [x] Ensure that both settings view (`/transactions/categories`) and inline (`/transactions`) work well together. There are already a few partials created for editing/creating that _might_ be helpful for this issue


app/views/shared/_modal.html.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: (content:) -%>
 <%= turbo_frame_tag "modal" do %>
-  <dialog class="bg-white border border-alpha-black-25 rounded-2xl max-h-[648px] max-w-[580px] w-full shadow-xs h-fit" data-controller="modal" data-action="click->modal#clickOutside">
+  <dialog class="bg-white border border-alpha-black-25 rounded-2xl max-h-[648px] max-w-[580px] w-min-content shadow-xs h-fit" data-controller="modal" data-action="click->modal#clickOutside">
```
josefarias, line 3: Modal width is now dictated by its contents, which makes it more flexible.

The modal for this feature is quite small — in contrast to the existing modals which are quite wide.

app/models/transaction/category.rb
```diff
@@ -46,6 +46,13 @@ def self.create_default_categories(family)
     self.insert_all(categories)
   end
 
+  def reassign_transactions_then_destroy!(replacement)
+    transaction do
+      transactions.update_all category_id: replacement&.id
```
josefarias, line 51: This won't run callbacks or expire caches. But we don't need either right now.

If we wanted to account for the above, we could iterate and update each record individually. Would probably do that in a job, too.

I think it's too early to know what we want here. Probably fine to just keep in mind and leave as-is.

virolea:
Hey @josefarias! 

I've been eying this particular one but did not end up attempting it by lack of available time. I wondered if you'd mind if I added a few comments - mainly out of curiosity - in the form of commented review?

josefarias:
Hey @virolea, sure thing!

Keep in mind this is WIP and things might still change. But happy to have you comment.

app/controllers/transactions/categories/deletions_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Transactions::Categories::DeletionsController < ApplicationController
+  before_action :set_category
+  before_action :set_replacement_category, only: :create
+
+  def new
+  end
+
+  def create
+    @category.reassign_transactions_then_destroy! @replacement_category
```
virolea, line 9: This could potentially represent hundreds of transactions, is it too much of an early optimization to consider a background job to handle the bulk reassignment?

app/views/transactions/categories/_form.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      <ul role="radiogroup" class="flex justify-between items-center py-2">
+        <% Transaction::Category::COLORS.each do |color| %>
+          <li tabindex="0"
+              role="radio"
+              data-action="click->color-select#select keydown.enter->color-select#select keydown.space->color-select#select"
+              data-value="<%= color %>"
+              class="flex shrink-0 justify-center items-center w-5 h-5 cursor-pointer hover:bg-gray-200 rounded-full">
+          </li>
+        <% end %>
+      </ul>
```
virolea, line 25: The need for JS could be drastically reduced by using HTML labels with some appropriate CSS: 

```html+erb
<%= form.collection_radio_buttons :color, Transaction::Category::COLORS, :itself, :itself, item_wrapper_tag: false do |b| %>
  <%= b.label class: "has-[:checked]:bg-red-500 lex shrink-0 justify-center items-center w-5 h-5 cursor-pointer hover:bg-gray-200 rounded-full", style: "background: #{b.value};" do %> 
    <%= b.radio_button class: :hidden %>
  <% end %> 
<% end %>
```
The code above uses a hidden input nested inside its label, and make uses of the CSS `has:checked` property to conditionally style the label based on whether or not the input is checked. 

However in order to work, this uses the same background color for all checked radios. Tailwind cannot parse classes declared dynamically. 

The alternative would be to enumerate all color inputs with their hardcoded color class. 

 Those may be the reasons why the JS route was taken!

virolea (commented):
Hey @josefarias ! 

As said earlier this is not a review, only discussions I wanted to open out of curiosity, having spent some time looking at this one as well! 

Thanks for taking the time !

app/controllers/transactions/categories/deletions_controller.rb
josefarias: Good callout! The bulk reassignment should be able to quickly handle even thousands of transactions at once without much problem (assuming a reasonably robust setup).

By performing this asynchronously, we'd be guarding against the request taking too long. That's fine! But at that scale I'd be more concerned about db throughput, not latency. Which means the actual fix would be batching the updates in addition to performing them async.

So the question becomes is it too early to batch? I think so. It's too defensive, too early. Suddenly we're deferring, batching, and throttling, and the product isn't even live yet. Which means requirements are likely to change, so there's more code to change when they do.

But people may reasonably disagree because this _will_ break with enough scale, even if it's a while before we reach it (recall we're not touching all transactions, but just the ones for this category). And the overhead is small.

Ultimately up to the maintainers how defensive they want to be, so let me know @zachgoll!

app/views/transactions/categories/_form.html.erb
josefarias: > Those may be the reasons why the JS route was taken!

Nailed it! That's exactly the reason. I'm not convinced JS is the way to handle this either. Could feasibly declare bg-color and box-shadow TW classes for these colors and go the radio button route instead. Planning to think some more about this before marking as ready for review.

app/controllers/transactions/categories/deletions_controller.rb
zachgoll: While certainly not a bad thought, I think @josefarias is correct that we don't need to worry about this optimization yet.  Our main priority is to get features built as fast as possible and take a "Just in time" approach to optimizations rather than "Just in case".

app/views/transactions/categories/_form.html.erb
josefarias: Seems like we're using a similar approach for merchants. I think we should consolidate at some point @zachgoll. Leaving this JS implementation in for now.

zachgoll: Sounds good, will keep this in mind.  I'm fine with leaving the duplicate logic here for now.

app/controllers/transactions/merchants_controller.rb
```diff
@@ -11,9 +11,9 @@ def new
 
   def create
     if Current.family.transaction_merchants.create(merchant_params)
-      redirect_to transactions_merchants_path, notice: t(".success")
+      redirect_to transaction_merchants_path, notice: t(".success")
     else
-      render transactions_merchants_path, status: :unprocessable_entity, notice: t(".error")
+      render transaction_merchants_path, status: :unprocessable_entity, notice: t(".error")
```
zachgoll, line 16: Know this was existing and not part of this issue, but if I'm not mistaken, `render transaction_merchants_path` is not valid, and `notice: t(".error")` will render a green notification with an error message (I must have missed this entirely in a prior review).

As a temp fix, you think something like this would work?

```rb
Current.family.transaction_merchants.create!(merchant_params)
redirect_to transaction_merchants_path, notice: t(".success")
```

app/helpers/transactions/categories_helper.rb
```diff
@@ -1,2 +1,7 @@
 module Transactions::CategoriesHelper
+  def null_category
+    Transaction::Category.new \
+      name: "Uncategorized",
+      color: Transaction::Category::UNCATEGORIZED_COLOR
+  end
```
zachgoll, line 6: nice idea here

zachgoll (approved):
Looks great.  Below are my responses to the design modifications:

RE: Double popover - 100% agree.  I was fighting this a lot when originally creating some of the stacked menus.  There's just a lot of additional JS to write and it also introduces "floating" elements which are always a pain IMO.  I've spoken to our designer about this and we'll be more heavily leveraging the modal moving forward.

RE: Double CTA - your solution makes sense to me

RE: Remove category - nice addition here!

app/controllers/transactions/categories/deletions_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  before_action :set_replacement_category, only: :create
+
+  def new
+  end
+
+  def create
+    @category.replace_and_destroy! @replacement_category
+
+    redirect_back_or_to transactions_path, notice: t(".success")
+  end
```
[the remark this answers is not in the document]
josefarias: Ah of course!

This goes one step further than the usual Rails controllers meant for invoking CRUD actions on active records.

In this case, we're CRUD-ing an abstract concept: a deletion.

Thinking of deletions as a resource allows us to write very clean Rails code. `DeletionsController#new` gives us a place to house the modal and its logic (including manipulating the CTA). `DeletionsController#create` is what naturally follows that `new` view. You create a deletion. Or, initiate it. The key thing is that the deletion is the resource here, not the category.

You can also think of this by eliminating the other possible RESTful actions. We're not fetching, updating, or destroying a deletion. We're creating it!

app/controllers/transactions/merchants_controller.rb
zachgoll: Yep, early on here, I'm generally thinking throwing 500s is a better strategy for us.  We _should_ have html validations and we can always come back later and handle specific errors with notifications if absolutely critical.

I think all the `if/else` stuff sort of organically started proliferating throughout the project (copy/paste, etc.), but I'll try to use this convention more moving forward.

app/controllers/transactions/categories/deletions_controller.rb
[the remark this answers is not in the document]
zachgoll: Got it, thanks for the explanation here!  I was not thinking of it as a resource, but that makes complete sense now that you explain it that way.

app/views/transactions/categories/_form.html.erb
virolea: > Could feasibly declare bg-color and box-shadow TW classes for these colors and go the radio button route instead. 

One can also use arbitrary color values with tailwind by using utiliies such as`bg-[#e99537]`. A comment can be added to the code to "trick" tailwind to generate the classes, so one can use it later on. Food for thought ! :)

zachgoll:
@josefarias I'm going to merge this as-is and address some of those controller comments I made in a subsequent PR as I'm about to update some dependencies (and Ruby version) for the project and don't want you to have to deal with any conflicts that creates.  Thanks for the work on this!

PR #697, opened 2024-04-29
Add trends to sidebar account list

### LINKED ISSUE #647: Improvement: Add trendlines and styling to sidebar account list
CleanShot 2024-04-18 at 14 35 30

## Designs

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=2828-2845&mode=design&t=Y0hF2UeUXaOhxGL4-0

## Overview

This issue is for getting the sidebar accounts list in parity with the official designs.

## Requirements

- [ ] Add trendline and percent change to account balance
- [ ] Add a period selector (above the list) and allow the following values: `Period.last_7_days`, `Period.last_30_days`.  Default should be `Period.last_30_days`.  Labels should be `7D` and `1M`
- [ ] Add icons left of account name (we don't have bank logos yet, so use the placeholder value design for all accounts)

## Implementation Suggestions

- Most of the backend logic for calculating the trendlines and percent change by period is already completed.  See: https://github.com/maybe-finance/maybe/blob/f5f624881f3b8900fa56d655ec040b5ee5d8a0af/app/models/account.rb#L68

## Reminders

JoshAntBrown:
/claim #647
Resolves #647 

This PR adds filterable series and trend information to the account list in the sidebar. In order to achieve this I have created a new route on the accounts controller that renders the account list in a turbo frame that can then be used to update the frame within the sidebar. The period select simply navigates to that url within the frame and sets the selected period.

As part of this I've also updated the designs to match those in the latest spec.

https://github.com/maybe-finance/maybe/assets/1793797/6b745d1d-0b9b-42f5-a2b7-caaa2ea500c6

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -14,6 +14,9 @@ class AccountsController < ApplicationController
     @account_groups = Current.family.accounts.by_group(period: @period, currency: Current.family.currency)
   end
 
+  def list
+  end
+
   def new
     @account = Account.new(
       balance: nil,
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -27,8 +27,8 @@ module ApplicationHelper
     render partial: "shared/modal", locals: { content: content }
   end
 
-  def account_groups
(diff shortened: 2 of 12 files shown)


zachgoll (approved):
Awesome stuff.  Fun to see a perfect use case for turbo frames!

PR #698, opened 2024-04-30
Add ability to delete Maybe account

### LINKED ISSUE #675: Feature: Allow user to delete their Maybe account and all data

JoshAntBrown:
/claim #675
Resolves #675 

/split @jemiluv8

This PR adds the ability to delete an account.

This approach keeps most of the logic centred around account deactivation, contained within the User model itself. 

It uses a validation method to ensure admins aren't able to deactivate their user unless they are the last remaining user in the family. Should validation pass and the account is deactivated then it automatically schedules a UserPurgeJob to call the purge method. If there's only one user in the family then purge will destroy the family, otherwise it will just destroy the user. The associated data to the User or Family is currently taken care of by the dependant destroy actions on the relations in each of the models, if needed though the purge function could be expanded to handle those.

To prevent the user from logging in again after deactivating their account before the user or family is purged we modify their email address.

https://github.com/maybe-finance/maybe/assets/1793797/922b0f02-4e15-4c4a-86dd-ed77d055d47d

### DIFF
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -17,6 +17,15 @@ class Settings::ProfilesController < ApplicationController
     end
   end
 
+  def destroy
+    if Current.user.deactivate
+      logout
+      redirect_to root_path, notice: t(".success")
+    else
+      redirect_to settings_profile_path, alert: Current.user.errors.full_messages.to_sentence
+    end
+  end
+
   private
 
   def user_params
--- a/app/javascript/controllers/application.js
+++ b/app/javascript/controllers/application.js
@@ -10,7 +10,7 @@ Turbo.setConfirmMethod((message) => {
   const dialog = document.getElementById("turbo-confirm");
 
   try {
-    const { title, body, accept } = JSON.parse(message);
+    const { title, body, accept, acceptClass } = JSON.parse(message);
 
     if (title) {
       document.getElementById("turbo-confirm-title").innerHTML = title;
@@ -22,6 +22,10 @@ Turbo.setConfirmMethod((message) => {
 
     if (accept) {
       document.getElementById("turbo-confirm-accept").innerHTML = accept;
+    }
+
+    if (acceptClass) {
+      document.getElementById("turbo-confirm-accept").className = acceptClass;
     }
   } catch (e) {
     document.getElementById("turbo-confirm-title").innerText = message;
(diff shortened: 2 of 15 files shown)


zachgoll (commented):
Everything here looks good and I like the "deactivate" mental model a lot.

Ultimately, I think this is the best solution so we'll get it merged, but given #683 has been open for several days with lots of iterations and collaborations, we'll throw a split on the bounty to embrace the collaborative nature of OSS.

JoshAntBrown:
Thanks, and yeah that sounds good re: the split, I actually feel a lot better about that approach. I was a bit cautious about making the PR given the outstanding one, but I had a few ideas I wanted to try out and explore so decided to go ahead anyway. 

I wasn't aware that splitting was a thing as part of Algora, very cool!

PR #702, opened 2024-05-02
Deps cleanup

zachgoll:
The major change here is removing some of the ActiveStorage adapters (which had severely outdated transitive deps).  This was causing some unexpected behavior in tests.

Leaving S3 as the sole adapter for now and can introduce additional ones as requested.

## Tests

Will be introducing a follow-up PR to re-introduce tests removed here without the flakiness.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gem "rails", github: "rails/rails", branch: "main"
 
 # Drivers
 gem "pg", "~> 1.5"
-gem "redis", ">= 4.0.1"
 
 # Deployment
 gem "puma", ">= 5.0"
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -33,37 +32,35 @@ gem "ransack"
 gem "stackprof"
 gem "sentry-ruby"
 gem "sentry-rails"
-gem "rails-settings-cached"
-gem "octokit"
 
 # Active Storage
 gem "aws-sdk-s3", require: false
-gem "azure-storage-blob", "~> 2.0", require: false
```
zachgoll, line 41: After a closer look at this—was causing Faraday to be forced to an earlier version and the maintainers have issued a deprecation notice (still in Rails guides though, which was why it was introduced)

PR #703, opened 2024-05-02
Test environment stability improvements

zachgoll:
## Overview

This is the second part of a series of PRs aimed to stabilize our dependencies and test suite so we're not getting unpredictable failures that have been happening as of late.

There will be 1-2 more PRs to finish this off, but wanted to open this in small increments for documentation purposes.

## App Mode

As part of this PR, I have moved the config for self hosting directly into the environment configs under the key `app_mode`, which can either be `self_hosted` or `managed`.

```rb
Rails.configuration.app_mode.self_hosted? # true | false
```

## Good Job config

By default, GoodJob offers different execution modes based on the Rails environment (i.e. prod defaults to `:external`, dev to `:async`, test to `:inline`), so I've consolidated the config into `application.rb`.

Prior, we didn't have anything set in test, so it was defaulting to the Rails `:async` adapter (rather than the GoodJob `:inline` adapter as expected).

## Test environment, overriding ENV in tests

Since the Maybe app will need to be tested under both "self hosted" mode and "managed" mode, I've introduced a first class helper to facilitate this:

```rb
test "something only relevant to self hosting mode" do
  with_self_hosting do 
    # Assertions
  end
end
```

Additionally, I've moved `dotenv-rails` out of the `:test` group to be more explicit with the test environment and added another helper (leveraging the `climate_control` gem) to explicitly override environment variables for individual tests:

```rb
test "something with ENV override(s)" do
  with_env_overrides SOME_ENV: "some value", ANOTHER_VAR: "another value" do
    # Assertions
  end
end
```

### DIFF
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -12,14 +12,16 @@ amazon:
   secret_access_key: <%= ENV["S3_SECRET_ACCESS_KEY"] %>
   region: <%= ENV["S3_REGION"] || "us-east-1" %>
   bucket: <%= ENV["S3_BUCKET"] %>
-google:
-  service: GCS
-  project: <%= ENV["GCS_PROJECT"] %>
-  credentials: <%= Rails.root.join("gcp-storage-keyfile.json") %>
-  bucket: <%= ENV["GCS_BUCKET"] %> 
 
-azure:
-  service: AzureStorage
-  storage_account_name: <%= ENV["AZURE_STORAGE_ACCOUNT_NAME"] %>
-  storage_access_key: <%= ENV["AZURE_STORAGE_ACCESS_KEY"] %>
-  container: <%= ENV["AZURE_STORAGE_CONTAINER"] %> 
+# Removed in #702.  Uncomment, add gems, update .env.example to enable.
+#google:
+#  service: GCS
+#  project: <%#= ENV["GCS_PROJECT"] %>
+#  credentials: <%#= Rails.root.join("gcp-storage-keyfile.json") %>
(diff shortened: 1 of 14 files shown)


config/storage.yml
```diff
@@ -12,14 +12,16 @@ amazon:
...
+#  service: GCS
+#  project: <%#= ENV["GCS_PROJECT"] %>
+#  credentials: <%#= Rails.root.join("gcp-storage-keyfile.json") %>
+#  bucket: <%#= ENV["GCS_BUCKET"] %>
+
+#azure:
+#  service: AzureStorage
+#  storage_account_name: <%#= ENV["AZURE_STORAGE_ACCOUNT_NAME"] %>
+#  storage_access_key: <%#= ENV["AZURE_STORAGE_ACCESS_KEY"] %>
+#  container: <%#= ENV["AZURE_STORAGE_CONTAINER"] %>
```
zachgoll, line 27: Unrelated cleanup here that I missed in #702

test/system/settings_test.rb
```diff
@@ -26,25 +26,6 @@ class SettingsTest < ApplicationSystemTestCase
...
-      assert_selector "h1", text: header_text
-      assert_current_path path
-      assert_link "Next", href: next_setting_path if next_setting_path.present?
-      assert_link "Back", href: prev_setting_path if prev_setting_path.present?
-    end
-
-    # Conditional nav items don't show by default
-    assert_no_text "Self-Hosting"
-  end
-
```
zachgoll, line 47: Will be re-introduced in follow-up PR.

Issue #699, opened 2024-04-30
Feature: Add Test Coverage Generation


jemiluv8:
## Feature Overview
The current `rake test` command doesn't generate test coverage report.
This can be useful especially for new features that require testing.

## Requirements
- [ ] Support generation of test coverage report when running tests

## Implementation Suggestions
Probably integrate simplecov-ruby or similar compatible library for test coverage report generation

## Designs
Not Applicable

## Reminders

- Make sure to review our [contributing guidelines](https://github.com/maybe-finance/maybe/blob/main/CONTRIBUTING.md) before starting on an issue
- We do our best to define a clear spec for new features and fixes, but think of them as "suggestions", not "hard requirements".  We welcome ideas and suggestions!
  - If you see missing requirements to this issue, please leave a comment below explaining what is missing and why it is important.
  - If you see a requirement that you think is _incorrect_ or _not optimal_, please leave a comment explaining what you think needs to change below.

jemiluv8:
@garhuku, FYI, I'm not a maintainer on this project and this issue doesn't have a bounty on it at the moment.
Also, you might want to ping the maintainers just in case they have something to add or even if they truly want test coverage on their project.

zachgoll:
@jemiluv8 @gathuku I think we could add the `simplecov` gem to generate a coverage report, but given how early we are with this project, this should be for _convenience only_ at the moment (i.e. a dev writing tests locally can easily see coverage to help them identify important areas that they missed)

In other words, we can set this up, but let's not enforce anything in CI for now.

gathuku:
Nice! @jemiluv8 can I help you with this?

PR #705, opened 2024-05-02
SimpleCov Coverage Report

Harry-kp:
## Description
fixes https://github.com/maybe-finance/maybe/issues/699
- This PR adds the functionality of generating a Code Coverage Report in Simple HTML format. The generation of code coverage is based on the ENV variable which is defaulted to true (meaning a report will be generated every time when the test runs).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,6 +51,8 @@ group :development, :test do
   gem "debug", platforms: %i[ mri windows ]
   gem "brakeman", require: false
   gem "rubocop-rails-omakase", require: false
+  gem "simplecov", require: false
+  gem "dotenv-rails"
   gem "i18n-tasks"
   gem "erb_lint"
 end
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,9 @@ DB_PORT=5432
 POSTGRES_PASSWORD=postgres
 POSTGRES_USER=postgres
 
+# Coverage Report
+ENABLE_CODE_COVERAGE_REPORT=true
+
 # App Domain
 # This is the domain that your Maybe instance will be hosted at. It is used to generate links in emails and other places.
 APP_DOMAIN=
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -51,6 +51,8 @@ group :development, :test do
   gem "debug", platforms: %i[ mri windows ]
   gem "brakeman", require: false
   gem "rubocop-rails-omakase", require: false
+  gem "simplecov", require: false
+  gem "dotenv-rails"
```
zachgoll, line 55: I think `simplecov` should only be in the `:test` group right?

Also, I recently moved `dotenv-rails` to the `:development` group only to avoid unexpected, failing tests when a dev has their `.env` setup incompatible with the test expectations.  I think this will still work if we pass the ENV to the test command right?

zachgoll (commented):
I pulled this down locally and ran it and for some reason, was getting some odd results.  For example:

CleanShot 2024-05-02 at 19 26 39

We have `test/models/upgrader_test.rb` which tests a _majority_ of `upgrader.rb`, but this coverage report shows 0% coverage.

Are you seeing a similar result locally?

zachgoll:
@harrrykp also, for future PRs, would you mind configuring your Github email (in settings) to match your git client email?  It makes it a little easier to review with a single user in the commit history.

zachgoll:
See: https://github.com/simplecov-ruby/simplecov/issues/718

Looks like Rails test parallelization is causing coverage to drop:

```rb
parallelize(workers: :number_of_processors)
```

We should be seeing ~85% coverage on this codebase.

Gemfile
Harry-kp: Yes, this should be in the test group and it works if we pass the env from the CMD. Working on changes

Harry-kp:
> I pulled this down locally and ran it and for some reason, was getting some odd results. For example:
> 
> CleanShot 2024-05-02 at 19 26 39
> 
> We have `test/models/upgrader_test.rb` which tests a _majority_ of `upgrader.rb`, but this coverage report shows 0% coverage.
> 
> Are you seeing a similar result locally?

Yes, happening same on my local.

Harry-kp:
> See: [simplecov-ruby/simplecov#718](https://github.com/simplecov-ruby/simplecov/issues/718)
> 
> Looks like Rails test parallelization is causing coverage to drop:
> 
> ```ruby
> parallelize(workers: :number_of_processors)
> ```
> 
> We should be seeing ~85% coverage on this codebase.

Ohh, got it. Will be reviewing this issue for a possible solution.

Harry-kp:
> @harrrykp also, for future PRs, would you mind configuring your Github email (in settings) to match your git client email? It makes it a little easier to review with a single user in the commit history.

Sure @zachgoll , need to check my git configs. Will fix this behaviour in future PRs.

zachgoll:
@harrrykp closing this out for now.  If you'd like to continue work on it just let me know and I'll re-open!

PR #707, opened 2024-05-02
Dashboard empty view

### LINKED ISSUE #706: Improvement: Add dashboard "empty" view
Currently, when there are no accounts, we're still showing insights cards.  We should show the following empty state:

https://www.figma.com/file/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?type=design&node-id=3435-238&mode=design&t=6v4Cy35JLnEVjtjq-0

CleanShot 2024-05-02 at 15 06 50

claudey:
Fixes #706 

- Show title, button and summary when no accounts are added.
- Increased the padding-right value to make the button balanced.

Screenshot 2024-05-02 at 9 47 36 PM

### DIFF
--- a/app/views/pages/dashboard.html.erb
+++ b/app/views/pages/dashboard.html.erb
@@ -3,9 +3,11 @@
     <div>
       <h1 class="sr-only">Dashboard</h1>
       <p class="text-xl font-medium text-gray-900 mb-1"><%= t(".greeting", name: Current.user.first_name ) %></p>
-      <p class="text-gray-500 text-sm"><%= t(".subtitle") %></p>
+      <% if !@accounts.blank? %>
+        <p class="text-gray-500 text-sm"><%= t(".subtitle") %></p>
+      <% end %>
     </div>
-    <%= link_to new_account_path, class: "flex text-white text-sm font-medium items-center gap-1 bg-gray-900 rounded-lg p-2", data: { turbo_frame: "modal" } do %>
+    <%= link_to new_account_path, class: "flex text-white text-sm font-medium items-center gap-1 bg-gray-900 rounded-lg p-2 pr-3", data: { turbo_frame: "modal" } do %>
       <%= lucide_icon("plus", class: "w-5 h-5") %>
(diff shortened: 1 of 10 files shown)


app/views/pages/dashboard.html.erb
```diff
@@ -3,13 +3,27 @@
...
+        <p class="text-gray-500 text-sm"><%= t(".subtitle") %></p>
+      <% end %>
     </div>
-    <%= link_to new_account_path, class: "flex text-white text-sm font-medium items-center gap-1 bg-gray-900 rounded-lg p-2", data: { turbo_frame: "modal" } do %>
+    <%= link_to new_account_path, class: "flex text-white text-sm font-medium items-center gap-1 bg-gray-900 rounded-lg p-2 pr-3", data: { turbo_frame: "modal" } do %>
       <%= lucide_icon("plus", class: "w-5 h-5") %>
       <span><%= t(".new") %></span>
     <% end %>
   </header>
+  <% if @accounts.blank? %>
```
zachgoll, line 15: In order to use the `@accounts` instance variable, it will need to be defined within the `pages_controller#dashboard` action.

Since we already have an `@account_groups` variable defined there, I'd recommend a small refactor like so:  

```ruby
@accounts = Current.family.accounts
@account_groups = @accounts.by_group(period: @period, currency: Current.family.currency)
```

zachgoll (commented):
Nice work here! There was one small change that will need to be made that I left a comment about.

In addition, could we also perform this logic on the `/accounts/summary` page as well?  To do this, I'd recommend placing the "No Accounts" message in a partial so that you can reuse it across these two views.

zachgoll:
@claudey also, don't worry about the failing tests—I'll be coming back to fix those after I get some other work done.

app/views/pages/dashboard.html.erb
claudey: Thanks for the recommendation. I've added it in now.

zachgoll (approved):
Looks good, nice work!

PR #711, opened 2024-05-03
Allow two decimals for value in transaction history

### LINKED ISSUE #669: History value unable to have cents
**Describe the bug**
This is a very similar issue to #613, but I found another place that is unable to have cents in the value, specifically a new entry in the history of an account.

**To Reproduce**
1. Go to an account
2. Click on "History"
3. Click on "New entry"
4. Attempt to put in cents in the "Value" field

**Expected behavior**
I should be able to record cents.

**Screenshots / Recordings**
image

sarvaiyanidhi:
Fixes #669

This fix now allows to add cents to transaction history value field.

transaction_history


zachgoll (approved):
Thanks for the fix :)

PR #712, opened 2024-05-03
Added custom template to show unsupported browser error

### LINKED ISSUE #637: Add custom template when `allow_browser` blocks a browser
Currently, we're showing a generic error message when a browser is blocked, making it hard to debug self-hosted apps:
image

We should add a custom `public/406-unsupported-browser.html` file to give the user more feedback.

This recently went into Rails `main` and used to be `public/426.html` - https://github.com/rails/rails/pull/51545

sarvaiyanidhi:
Fixes #637 

Tested error on Floorp web browser which is currently not supported in the list of allowed modern browsers.

unsupported_browser

### DIFF
--- a/public/406-unsupported-browser.html
+++ b/public/406-unsupported-browser.html
@@ -1,0 +1,97 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Your browser is not supported (426)</title>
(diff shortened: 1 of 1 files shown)


zachgoll (approved):
Awesome, looks good!

PR #714, opened 2024-05-04
Draft: Add Tags

flacnut:
*I am not familiar with Ruby or Rails, constructive feedback welcomed!*

This PR is a draft of possible ways to add tags.

[Context (713)](https://github.com/maybe-finance/maybe/discussions/713)

I have in the past used the ORM / ActiveRecord completely, mapping a many-to-many with a join table, however this negatively impacted the performance of my application for the use case of generating many charts which include tag filters.

Tags_demo_data_2

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -82,7 +82,11 @@ class AccountsController < ApplicationController
   private
 
   def set_account
-    @account = Current.family.accounts.find(params[:id])
+    begin
+      @account = Current.family.accounts.find(params[:id])
+    rescue ActiveRecord::RecordNotFound
+      redirect_to accounts_path, alert: "Account not found"
+    end
   end
 
   def account_params
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -23,6 +23,14 @@ class Transaction < ApplicationRecord
       .where("er.rate IS NOT NULL OR transactions.currency = ?", currency)
   }
 
+  def tags
+    Tag.where(id: self.tag_ids)
+  end
+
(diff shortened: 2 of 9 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -82,7 +82,11 @@ def sync
   private
 
   def set_account
-    @account = Current.family.accounts.find(params[:id])
+    begin
+      @account = Current.family.accounts.find(params[:id])
```
zachgoll, line 86: In Rails this error handling is typically not necessary as the `@account` would just be `nil` if not found.

zachgoll (commented):
Appreciate the PR here, as I mentioned in the [Discussion](https://github.com/maybe-finance/maybe/discussions/713#discussioncomment-9330362), I definitely think tags are a must-have for the system.

We don't have any internal designs for this quite yet, so I think in order to introduce this, it would have to be _backend only_ for the moment.

If you wanted to go down that route, I'd probably favor a many to many relationship as ActiveRecord has a [lot of neat helpers](https://edgeguides.rubyonrails.org/association_basics.html#methods-added-by-has-and-belongs-to-many) and handles this quite well.

app/controllers/accounts_controller.rb
flacnut: Yeah, this was actually a fix for something else. If you navigate to an `account_id` that doesn't exist, rails spits out stack traces which is not very graceful.

zachgoll:
@flacnut going to close this out for now and we can re-open it (or open a new one) when we get the designs for it.  We've been doing some roadmap planning and will definitely be getting this in soon!

PR #725, opened 2024-05-07
Add ARM target for Docker images

zachgoll:
Adds `linux/arm/v8` for Docker image publishing to run Maybe on Raspberry Pi / Mac devices.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,6 @@ name: CI
 
 on:
   pull_request:
-  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -2,8 +2,6 @@ name: CI
 
 on:
   pull_request:
-  push:
-    branches: [main]
```
zachgoll, line 6: Redundant now that we have a test -> build -> push workflow

Issue #565, opened 2024-03-25
Request: make the app port part of env config


karanj:
Since the default port 3000 clashes with Grafana's default port, and I don't particularly want to mess with my existing Grafana setup, I had to change the config and code in a few places to allow me to start up the dev container without clashing with Grafana occupying port 3000.

Suggestion is to make the app port a config item in the .env file to enable easy changing of this attribute for those running with complex setups.

claudey:
@zachgoll Are you able to close this since #726 has been merged?

zachgoll:
@claudey yep, closing it out!

In the future, you can [reference an issue in the PR](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) and it will automatically close on merge.

claudey:
Thanks, noted.

PR #726, opened 2024-05-08
Allow users to set port in env

claudey:
Users can set a custom `PORT` variable in .env if 300 is used by other other apps.

Solves #565 

https://github.com/maybe-finance/maybe/assets/1866816/380dbe54-7cc3-497d-9637-890764395de0


bin/dev
```diff
@@ -6,7 +6,7 @@ if ! gem list foreman -i --silent; then
 fi
 
 # Default to port 3000 if not specified
-export PORT="${PORT:-3000}"
+export PORT="${PORT:-ENV.fetch("PORT") { 3000 }}"
```
zachgoll, line 9: I don't think we'll need to change this—it already allows for a `PORT` configuration with a default

README.md
```diff
@@ -53,7 +53,7 @@ bin/dev
 rake demo_data:reset
 ```
 
-And visit http://localhost:3000 to see the app. You can use the following credentials to log in (generated by DB seed):
+And visit http://localhost:3000 to see the app. You can also set a custom port with the `PORT` variable in .env file. You can use the following credentials to log in (generated by DB seed):
```
zachgoll, line 56: I think we can remove this and leverage the `.env.example` file as documentation for new devs

zachgoll (commented):
Thanks for the changes!

zachgoll (approved):

PR #728, opened 2024-05-08
Get default valuation currency from the related account, not family

jakubkottnauer:
Before this fix, manually created valuations always used the currency of the family (the base currency), even when the valuations were created for accounts in foreign currencies.

Screenshot 2024-05-08 at 19 50 33


zachgoll (approved):
Makes sense!  Thanks for the fix

Issue #724, opened 2024-05-07
When exchange rates are not available, foreign currency account rollups should show a warning UI + message


jakubkottnauer:
**Describe the bug**
The balance sums shown next to the collapsible account groups in the sidebar don't include accounts in foreign currencies, if today's exchange rates for those currencies are missing in the database.

**To Reproduce**
1. Create a new account in a foreign (= different from your base) currency (one that you haven't used before) with a non-zero balance
2. Observe that in the sidebar, the balance of the new account isn't included in the sum for that account type group

**Expected behavior**
Account group balances in the sidebar should take foreign currencies into account. In the screenshot below, the number next to `Other Liability` should be roughly `$1020.00` (200 SEK ~ 20 USD)

**Screenshots / Recordings**
<img width="321" alt="Screenshot 2024-05-07 at 22 31 03" src="https://github.com/maybe-finance/maybe/assets/113784/c6b06d60-8dbe-43d0-9744-75e08429082a">

**Additional context**
The issue is caused by a missing exchange rate when this check is performed: https://github.com/maybe-finance/maybe/blob/79789bd69630ad29d65b53f138da2ab6abaf69e4/lib/money.rb#L31

Perhaps we should sync the account right after it's created? Or at the very least call `account.sync_exchange_rates`. We'll also need to fetch exchange rates periodically or upon login, otherwise the bug reappears the next day.

zachgoll:
@jakubkottnauer since populating exchange rates requires a user to _potentially_ spend money, the app needs to _assume_ that rates cannot always be populated in full and have sensible ways of handling that.

The behavior we're seeing here is the original intention, but may not be the best way to handle it.

What do you think is the best way to display the data when a user has not synced exchange rates?

zachgoll:
I haven't quite gotten to this yet, but my thinking was that in these scenarios where **user action** is required to fix the data, we display global warnings to alert them that their data is not 100% correct and they need to do something to fix it:
![CleanShot 2024-05-08 at 08 40 08](https://github.com/maybe-finance/maybe/assets/16676157/e47ab306-ba99-4587-96db-288598e79ddd)

jakubkottnauer:
Gotcha, yes, I think what you suggest would be fine. Maybe we could, in addition to the user action prompt, also show a little icon next to all the balances that are impacted by the missing exchange rate (as an example, in your screenshot the USD accounts in the sidebar would be shown normally, but foreign currency accounts would have the icon next to the balance)

[...]

PR #735, opened 2024-05-10
Add start balance to manual accounts

jakubkottnauer:
This PR makes `balance` of manual accounts be automatically updated whenever a new valuation or a transaction is added to that account. In this first iteration, the solution "hijacks" the existing sync flow by first computing the correct current balance which is then treated as the "source of truth" in the rest of the sync (as if it was an account connected to a 3rd party provider).

Current balance of a manual account is calculated as follows:
- take the latest valuation's value, or `start_balance` if no valuation exists
- get the sum of all transactions that occurred  between the valuation (or all transactions if no valuation exists) and today
- -> sum of these two numbers gives you the current balance

## Changes
A new `accounts.start_balance` column has been added. I have considered making the migration populate the column for all accounts (as we only have manual accounts at the time of running the migration) but then to be consistent I would also have to change all of the seed data to create manual accounts, which doesn't sound like something we'd like to do. Happy to discuss this further.

All newly created accounts will be manual by default as we don't have support for connected accounts yet.

## Next steps

- Start balance of a manual account should be made editable (are there designs for an account edit dialog yet?)
- Valuations and transactions can have their `date` set in the future. This means that accounts need to be synced periodically to make sure the balance is up to date every day the user opens Maybe. Somewhat similar to the issue https://github.com/maybe-finance/maybe/issues/724 and can be fixed the same way.
- Come up with a better way of telling manual and connected accounts apart - currently it's done by checking `account.start_balance.nil?`. In the future it would be nicer to check if  a provider connection exists for the account, but we're not quite there yet.
- We should have some UI elements distinguishing between manual/connected accounts - maybe we could temporarily show a fake-ish error message on connected accounts saying that the "provider for this account is inactive"? After this PR, some accounts will react to transactions/valuations being added (manual accounts) while others will not (connected accounts) and this might cause confusion without any way of telling them apart.


jakubkottnauer:
@zachgoll After this is reviewed I will need some help with updating the Google Sheet to include the new manual account fixtures.

zachgoll:
@jakubkottnauer will take a look this afternoon!

zachgoll:
@jakubkottnauer thanks for the work on this!  I've addressed some of the various questions below along with some overall thoughts:

> Start balance of a manual account should be made editable (are there designs for an account edit dialog yet?)

We do have some designs, I'll be getting them out for dev soon!

> Valuations and transactions can have their date set in the future.

Should future dates be valid?  I know we're not enforcing this currently but it may be a good idea to set a max date of `Date.current` for valuations and transactions.  

> Come up with a better way of telling manual and connected accounts apart - currently it's done by checking account.start_balance.nil?. In the future it would be nicer to check if a provider connection exists for the account, but we're not quite there yet.

> We should have some UI elements distinguishing between manual/connected accounts - maybe we could temporarily show a fake-ish error message on connected accounts saying that the "provider for this account is inactive"? After this PR, some accounts will react to transactions/valuations being added (manual accounts) while others will not (connected accounts) and this might cause confusion without any way of telling them apart.

There are some designs in the works to distinguish between the two, but I'm starting to think that we should simply overwrite the existing sync process and tests and assume that all accounts are manual for now.  I'm completely fine deleting the existing sync logic and making this the new "base case" until we have connected accounts to work with.  We can always go back in source control to find the "connected account logic".  For now, I think it will be a lot simpler to deal with concrete flows rather than trying to design for the future.

Off the top of my head, I think the pieces that we'd need to address would be:

### Strategy

In my head, a "start balance" is ultimately the same thing as a `Valuation`.  It has a monetary value + date.

Rather than introducing a new DB field, I think we can just use the "oldest Valuation" as our start balance.  When a user creates an account, we can ask them for this information and our system will automatically create the `Valuation`:

1. Current day balance
2. Account start value (we create a `Valuation` equal to the date + amount input)

If we did this, I _think_ we'd be able to keep almost all of our existing balance calculator logic.  The only thing that would then change would be this line:

https://github.com/maybe-finance/maybe/blob/5ed2c47c200e11840527f94a9fa164656ba3da40/app/models/account/balance/calculator.rb#L34

We would then need to make sure that on every sync, `@account.balance` is updated to reflect the calculated series.

jakubkottnauer:
Thank you for the review! I'll incorporate the changes you've suggested, making manual accounts the  "base case" for now 👍 

> Should future dates be valid?
Some other apps I have used allow future transactions, though I don't see much value in allowing future valuations. And if we won't be allowing those, maybe it makes not to allow transactions either for consistency 😄 Up to you!

> Rather than introducing a new DB field, I think we can just use the "oldest Valuation" as our start balance
I like the proposed solution, I'll make the change today/tomorrow

zachgoll:
@jakubkottnauer I think for now we can enforce a max date of `Date.current` for simplicity.  I could see a case for future dates for both, but it's not worth the complexity right now:

- Valuation - for example, a user provides a "projected" account balance
- Transaction - for example, a user provides a "recurring" transaction that is expected

jakubkottnauer:
@zachgoll I've implemented the changes, seems to work well

app/models/account/syncable.rb
```diff
@@ -16,6 +16,10 @@ def sync(start_date = nil)
         calculator.calculate
         self.balances.upsert_all(calculator.daily_balances, unique_by: :index_account_balances_on_account_id_date_currency_unique)
         self.balances.where("date < ?", effective_start_date).delete_all
+        new_balance = calculator.daily_balances.last[:balance]
+        self.balance = new_balance
+        self.save!
+
         update!(status: "ok", last_sync_date: Date.today)
```
zachgoll, line 23:
```suggestion
        update!(status: "ok", last_sync_date: Date.today, balance: new_balance)
```

app/models/exchange_rate.rb
```diff
@@ -18,5 +18,12 @@ def find_rate_or_fetch(from:, to:, date:)
     def get_rate_series(from, to, date_range)
       where(base_currency: from, converted_currency: to, date: date_range).order(:date)
     end
+
+    def convert(value:, from:, to:, date:)
+      rate = ExchangeRate.find_by(base_currency: from, converted_currency: to, date:)
+      raise "Rate for #{from} to #{@account.to} not found" unless rate
```
zachgoll, line 24:
```suggestion
      raise "Conversion from: #{from} to: #{to} on: #{date} not found" unless rate
```

zachgoll (commented):
Looks good!  Can you update the tests to expect this new convention?  Will get it merged after those pass.

jakubkottnauer:
@zachgoll I've pushed a few new tests. Can you help out with updating the CSVs?

zachgoll:
@jakubkottnauer here is the link to the Google sheet that I've been using to visualize the test cases: 

https://docs.google.com/spreadsheets/d/18LN5N-VLq4b49Mq1fNwF7_eBiHSQB46qQduRtdAEN98/edit?usp=sharing

I'll go ahead and get this updated after this PR is merged.  But for now, I think all we need to do is update the final row in each of these CSVs:

https://github.com/maybe-finance/maybe/blob/daf7ff8ef4da8718a1038f4c679539f2305195f2/test/fixtures/family/expected_snapshots.csv?plain=1#L32

https://github.com/maybe-finance/maybe/blob/daf7ff8ef4da8718a1038f4c679539f2305195f2/test/fixtures/account/expected_balances.csv?plain=1#L32

jakubkottnauer:
@zachgoll All tests are green now 😄

zachgoll (approved):
Awesome, looks great!  Thanks for the work on this!

Issue #721, opened 2024-05-07
Bug: Income transactions summed under Expenses in summary


jakubkottnauer:
**The bug**
Creating a new Income transaction results in "Expenses" section in the summary increasing. Likewise when an Expense transaction is created, "Income" summary increases.

**To Reproduce**
1. Create a new Income transaction (for example in a Checking account)
2. Go to /transactions
3. Observe that the Expenses section at the top of the page has increased by the transaction amount

**Expected behavior**
Income transactions should be summed under Incomes (similarly for expenses).

**Screenshots**
Income of $100 in a checking account shown as a $100 total expense.
<img width="967" alt="Screenshot 2024-05-07 at 21 27 50" src="https://github.com/maybe-finance/maybe/assets/113784/025c2e6a-0fe3-45fb-a3b7-43728fc20f91">

PR #736, opened 2024-05-10
fix: Income summed under Expense

Harry-kp:
## Description
- Fixes https://github.com/maybe-finance/maybe/issues/721
- Adjusted the transaction controller to ensure correct handling of income and expense amounts.
  - Incomes are now processed with positive amounts, aligning with the `inflows` scope.
  - Expenses are now processed with negative amounts, ensuring consistency.
  
https://github.com/maybe-finance/maybe/assets/55315065/f0e60d03-7934-476c-a980-f647def72180

### DIFF
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -141,9 +141,9 @@ class TransactionsController < ApplicationController
 
     def amount
       if nature.income?
-        transaction_params[:amount].to_d * -1
-      else
         transaction_params[:amount].to_d
+      else
+        transaction_params[:amount].to_d * -1
       end
     end
 


Harry-kp:
@zachgoll The tests are written in the way that it expects income amount to be negative. Was this intentional?. If we are going to fix this issue, what should we going to modify the scope `inflows` or the amount.

zachgoll:
@Harry-kp yes, the convention that we're using is that any _inflow_ to an account has a negative amount, while any _outflow_ has a positive amount.  This affects the account balance differently based on the `classification` of the account.  You can find some examples of this [described here](https://github.com/maybe-finance/maybe/wiki/vision#signage-of-money).

Harry-kp:
@zachgoll  I am slightly confused between `With this approach, a positive value is an "inflow" and a negative value is an "outflow".` and as you described above ` the convention that we're using is that any inflow to an account has a negative amount, while any outflow has a positive amount. `
Also, Can you make me understand what happens in these scenarios
1. Say, we added 100 USD as income in a Account whose classification is `Asset`. What value should be stored in the amount column of transaction.
2. Say, we added 100 USD as income in a Account whose classification is `Liability`. What value should be stored in the amount column of the transaction.
2. How are you mapping income and expense to inflow or outflow? Are these terms also dependent on the classification of account. Like, it differs when account is asset to when account is liability.

zachgoll:
@Harry-kp that's my bad, at second glance, looks like the Wiki was not up-to-date with the convention we're using in the app. 

I've re-written it with a bit more clarity, let me know if that helps!

https://github.com/maybe-finance/maybe/wiki/Vision#signage-of-money

And to your question about `classification`, yes, we're using it to determine how to sum account balances.  Here's one example in the sync process:

https://github.com/maybe-finance/maybe/blob/61fae9683259861d1deeda190cc2e24075933d21/app/models/account/balance/calculator.rb#L108

Harry-kp:
@zachgoll I found the updated wiki much clearer, thank you! I just want to clarify something: When we talk about income, is it the same as inflow? And similarly, is expense the same as outflow, or does it depend on how the accounts are classified?

zachgoll:
> @zachgoll I found the updated wiki much clearer, thank you! I just want to clarify something: When we talk about income, is it the same as inflow? And similarly, is expense the same as outflow, or does it depend on how the accounts are classified?

In theory, income and expense should be a subset of inflow and outflow.  Right now, since we don't have the concept of transaction types yet, we're approximating anything with a negative amount as "income" and anything with a positive amount as "expense".  This is an "interim" solution at best, but we don't have a much better one until those transaction types are introduced.

zachgoll:
Covered by #759

PR #741, opened 2024-05-12
Fix password resets

duyle-py:
- Fix the incorrect path when submitting the edit password form.
- Redirect to the invalid token page when accessing the edit path.


zachgoll:
Looks good, thanks for the fix!

PR #770, opened 2024-05-20
Add migration to make all existing users admins

jakubkottnauer:
Updates the role of all existing users to `admin` to make older users match the change introduced here https://github.com/maybe-finance/maybe/commit/daf7ff8ef4da8718a1038f4c679539f2305195f2

As far as I know at this moment each family can only have one user, so running this should be safe.


zachgoll:
@jakubkottnauer would you mind removing the unrelated schema change for the virtual column?  I think this may be happening due to your local Postgres version.  Are you running v16+?

jakubkottnauer:
Ah you're right, I'm using 15.6, will upgrade. I assumed the change was due to the recent rails version change. Fixing!

jakubkottnauer:
@zachgoll updated

PR #789, opened 2024-05-22
Ignore empty categories while importing

jakubkottnauer:
- Empty transaction categories in the imported CSV are now ignored
- Added tests for this and a bunch of other fixed we've merged recently
- Removed `cursor-pointer` class from the badge itself (the class is already provided by the _menu partial where we actually want the cursor to change)

Before:
Screenshot 2024-05-22 at 07 44 27

After:
Screenshot 2024-05-22 at 07 43 57

### DIFF
--- a/test/fixtures/imports.yml
+++ b/test/fixtures/imports.yml
@@ -14,6 +14,18 @@ loaded_import:
     2024-01-02,Amazon stuff,Shopping,200
   created_at: <%= 2.days.ago %>
 
+loaded_import_with_missing_data:
+  account: checking
+  raw_csv_str: |
+    date,name,category,amount
+    2024-01-01,,Food,20
+    2024-01-02,Amazon stuff,,200
+  normalized_csv_str: |
+    date,name,category,amount
+    2024-01-01,,Food,20
+    2024-01-02,Amazon stuff,,200
+  created_at: <%= 2.days.ago %>
+
 completed_import:
   account: checking
   column_mappings:
--- a/test/models/import_test.rb
+++ b/test/models/import_test.rb
@@ -6,6 +6,7 @@ class ImportTest < ActiveSupport::TestCase
   setup do
     @empty_import = imports(:empty_import)
     @loaded_import = imports(:loaded_import)
+    @loaded_import_with_missing_data = imports(:loaded_import_with_missing_data)
   end
 
   test "raw csv input must conform to csv spec" do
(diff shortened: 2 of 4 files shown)


test/models/import_test.rb
```diff
@@ -39,15 +40,31 @@ class ImportTest < ActiveSupport::TestCase
...
 
     assert @loaded_import.complete?
   end
 
+  test "publishes a valid import with missing data" do
+    assert_difference("Transaction::Category.count", 1) do
+      assert_difference "Transaction.count", 2 do
+        @loaded_import_with_missing_data.publish
+      end
+    end
```
zachgoll, line 59: Tiny improvement, believe you can [combine these](https://edgeapi.rubyonrails.org/classes/ActiveSupport/Testing/Assertions.html#method-i-assert_difference) into a single assertion: 

```rb
assert_difference ->{ Transaction::Category.count } => 1, ->{ Transaction.count } => 2 do
  @loaded_import_with_missing_data.publish
end
```

test/fixtures/imports.yml
```diff
@@ -14,6 +14,18 @@ loaded_import:
...
+  raw_csv_str: |
+    date,name,category,amount
+    2024-01-01,,Food,20
+    2024-01-02,Amazon stuff,,200
+  normalized_csv_str: |
+    date,name,category,amount
+    2024-01-01,,Food,20
+    2024-01-02,Amazon stuff,,200
+  created_at: <%= 2.days.ago %>
+
```
zachgoll, line 28: Small nitpick, but I was generally keeping these fixtures as the "base cases" and then using that `import_test_helper.rb` (or directly in the test block if the data is not shared) to provide variation strings for tests.  This is mostly to stay in line with [the guide](https://edgeguides.rubyonrails.org/testing.html#what-are-fixtures-questionmark):

 > Fixtures are not designed to create every object that your tests need, and are best managed when only used for default data that can be applied to the common case.

We can then build the variation object directly in the test.  My _guess_ is that we'll be adding several more of these as we add complexity to this flow.

zachgoll (approved):
Looks good!  Left 2 small stylistic improvements to consider, but logic makes sense to me.

test/models/import_test.rb
jakubkottnauer: Cool, didn't know that, fixed!

test/fixtures/imports.yml
jakubkottnauer: Makes sense, fixed 😄

PR #791, opened 2024-05-22
Fix duplicate category creation on import

zachgoll:
Caches categories while generating `ActiveRecord` objects during import.


zachgoll:
@jakubkottnauer does this fix make sense to you based on the original issue?

jakubkottnauer:
@zachgoll looks good to me!

PR #792, opened 2024-05-22
Create tagging system

zachgoll:
## Overview

This PR adds a general purpose tagging system to the app via the `taggable` association.  To start, users can add tags to transactions, but will be able to tag other entities like `Account` in the future.

- [x] Add tags model
- [x] Add tags row to CSV imports
- [ ] Add tag dropdown to transaction detail drawer
- [ ] Add tag management view in settings

### Note on code duplication

A lot of the code here is duplicative to the categories and tags views that have already been created.  This is to avoid early abstractions as we'll likely need a larger refactor of these once we get all the entities established in the app.

### DIFF
--- a/app/models/family.rb
+++ b/app/models/family.rb
@@ -1,5 +1,6 @@
 class Family < ApplicationRecord
   has_many :users, dependent: :destroy
+  has_many :tags, dependent: :destroy
   has_many :accounts, dependent: :destroy
   has_many :transactions, through: :accounts
   has_many :imports, through: :accounts
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -111,16 +111,24 @@ class Import < ApplicationRecord
     def generate_transactions
       transactions = []
       category_cache = {}
+      tag_cache = {}
 
       csv.table.each do |row|
-        category_name = row["category"]
+        category_name = row["category"].presence
+        tag_strings = row["tags"].presence&.split("|") || []
+        tags = []
 
-        category = category_cache[category_name] ||= account.family.transaction_categories.find_or_initialize_by(name: category_name) if row["category"].present?
+        tag_strings.each do |tag_string|
+          tags << tag_cache[tag_string] ||= account.family.tags.find_or_initialize_by(name: tag_string)
+        end
+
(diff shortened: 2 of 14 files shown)


flacnut:
Thanks for adding tags! It's something I make heavy use of.

It looks as though the CSV separates tags with `|`, which means we will allow tags to contain spaces. This is ideal.

I'm not familiar with ActiveRecord in Ruby, but using a join table like this may have very poor performance scaling. For example, to show a table listing 100 most recent transactions and their respective tags, each transaction may require a join to materialize the tags.

In other ORM's I've used, this model first fetches 100 transaction rows, then in code fetches the tags individually for each of the 100 transactions.

flacnut:
Can tags be an auto-complete tokenized input like this?
https://primer.style/components/text-input-with-tokens

Having to add them one at a time with a drop down is very slow.

zachgoll:
@flacnut no problem, I think a lot of users will find this valuable for organizing entities across the system!  And yes, the `|` will be the delimiter for adding multiple tags to a transaction during an import for the reason you mention.

In regards to your performance concern, Rails allows for [eager loading of associations](https://edgeguides.rubyonrails.org/active_record_querying.html#eager-loading-associations) like this to avoid that N + 1 query problem.  While it could still get expensive to load a huge list of transactions, there are quite a few options we have to optimize where needed (caching, eager loading, lazy loaded turbo frames)

I think the benefit we get with the referential integrity of the join table is the user can then manage their tags in settings, assign colors, we get ActiveRecord validations, and we can easily do things like `replace_and_destroy!` as implemented here:

https://github.com/maybe-finance/maybe/blob/c0a2857d40a5e3e734d92105e7feea63d8ddfd78/app/models/tag.rb#L14-L24

zachgoll:
> Can tags be an auto-complete tokenized input like this? https://primer.style/components/text-input-with-tokens
> 
> Having to add them one at a time with a drop down is very slow.

Yes, those tokenized inputs will be the long term design that we'll use for this.  The solution here in this PR is more of an interim step to get this feature in quickly.  That could be a great follow-up issue I think as we will need to make some tweaks to our Stimulus dropdown controller to make it work.

PR #794, opened 2024-05-22
Fix foreign account sync crash

jakubkottnauer:
Syncing foreign currency accounts currently crashes as soon as a new exchange rate needs to be fetched.

`find_rate_or_fetch` already saves the fetched rate to the database (https://github.com/maybe-finance/maybe/blob/main/app/models/exchange_rate.rb#L15) so this PR removes the duplicate save from the Syncable module

### DIFF
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -8,12 +8,6 @@ module Account::Syncable
   def sync(start_date = nil)
     update!(status: "syncing")
 
-    if self.foreign_currency? && !ExchangeRate.exchange_rates_provider.initialized?
-      logger.error("Cancelling sync of foreign account #{id}: No exchange rate provider ready")
-      update!(status: "error")
-      return
-    end
-
     sync_exchange_rates
 
     calc_start_date = start_date - 1.day if start_date.present? && self.balance_on(start_date - 1.day).present?
@@ -24,14 +18,13 @@ module Account::Syncable
     self.balances.where("date < ?", effective_start_date).delete_all
     new_balance = calculator.daily_balances.select { |b| b[:currency] == self.currency }.last[:balance]
 
-    update!(status: "ok", last_sync_date: Date.today, balance: new_balance)
+    update!(status: "ok", last_sync_date: Date.today, balance: new_balance, sync_errors: calculator.errors, sync_warnings: calculator.warnings)
   rescue => e
     update!(status: "error")
     logger.error("Failed to sync account #{id}: #{e.message}")
   end
 
   def can_sync?
-    return false if self.foreign_currency? && !ExchangeRate.exchange_rates_provider.initialized?
     # Skip account sync if account is not active or the sync process is already running
     return false unless is_active
     return false if syncing?
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -20,7 +20,7 @@ module Account::Syncable
 
     update!(status: "ok", last_sync_date: Date.today, balance: new_balance, sync_errors: calculator.errors, sync_warnings: calculator.warnings)
   rescue => e
-    update!(status: "error")
+    update!(status: "error", sync_errors: [ :sync_message_unknown_error ])
     logger.error("Failed to sync account #{id}: #{e.message}")
   end
 
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -8,6 +8,12 @@ module Account::Syncable
   def sync(start_date = nil)
     update!(status: "syncing")
 
+    if self.foreign_currency? && !ExchangeRate.exchange_rates_provider.initialized?
+      logger.error("Cancelling sync of foreign account #{id}: No exchange rate provider ready")
+      update!(status: "error")
+      return
+    end
+
     sync_exchange_rates
 
     calc_start_date = start_date - 1.day if start_date.present? && self.balance_on(start_date - 1.day).present?
@@ -21,10 +27,11 @@ module Account::Syncable
     update!(status: "ok", last_sync_date: Date.today, balance: new_balance)
   rescue => e
     update!(status: "error")
-    Rails.logger.error("Failed to sync account #{id}: #{e.message}")
+    logger.error("Failed to sync account #{id}: #{e.message}")
   end
 
   def can_sync?
+    return false if self.foreign_currency? && !ExchangeRate.exchange_rates_provider.initialized?
     # Skip account sync if account is not active or the sync process is already running
     return false unless is_active
     return false if syncing?
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -77,8 +77,7 @@ module Account::Syncable
       next if existing_rates_set.include?([ rc_from, rc_to, rc_date.to_s ])
 
       logger.info "Fetching exchange rate from provider for account #{self.name}: #{self.id} (#{rc_from} to #{rc_to} on #{rc_date})"
-      rate = ExchangeRate.find_rate_or_fetch from: rc_from, to: rc_to, date: rc_date
-      ExchangeRate.create! base_currency: rc_from, converted_currency: rc_to, date: rc_date, rate: rate if rate
+      ExchangeRate.find_rate_or_fetch from: rc_from, to: rc_to, date: rc_date
     end
 
     nil
--- a/app/models/exchange_rate.rb
+++ b/app/models/exchange_rate.rb
@@ -12,7 +12,7 @@ class ExchangeRate < ApplicationRecord
     end
 
     def find_rate_or_fetch(from:, to:, date:)
-      find_rate(from:, to:, date:) || fetch_rate_from_provider(from:, to:, date:).tap(&:save!)
+      find_rate(from:, to:, date:) || fetch_rate_from_provider(from:, to:, date:)&.tap(&:save!)
     end
 
     def get_rate_series(from, to, date_range)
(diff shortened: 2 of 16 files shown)


zachgoll:
@jakubkottnauer I could be remembering this incorrectly, but wouldn't `find_rate_or_fetch` raise an error if the user has not configured an exchange rates provider?

I'm thinking we'll want to make sure that no errors are raised if the user hasn't yet configured the provider and go with the "warning" approach we were talking about [here](https://github.com/maybe-finance/maybe/issues/724#issuecomment-2100489937).

jakubkottnauer:
@zachgoll I went ahead and implemented a UI alert that shows up if the `SYNTH_API_KEY` env var hasn't been provided. I have added a `Provider::Synth.initialized?` and use it to show the alert:

Screenshot 2024-05-23 at 20 49 59

I have also added the use of `Provider::Synth.initialized?` into the `sync` call itself. This took me down a fun rabbit hole of failing tests, because in tests `initialized?` would always return false. I have refactored the Synth provider to be properly configurable using the pattern described in [this article](https://thoughtbot.com/blog/mygem-configure-block), instead of depending on the env var directly. What do you think?

app/models/account/syncable.rb
```diff
@@ -8,6 +8,12 @@ def sync_later(start_date = nil)
   def sync(start_date = nil)
     update!(status: "syncing")
 
+    if self.foreign_currency? && !ExchangeRate.exchange_rates_provider.initialized?
+      logger.error("Cancelling sync of foreign account #{id}: No exchange rate provider ready")
+      update!(status: "error")
+      return
+    end
```
zachgoll, line 15: I like the idea of canceling a sync based on a condition here, but I think a `foreign_currency?` account should still be able to sync regardless of whether the exchange rates are available.

The reason being—as long as all the transactions/valuations are in the same currency, we can still successfully build a historical balance graph in that currency.

If we updated this to `self.multi_currency?` it makes more sense to me as we would require exchange rates to build even the native historical graph.

The "rollup" is the point of abstraction where the exchange rates missing cause an issue I think.

config/initializers/synth.rb
```diff
@@ -0,0 +1,5 @@
+Rails.application.config.to_prepare do
+  Provider::Synth.configure do |config|
+    config.api_key = ENV["SYNTH_API_KEY"]
+  end
+end
```
zachgoll, line 5: The issue I think we'll run into here is for self-hosters.  They'll need to be able to configure their Synth API key directly in the UI via the [settings model](https://github.com/maybe-finance/maybe/blob/main/app/models/setting.rb), which we haven't configured quite yet, but will need to soon.

You had mentioned that a big reason for this refactor is due to `initialized?` always returning false during the tests.  I'm thinking that may be more of a symptom of tightly-coupled dependencies rather than a reason to change the class itself.  I'm not sure which tests were failing on you, but I think we should try to address it elsewhere so we can keep the config for `Synth` available at runtime.  Let me know which tests are causing problems and I'd be happy to throw around some ideas.

Something that may help out is the `with_env_override` helper, which allows us to toggle environment variables on and off during tests for special cases where we need to test "enabled/disabled" behaviors:

https://github.com/maybe-finance/maybe/blob/457247da8e40254c469a1133fefbba4ffc5013b5/test/test_helper.rb#L57-L59

Here's an example usage:

https://github.com/maybe-finance/maybe/blob/457247da8e40254c469a1133fefbba4ffc5013b5/test/controllers/registrations_controller_test.rb#L28

app/models/account/syncable.rb
jakubkottnauer: Depending on what exchange rates are available during a sync, there is a chance that an error is raised in the calculator here https://github.com/maybe-finance/maybe/blob/main/app/models/account/balance/calculator.rb#L54
which is why I went with short-circuiting the entire sync process. 

If we're ok with not raising errors here and just generating invalid balances (together with showing the UI messages) then I can change this

config/initializers/synth.rb
jakubkottnauer: It was specificaly the `sync` tests were failing because of the new early return we discuss in the other thread. If we were to remove the early return and instead proceed with the sync generating possibly invalid balances, then that would also remove the need for this refactor

app/models/account/syncable.rb
zachgoll: Yeah this is a tricky one.  To me this feels like the calculator logic I wrote is more at fault here than anything.

Ultimately, there are 2 stages of currency conversion going on in this calculator:

1. Normalizing `multi_currency?` balances - if this fails, then we can't show anything for the account 
2. Converting an entire series of `foreign_currency?` balances - if this fails, we can't rollup the account, but we can still show a history graph in its native currency

We could either:

1. Loop through balances and for balances where there is no exchange rate, we don't generate it at all (leaving balance gaps)
2. Pre-fetch exchange rates for the known balances and if any exchange fails, we abort the entire conversion

The "all or nothing" approach in #2 seems a bit simpler here and in order to make that happen I think this calculator is what needs the refactor to decouple the balance logic from the exchange rate logic.  The two are intermingled right now.  For example:

https://github.com/maybe-finance/maybe/blob/6e59fdb369d18dcd1163e9088f272a716ace6acb/app/models/account/balance/calculator.rb#L76-L82

config/initializers/synth.rb
zachgoll: Yes this does allow for configuration at runtime, but it would be more of a "re-configuration" (initializes, then user changes the api key in their settings), which feels like an extra layer of abstraction (global state) that is not necessary.

Here is the [original data providers PR](https://github.com/maybe-finance/maybe/pull/574) where I had started down this "global state" path and shows how we ended up with the current design which is more tightly integrated with the domain models.

I think rather than focusing on configuration and the `initialized?` method that checks whether an API key is present (which still doesn't guarantee rates are present), we need to dig into our calculator class and refactor the logic to make these decisions based on the presence/absence of the exchange rates as I mentioned in the other comment.

In other words, if we're syncing foreign currency account XYZ, we need to figure out which exchange rates are needed, check if we have all of them available (regardless of providers as they could be in our database), and then proceed/abort the sync depending on that response.

zachgoll: And then for the UI warnings/errors, we can check on an account-by-account basis by populating these during the sync process (these are currently not in use, but I had added as placeholders):

https://github.com/maybe-finance/maybe/blob/6e59fdb369d18dcd1163e9088f272a716ace6acb/db/schema.rb#L93-L94

That way, we can simply say:

```erb
<% if @account.sync_errors.present? %>
  <%= render partial: "accounts/some-partial", locals: { errors: @account.sync_errors } %>
<% end %>
```

jakubkottnauer:
@zachgoll Ready for another round of review. I tried to keep the changes to a bare minimum (considering that my original aim with this PR was just to unblock myself from being able to sync foreign currency accounts :)) so I have left `sync_warnings` and `sync_errors` the way they were setup, just fixed a few related bugs I found along the way.

app/models/exchange_rate.rb
```diff
@@ -12,11 +12,11 @@ def find_rate(from:, to:, date:)
     end
 
     def find_rate_or_fetch(from:, to:, date:)
-      find_rate(from:, to:, date:) || fetch_rate_from_provider(from:, to:, date:).tap(&:save!)
+      find_rate(from:, to:, date:) || fetch_rate_from_provider(from:, to:, date:)&.tap(&:save!)
     end
 
-    def get_rate_series(from, to, date_range)
-      where(base_currency: from, converted_currency: to, date: date_range).order(:date)
+    def get_rates(from, to, dates)
```
jakubkottnauer, line 18: Renamed the method as it can be used not just for a continuous series of dates, but also for an arbitrary array of dates

zachgoll (approved):
@jakubkottnauer nice updates here, I think this is a good improvement on what we had.

A few notes that could be good to keep in mind moving forward:

- In the future, we'll probably have to introduce some sort of UI that warns the user that they will be using API credits for a sync process (not needed yet)
- As I'm looking through the errors vs. warnings, my _guess_ is that we'll probably end up with a single `errors` array with varying severities.  I'm thinking that it might be adding a bit of ambiguity to our sync process trying to decide whether something is an "error" or "warning".   Not something we need to update here, but just a thought moving forward.

app/views/shared/_alert.html.erb
```diff
@@ -0,0 +1,12 @@
+<%# locals: (type: "error", content: "") -%>
+<% color = type == "error" ? "red" : "yellow" %>
+<%= content_tag :div,
+  class: "flex justify-between bg-#{color}-50 rounded-xl p-3",
```
zachgoll, line 4: Since Tailwind does static analysis of the classes, `bg-#{color}-50` won't always work properly:

https://tailwindcss.com/docs/content-configuration#dynamic-class-names

app/views/shared/_alert.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  data: {controller: "element-removal" },
+  role: type == "error" ? "alert" : "status" do %>
+  <div class="flex gap-3 text-<%= color %>-500 items-center">
+    <%= lucide_icon("info", class: "w-5 h-5 shrink-0") %>
+    <p class="text-sm"><%= content %></p>
+  </div>
+  <%= content_tag :a, lucide_icon("x", class: "w-5 h-5 shrink-0 text-#{color}-500"), data: { action: "click->element-removal#remove" }, class:"flex gap-1 font-medium items-center text-gray-900 px-3 py-1.5 rounded-lg cursor-pointer" %>
```
zachgoll, line 11: Same as above for `text-#{color}-500`

app/models/exchange_rate.rb
jakubkottnauer: Great point, I always forget about this 🙇

app/views/shared/_alert.html.erb
jakubkottnauer: Great point, I always forget about this 🙇

PR #797, opened 2024-05-23
Adding one-click deploy button for RepoCloud.io to README.md

cosark:
Adding button enabling community access to deploy template for self-hosting on RepoCloud.io with one click.

### DIFF
--- a/docs/hosting/one-click-deploy.md
+++ b/docs/hosting/one-click-deploy.md
@@ -2,6 +2,14 @@
 
 Below are our "one-click deploy" options for running Maybe in the cloud:
 
+## RepoCloud.io
+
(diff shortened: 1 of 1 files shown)


zachgoll:
@cosark unfortunately we won't be able to accept this without doing quite a bit of additional work to enable self-hosters to manage upgrades for their self hosted instance.  [Here's the original PR](https://github.com/maybe-finance/maybe/pull/612) for Render that has some examples of the work required to add new one-click deploys.

We intend to provide official support for more in the future, but in order to prioritize feature work we'll be sticking with Docker and Render for a while.

Appreciate the contribution though.  If you'd like this as a feature you can open a [feature request here.](https://github.com/maybe-finance/maybe/discussions/categories/feature-requests)

cosark:
Okay no problem. There's no documentation required. It's just one-click deploy. 

There are already many happy users on the platform who didn't have to pay 3-4 times more for Render, and thought you might want to give the option to your community.

PR #800, opened 2024-05-23
Add tag preview when importing and fix empty category bug

jakubkottnauer:
- Fixed reintroduced bug when importing transactions with empty categories (see screenshot below)
- I have added tags to the import preview to verify visually that tags don't suffer from the same issue

Before category fix:

Screenshot 2024-05-23 at 21 53 14

After category fix:

Screenshot 2024-05-23 at 21 52 56

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -122,7 +122,7 @@ class Import < ApplicationRecord
           tags << tag_cache[tag_string] ||= account.family.tags.find_or_initialize_by(name: tag_string)
         end
 
-        category = category_cache[category_name] ||= account.family.transaction_categories.find_or_initialize_by(name: category_name)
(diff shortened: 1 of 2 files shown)


app/models/import.rb
```diff
@@ -122,7 +122,7 @@ def generate_transactions
           tags << tag_cache[tag_string] ||= account.family.tags.find_or_initialize_by(name: tag_string)
         end
 
-        category = category_cache[category_name] ||= account.family.transaction_categories.find_or_initialize_by(name: category_name)
+        category = category_cache[category_name] ||= account.family.transaction_categories.find_or_initialize_by(name: category_name) if category_name.present?
```
zachgoll, line 125: Ahh my bad, I think that was a regression on my part

zachgoll (approved):
Looks good!

PR #819, opened 2024-05-30
Reusable CI workflow for GH actions

zachgoll:
Small consolidation of duplicate workflows so we're running the same CI process for PRs and merges to main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
 on:
-  pull_request:
+  workflow_call:
 
 jobs:
   scan_ruby:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -82,16 +86,19 @@ jobs:
...
           bin/rails db:schema:load
-          bin/rails test
           bin/rails db:seed
 
+      - name: Unit and integration tests
+        run: bin/rails test
+
+      - name: System tests
+        run: DISABLE_PARALLELIZATION=true bin/rails test:system
+        continue-on-error: true # TODO: Eventually we'll enforce for PRs
```
zachgoll, line 100: As explained in #817, the intention here is to allow for us to start running system tests during the CI process without enforcing them yet.  We'll eventually remove this and require system tests once the app reaches a more stable point.

PR #836, opened 2024-06-04
Account state machine

timlawrenz:
Hi,

I suggest using state machines for columns that should have controlled workflows.
This PR also tries to ensure that the decision logic (whether an account can be synced) stays close to the model. A family should not know too much about whether an account is syncable.

Please let me know what you think.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -38,7 +38,7 @@ class AccountsController < ApplicationController
   def update
     if @account.update(account_params.except(:accountable_type))
 
-      @account.sync_later if account_params[:is_active] == "1" && @account.can_sync?
+      @account.sync_later if @account.may_start_sync?
 
       respond_to do |format|
         format.html { redirect_to accounts_path, notice: t(".success") }
@@ -74,7 +74,7 @@ class AccountsController < ApplicationController
   end
 
   def sync
-    if @account.can_sync?
+    if @account.may_start_sync?
       @account.sync_later
       respond_to do |format|
         format.html { redirect_to account_path(@account), notice: t(".success") }
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -38,7 +38,7 @@ class AccountsController < ApplicationController
   def update
     if @account.update(account_params.except(:accountable_type))
 
-      @account.sync_later if @account.may_start_sync?
+      @account.sync_later
 
       respond_to do |format|
         format.html { redirect_to accounts_path, notice: t(".success") }
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -74,7 +74,7 @@ class AccountsController < ApplicationController
   end
 
   def sync
-    if @account.may_start_sync?
+    if @account.may_sync?
       @account.sync_later
       respond_to do |format|
         format.html { redirect_to account_path(@account), notice: t(".success") }
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -13,8 +13,6 @@ class Account < ApplicationRecord
 
   monetize :balance
 
-  enum :status, { ok: "ok", syncing: "syncing", error: "error" }, validate: true
-
   scope :active, -> { where(is_active: true) }
   scope :assets, -> { where(classification: "asset") }
   scope :liabilities, -> { where(classification: "liability") }
@@ -42,11 +40,6 @@ class Account < ApplicationRecord
     [ { name: "Manual accounts", accounts: all.order(balance: :desc).group_by(&:accountable_type) } ]
   end
 
-  def self.some_syncing?
-    exists?(status: "syncing")
-  end
-
-
   def series(period: Period.all, currency: self.currency)
     balance_series = balances.in_period(period).where(currency: Money::Currency.new(currency).iso_code)
 
@@ -69,7 +62,7 @@ class Account < ApplicationRecord
       types.each do |type|
         group = grouped_accounts[classification.to_sym].add_child_group(type, currency)
         self.where(accountable_type: type).each do |account|
-          value_node = group.add_value_node(
+          group.add_value_node(
             account,
             account.balance_money.exchange_to(currency) || Money.new(0, currency),
             account.series(period: period, currency: currency)
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -58,7 +58,6 @@ module Account::Syncable
   def can_sync?
     # Skip account sync if account is not active or the sync process is already running
     return false unless is_active
-    return false if syncing?
     # If last_sync_date is blank (i.e. the account has never been synced before) allow syncing
     return true if last_sync_date.blank?
 
--- a/app/models/account/syncable.rb
+++ b/app/models/account/syncable.rb
@@ -7,7 +7,7 @@ module Account::Syncable
     enum :status, { ok: "ok", syncing: "syncing", error: "error" }, validate: true
 
     def some_syncing?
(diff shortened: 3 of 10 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -38,7 +38,7 @@ def edit
   def update
     if @account.update(account_params.except(:accountable_type))
 
-      @account.sync_later if account_params[:is_active] == "1" && @account.can_sync?
+      @account.sync_later
```
timlawrenz, line 41: The controller shouldn't know too much about the model's internal behavior.

app/models/account.rb
```diff
@@ -13,8 +13,6 @@ class Account < ApplicationRecord
 
   monetize :balance
 
-  enum :status, { ok: "ok", syncing: "syncing", error: "error" }, validate: true
-
```
timlawrenz, line 17: moved to `syncable`

app/models/account.rb
```diff
@@ -42,11 +40,6 @@ def self.by_provider
     [ { name: "Manual accounts", accounts: all.order(balance: :desc).group_by(&:accountable_type) } ]
   end
 
-  def self.some_syncing?
-    exists?(status: "syncing")
-  end
```
timlawrenz, line 47: moved to `syncable`

app/models/account.rb
```diff
@@ -69,7 +62,7 @@ def self.by_group(period: Period.all, currency: Money.default_currency)
       types.each do |type|
         group = grouped_accounts[classification.to_sym].add_child_group(type, currency)
         self.where(accountable_type: type).each do |account|
-          value_node = group.add_value_node(
+          group.add_value_node(
```
timlawrenz, line 65: Cleanup: the assigned `value_node` was never used.

app/models/account/syncable.rb
```diff
@@ -1,13 +1,42 @@
 module Account::Syncable
   extend ActiveSupport::Concern
 
+  included do
+    include AASM
+
+    enum :status, { ok: "ok", syncing: "syncing", error: "error" }, validate: true
```
timlawrenz, line 7: Coming from the model. It might be sensible to name this to `sync_state`. An `Account` may have other states (maybe related to the financial organization, such as `closed`, `unavailable`, etc.).

app/models/family.rb
```diff
@@ -104,6 +104,6 @@ def liabilities
   end
 
   def sync_accounts
-    accounts.each { |account| account.sync_later if account.can_sync? }
+    accounts.each { |account| account.sync_later }
```
timlawrenz, line 107: The `Family` shouldn't know too much about internal states of `Accounts`.

app/models/account/syncable.rb
```diff
@@ -18,16 +47,17 @@ def sync(start_date = nil)
...
-    update!(status: "error", sync_errors: [ :sync_message_unknown_error ])
+    sync_fails
+    update!(sync_errors: [ :sync_message_unknown_error ])
     logger.error("Failed to sync account #{id}: #{e.message}")
   end
 
   def can_sync?
     # Skip account sync if account is not active or the sync process is already running
     return false unless is_active
-    return false if syncing?
```
timlawrenz, line 30: This is now tested earlier through the state machine.

zachgoll (commented):
Thanks for this contribution!  Overall, I agree with the consolidation you've done to move everything into the model and within the `Syncable` module.

Before we get this merged, I just had a few questions below:

**Naming**

I agree that we should probably consider renaming a few things to make it clear that these state transitions are in the context of "Account Syncing".

What do you think of something like this?

(would require us to do a DB migration to change `:status` -> `:sync_state` and remove the enum)

```rb
aasm column: :sync_state do
  # implementation
end
```

**Possible states**

The account sync will soon have several steps within it and I'm wondering your thoughts on the best way to incorporate a state machine for this.  For example, a "Sync" may include all of the following steps:

1. Sync data provider transactions
2. Sync data provider exchange rates
3. Sync data provider account data
4. Sync data provider stock securities prices
5. Sync balances (takes all 3rd party data and uses it to calculate balances)

Essentially, it's an ETL process that will be similar to the [V1 implementation](https://github.com/maybe-finance/maybe-archive/blob/d18d5bab6e5c20124158b07370737e1e5fa9cbaf/libs/server/features/src/account/account.processor.ts#L19-L32).

While I suppose we could keep these all combined into an `event :sync, after: :start_sync`, it may be nicer if we're able to see the exact step that failed in the DB.

My thought is that rather than including an `:error` state, we'd just have:

```rb
aasm column: :sync_state do 
  state :idle, initial: true
  state :syncing_exchange_rates
  state :syncing_balances

  event :sync_exchange_rates do
    after { |start_date| start_exchange_rate_sync(start_date) }
    error { |e| append_sync_error(e) }
    transitions from: :idle, to: :syncing_exchange_rates
  end

  event :sync_balances do
    after { |start_date| start_balance_sync(start_date) }
    error { |e| append_sync_error(e) }
    transitions from: :syncing_exchange_rates, to: :syncing_balances
  end

  event :complete_sync do 
     transitions from: :syncing_balances, to: :idle
  end

  def start_exchange_rate_sync(start_date = nil)
  end

  def start_balance_sync(start_date = nil)
  end

  def complete_sync 
  end

  def append_sync_error(e)
    # Updates the `sync_errors` column
  end
end
```

There may be a better way to represent this.  Curious to hear your thoughts.

timlawrenz:
Hi @zachgoll,

Thank you for your feedback!

I like your approach, which is to step through the individual stages and label them as states.

> While I suppose we could keep these all combined into an event :sync, after: :start_sync, it may be nicer if we're able to see the exact step that failed in the DB.

To improve observability, it's not unreasonable to create a model `Sync` (or `ProviderSync`, or similar) that logs each individual sync step, together with account, and possible error message. If necessary we can clean that out after 30.days or so, but could provide a log with latest sync issues.

zachgoll:
@timlawrenz yeah I'd agree with that.  If we created an `Account::Sync` model, I'd assume `sync_state` (and state machine) stay on `Account` and then we'd possibly lift and shift some (or all) of the `Syncable` logic to the new model?

My guess is that 3rd party provider data could be stored directly on the `Account::Sync` model as JSONB columns, so we could do something like:

```rb
sync = Account::Sync.new(start_date: 10.days.ago)

sync.fetch_exchange_rates # stores in a JSONB column 
sync.fetch_other_provider_data
```

And as you mentioned, future improvements may include things like purging old syncs or maybe even setting up associations (rather than JSONB) as we become more familiar with exactly what data we need.

zachgoll:
@timlawrenz I meant the actual results that were fetched from the provider.  Each sync will follow some sort of ETL pattern, and it may not be possible to "transform + load" the data right away (i.e. we might need to combine transaction data with exchange rate data).  My line of thinking is that we should be able to perform the "extract" step in parallel with retries, which could be challenging if we don't save the raw fetched data in some sort of interim "staging" area.

timlawrenz:
@zachgoll Thanks for clarifying. I'll update this PR to

- include a sync model that
  - has a relationship with an account
  - is responsible for the actual syncing
  - stores an error message if applicable

We should create a new sync object (row) for every sync step to keep the individual results/error messages around. The account state machine orchestrates the creation of those objects. What is now in the syncable concern should be moved to the sync class.

zachgoll:
@timlawrenz yep, I like that approach.  If you'd like to introduce a sync "row", that's fine.  The one thing I think we should optimize most for right now is _clarity and simplicity_.  100% fine suffering on performance a little bit to get us to the clearest and most understandable flow possible.

timlawrenz:
@zachgoll, that is very good guidance.

PR #840, opened 2024-06-06
Add bulk selection UI controls

zachgoll:
This PR introduces a Stimulus controller to handle bulk selection of transactions (although the controller is generic to any collection of records).

https://github.com/maybe-finance/maybe/assets/16676157/b0e7d89f-a534-4245-b60c-4ca44cadb65b

Future PRs will handle the bulk edit and delete operations.

### DIFF
--- a/app/views/accounts/transactions/_transaction.html.erb
+++ b/app/views/accounts/transactions/_transaction.html.erb
@@ -1,0 +1,17 @@
+<%= turbo_frame_tag dom_id(transaction), class: "grid grid-cols-12 items-center text-gray-900 py-4 text-sm font-medium px-4" do %>
+  <div class="col-span-4">
+    <%= render "transactions/name", transaction: transaction %>
+  </div>
+
+  <div class="col-span-3">
+    <%= render "transactions/categories/badge", category: transaction.category %>
+  </div>
+
+  <%= link_to transaction.account.name,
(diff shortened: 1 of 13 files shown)


app/views/accounts/transactions/_transaction.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  </div>
+
+  <%= link_to transaction.account.name,
+              account_path(transaction.account),
+              class: ["col-span-3 hover:underline"] %>
+
+  <div class="col-span-2 ml-auto">
+    <%= render "transactions/amount", transaction: transaction %>
+  </div>
+<% end %>
```
zachgoll, line 17: As the global transactions UI is getting a lot more complex with bulk selection and inline editing, I decided to make the transaction rows in the account view read-only for the time being.

Eventually we'll probably want to mirror some of the functionality implemented in the global view here, but trying to manage this in two spots adds unnecessary complexity right now.

config/routes.rb
```diff
@@ -43,7 +43,9 @@
 
   resources :transactions do
     collection do
-      match "search" => "transactions#search", via: %i[ get post ]
```
zachgoll, line 46: stale route that I missed earlier in the transactions cleanup

Issue #842, opened 2024-06-07
Bug: pie chart on dashboard does not support multi-currency formatting 


AurelienConte:
**Where did this bug occur?**

- [x] Local development
- [x] Self hosted app (i.e. Docker)

**Describe the bug**

On the dashboard after adding my first checking account the assets graph is showing my checking account value with an "undefined" value.

**To Reproduce**

- Follow the self hosted docker installation and open the dashboard.
- Create your first account
- Add a bank account with a positive amount
- Refresh the dashboard
- You should see undefined in the upper right part of the dashboard ( Above total assets )

**Expected behavior**

I expect to show my account value but without the undefined keyword

**Screenshots / Recordings**

![image](https://github.com/maybe-finance/maybe/assets/39769262/840467d7-383f-4579-8aeb-aad680f40dec)

AurelienConte:
By the way i'm looking for some "Good first issue" to start working on the project and i think this can be a good one !

zachgoll:
@AurelienConte yeah this would be a great first issue, you're welcome to take it!

AurelienConte:
Okay ! 
Thanks i take it !

AurelienConte:
I created a PR about this #844

PR #844, opened 2024-06-07
fix: #842 - "undefined" value on dashboard assets graph

AurelienConte:
Fixing issue of ``FractionalPart`` displaying "undefined" in the pie chart element used to display assets and debts on the dashboard

EDIT : Please, do not hesitate to tell me about the testing part, i'm not familiar with ruby and tools in this ecosystem. I'm still learning and maybe i used them in a not proper way !

### DIFF
--- a/app/helpers/value_groups_helper.rb
+++ b/app/helpers/value_groups_helper.rb
@@ -1,28 +1,40 @@
 module ValueGroupsHelper
   def value_group_pie_data(value_group)
-    values = value_group.children
-      .map do |child|
-        {
-          label: to_accountable_title(Accountable.from_type(child.name)),
-          percent_of_total: child.percent_of_total.round(1).to_f,
-          value: child.sum.amount.to_f,
-          currency: child.sum.currency.iso_code,
-          currency_symbol: child.sum.currency.symbol,
-          formatted_value: format_money(child.sum.amount.to_f, precision: 2, unit: child.sum.currency.symbol, separator: ".", delimiter: ","),
-          bg_color: accountable_bg_class(child.name),
-          fill_color: accountable_fill_class(child.name)
-        }
-      end
-      .filter { |child| child[:value] > 0 }
+    values = value_group.children.map do |child|
+      {
+        label: to_accountable_title(Accountable.from_type(child.name)),
+        percent_of_total: child.percent_of_total.round(1).to_f,
+        value: child.sum.amount.to_f,
+        value_str: {
+          fractional_part: child.sum.cents_str,
+          main_part: format_money_without_symbol(child.sum.amount.to_f, precision: 0, unit: child.sum.currency.symbol)
+        },
+        currency: {
+          iso_code: child.sum.currency.iso_code,
+          symbol: child.sum.currency.symbol,
+          displayed_before_value: child.sum.currency.default_format.start_with?("%u")
+        },
+        bg_color: accountable_bg_class(child.name),
+        fill_color: accountable_fill_class(child.name)
+      }
+    end.filter { |child| child[:value] > 0 }
 
-    total = values.sum { |child| child[:value] }
-    formatted_total = format_money(total, precision: 2, unit: value_group.sum.currency.symbol, separator: ".", delimiter: ",")
+    total_value = values.sum { |child| child[:value] }
+    main_part = format_money_without_symbol(total_value, precision: 0, unit: value_group.sum.currency.symbol)
+    fractional_part = Money.new(total_value, value_group.sum.currency).cents_str
 
     {
       values: values,
-      total: total,
-      formatted_total: formatted_total,
-      currency_symbol: value_group.sum.currency.symbol
+      value: total_value,
+      value_str: {
+        main_part: main_part,
+        fractional_part: fractional_part
+      },
+      currency: {
+        iso_code: value_group.sum.currency.iso_code,
+        symbol: value_group.sum.currency.symbol,
+        displayed_before_value: value_group.sum.currency.default_format.start_with?("%u")
+      }
     }.to_json
 
   end
--- a/app/helpers/value_groups_helper.rb
+++ b/app/helpers/value_groups_helper.rb
@@ -1,41 +1,14 @@
 module ValueGroupsHelper
   def value_group_pie_data(value_group)
-    values = value_group.children.map do |child|
+    value_group.children.filter { |child| child.sum > 0 }.map do |child|
       {
         label: to_accountable_title(Accountable.from_type(child.name)),
         percent_of_total: child.percent_of_total.round(1).to_f,
         value: child.sum.amount.to_f,
-        value_str: {
-          fractional_part: child.sum.cents_str,
-          main_part: format_money_without_symbol(child.sum.amount.to_f, precision: 0, unit: child.sum.currency.symbol)
-        },
-        currency: {
-          iso_code: child.sum.currency.iso_code,
-          symbol: child.sum.currency.symbol,
-          displayed_before_value: child.sum.currency.default_format.start_with?("%u")
-        },
+        formatted_value: format_money(child.sum, precision: 0),
         bg_color: accountable_bg_class(child.name),
         fill_color: accountable_fill_class(child.name)
       }
-    end.filter { |child| child[:value] > 0 }
-
-    total_value = values.sum { |child| child[:value] }
-    main_part = format_money_without_symbol(total_value, precision: 0, unit: value_group.sum.currency.symbol)
-    fractional_part = Money.new(total_value, value_group.sum.currency).cents_str
-
-    {
-      values: values,
-      value: total_value,
-      value_str: {
-        main_part: main_part,
-        fractional_part: fractional_part
-      },
-      currency: {
-        iso_code: value_group.sum.currency.iso_code,
-        symbol: value_group.sum.currency.symbol,
-        displayed_before_value: value_group.sum.currency.default_format.start_with?("%u")
-      }
-    }.to_json
-
+    end.to_json
   end
 end
--- a/app/helpers/value_groups_helper.rb
+++ b/app/helpers/value_groups_helper.rb
@@ -7,6 +7,7 @@ module ValueGroupsHelper
           percent_of_total: child.percent_of_total.round(1).to_f,
           value: child.sum.amount.to_f,
           currency: child.sum.currency.iso_code,
+          currency_symbol: child.sum.currency.symbol,
           formatted_value: format_money(child.sum.amount.to_f, precision: 2, unit: child.sum.currency.symbol, separator: ".", delimiter: ","),
           bg_color: accountable_bg_class(child.name),
           fill_color: accountable_fill_class(child.name)
@@ -21,6 +22,7 @@ module ValueGroupsHelper
       values: values,
       total: total,
       formatted_total: formatted_total,
+      currency_symbol: value_group.sum.currency.symbol
     }.to_json
 
   end
--- a/app/helpers/value_groups_helper.rb
+++ b/app/helpers/value_groups_helper.rb
@@ -4,7 +4,6 @@ module ValueGroupsHelper
       {
         label: to_accountable_title(Accountable.from_type(child.name)),
         percent_of_total: child.percent_of_total.round(1).to_f,
-        value: child.sum.amount.to_f,
         formatted_value: format_money(child.sum, precision: 0),
         bg_color: accountable_bg_class(child.name),
         fill_color: accountable_fill_class(child.name)
--- a/app/helpers/value_groups_helper.rb
+++ b/app/helpers/value_groups_helper.rb
@@ -1,17 +1,27 @@
 module ValueGroupsHelper
   def value_group_pie_data(value_group)
-    value_group.children
+    values = value_group.children
       .map do |child|
         {
           label: to_accountable_title(Accountable.from_type(child.name)),
           percent_of_total: child.percent_of_total.round(1).to_f,
           value: child.sum.amount.to_f,
           currency: child.sum.currency.iso_code,
+          formatted_value: format_money(child.sum.amount.to_f, precision: 2, unit: child.sum.currency.symbol, separator: ".", delimiter: ","),
           bg_color: accountable_bg_class(child.name),
           fill_color: accountable_fill_class(child.name)
         }
       end
       .filter { |child| child[:value] > 0 }
-      .to_json
+
+    total = values.sum { |child| child[:value] }
+    formatted_total = format_money(total, precision: 2, unit: value_group.sum.currency.symbol, separator: ".", delimiter: ",")
+
+    {
+      values: values,
+      total: total,
+      formatted_total: formatted_total,
+    }.to_json
+
   end
 end
--- a/app/javascript/controllers/pie_chart_controller.js
+++ b/app/javascript/controllers/pie_chart_controller.js
@@ -95,16 +95,6 @@ export default class extends Controller {
     `;
   }
 
-  #currencyValue(datum) {
-    return `
-      <p class="text-gray-500 -space-x-0.5">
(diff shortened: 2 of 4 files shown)


test/controllers/pie_chart_controller_test.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  private
+  def assert_amount_balance(expected_value)
+    value_element = find('div.absolute.inset-0.w-full.text-center p span.text-xl') # Fetch the pie chart text
+    entire_text = value_element.find(:xpath, '..').text # Get full text of the <p> element
+    extracted_value = entire_text.strip.gsub(/[^\d.]/, '').sub(/^0+/, '') # Removing non-digit characters except the decimal point
+    assert_equal expected_value, extracted_value, "The pie chart value is not displayed correctly"
+  end
+
+end
```
zachgoll, line 22: Great idea to add these tests!  A few ideas to make this test a bit more resilient:

- We should try and be less specific with our selectors here.  I'd go ahead and add an `id` or `data-id` on the specific element that you're looking to test so that when classes change in the future (which they likely will), this test will not break.
- I think if you want to write a system test like this, it should be in a file called `test/system/dashboard_test.rb`.  That said, since this is specific to the formatting of a monetary value which is provided by `value_groups_helper.rb`, it might be best to test this directly in `test/helpers/value_groups_helper_test.rb`

This is a pretty tricky one as it deals with the formatting of a currency.  Since we support multi-currency apps, it appears that the problem here may exist in the `pie_chart_controller.js`, which relies on the values to be in `USD`.

I think the change that we should consider making is removing or altering the `#currencyValue` method from `pie_chart_controller.js`.  This method relies on the decimal separator always being a `.` which will not apply for all currencies.  Switching this to `,` will then cause the opposite problem.  I'm not sure what the best solution here is, but my guess is that it would be easier to pass the Stimulus controller an already-formatted currency value so that we can leverage our robust collection of `Money::Currency` objects on the backend.

AurelienConte: I have made the logic in the backend to generate the value but i stick with a problem. The display is not really looking like time series chart is doing right now ( which is better ! ) 

Pie chart display : 
image

TimeSeries chart display : ( the title part )
image

I can't find a way to display in gray the symbol, i don't known if using the ``format`` parameter of ``format_money`` function i use is a good approach to fix this problem..

I'm pushing my current changes, if you have some advice i will be happy !

zachgoll: @AurelienConte if I'm following you correctly here, I believe this is expected behavior.

The time series graph will show _net worth_ (assets - liabilities) while the pie chart here is showing assets only.

app/javascript/controllers/pie_chart_controller.js
```diff
@@ -15,6 +20,7 @@ export default class extends Controller {
   #d3ViewboxHeight = 200;
 
   connect() {
+    console.log(this.dataValue);
```
zachgoll, line 23: Cleanup?

app/helpers/value_groups_helper.rb
```diff
@@ -1,17 +1,29 @@
...
+          formatted_value: format_money(child.sum.amount.to_f, precision: 2, unit: child.sum.currency.symbol, separator: ".", delimiter: ","),
           bg_color: accountable_bg_class(child.name),
           fill_color: accountable_fill_class(child.name)
         }
       end
       .filter { |child| child[:value] > 0 }
-      .to_json
+
+    total = values.sum { |child| child[:value] }
+    formatted_total = format_money(total, precision: 2, unit: value_group.sum.currency.symbol, separator: ".", delimiter: ",")
```
zachgoll, line 19: I think we're getting closer here, but we cannot rely on the `separator` and `delimiter` being fixed values.  These will be reversed in some cases depending on the currency that is being formatted.

A potential data structure that we could pass to the frontend could be:

```
{
  cents_str: child.sum.cents_str,
  dollar_str: format_money_without_symbol(child.sum, precision: 0)
  currency_symbol: child.sum.currency.symbol
}
```

This allows us to retain the original value while also getting the individual parts of the value that can be independently displayed.

zachgoll:
@AurelienConte do you plan to continue work on this?  No worries either way, just let me know!

AurelienConte:
Sorry i was on vacation, I just come back today at work ! 
I plan to come back on the pull request tomorrow or Wednesday 😄

zachgoll:
@AurelienConte sounds great!

app/helpers/value_groups_helper.rb
```diff
@@ -1,28 +1,40 @@
...
+        percent_of_total: child.percent_of_total.round(1).to_f,
+        value: child.sum.amount.to_f,
+        value_str: {
+          fractional_part: child.sum.cents_str,
+          main_part: format_money_without_symbol(child.sum.amount.to_f, precision: 0, unit: child.sum.currency.symbol)
+        },
+        currency: {
+          iso_code: child.sum.currency.iso_code,
+          symbol: child.sum.currency.symbol,
+          displayed_before_value: child.sum.currency.default_format.start_with?("%u")
```
AurelienConte, line 15: i have added this value to known how to display the symbol in the front like : 

image

And

image


Where sometimes it is on the front of the value and sometimes in the back

app/helpers/value_groups_helper.rb
```diff
@@ -1,28 +1,40 @@
...
 
     {
       values: values,
-      total: total,
-      formatted_total: formatted_total,
-      currency_symbol: value_group.sum.currency.symbol
+      value: total_value,
+      value_str: {
+        main_part: main_part,
+        fractional_part: fractional_part
```
AurelienConte, line 31: i have choose terms more generic like ``main_part`` and ``fractional_part``, do you think this is good or you want to stick with something like ``dollars_str`` and ``cents_str`` ?

AurelienConte:
I made changes to the feat, if you think this is good to you i will start modifying my tests to match the code changes

AurelienConte:
> I made changes to the feat, if you think this is good to you i will start modifying my tests to match the code changes

@zachgoll did you find any time to check my work ?

app/helpers/value_groups_helper.rb
```diff
@@ -1,17 +1,41 @@
...
+      },
+      currency: {
+        iso_code: value_group.sum.currency.iso_code,
+        symbol: value_group.sum.currency.symbol,
+        displayed_before_value: value_group.sum.currency.default_format.start_with?("%u")
+      }
+    }.to_json
```
zachgoll, line 38: I think this method here is getting a bit too complex and we need to explore some alternative solutions.  Based on this logic, it seems that the challenge we need to solve is:

- How do we split a monetary value into "parts" (symbol, major unit, minor unit)
- How do we _order_ the "parts" correctly (symbol before / after)

To solve this, I think we have 2 possible solutions:

1. Add some additional methods to `Money` class to help us 
2. Simplify all of this and just pass a single, formatted currency value to the frontend

I'm thinking we go with option 2.  

I worked some of this out locally, here's my git diff:

```diff
maybe $ git diff 
diff --git a/app/helpers/value_groups_helper.rb b/app/helpers/value_groups_helper.rb
index c8e96d0..b6d0448 100644
--- a/app/helpers/value_groups_helper.rb
+++ b/app/helpers/value_groups_helper.rb
@@ -1,17 +1,13 @@
 module ValueGroupsHelper
   def value_group_pie_data(value_group)
:...skipping...
diff --git a/app/helpers/value_groups_helper.rb b/app/helpers/value_groups_helper.rb
index c8e96d0..b6d0448 100644
--- a/app/helpers/value_groups_helper.rb
+++ b/app/helpers/value_groups_helper.rb
@@ -1,17 +1,13 @@
 module ValueGroupsHelper
   def value_group_pie_data(value_group)
-    value_group.children
-      .map do |child|
-        {
-          label: to_accountable_title(Accountable.from_type(child.name)),
-          percent_of_total: child.percent_of_total.round(1).to_f,
-          value: child.sum.amount.to_f,
-          currency: child.sum.currency.iso_code,
-          bg_color: accountable_bg_class(child.name),
-          fill_color: accountable_fill_class(child.name)
-        }
-      end
-      .filter { |child| child[:value] > 0 }
-      .to_json
+    value_group.children.map do |child|
+      {
+        label: to_accountable_title(Accountable.from_type(child.name)),
+        percent_of_total: child.percent_of_total.round(1).to_f,
+        formatted_value: format_money(child.sum),
+        bg_color: accountable_bg_class(child.name),
+        fill_color: accountable_fill_class(child.name)
+      }
+    end.to_json
   end
 end
diff --git a/app/javascript/controllers/pie_chart_controller.js b/app/javascript/controllers/pie_chart_controller.js
index f68fe5b..1bb4144 100644
--- a/app/javascript/controllers/pie_chart_controller.js
+++ b/app/javascript/controllers/pie_chart_controller.js
@@ -5,6 +5,7 @@ import * as d3 from "d3";
 export default class extends Controller {
   static values = {
     data: Array,
+    total: String,
     label: String,
   };
 
@@ -38,7 +39,7 @@ export default class extends Controller {
 
   #draw() {
     this.#d3Container.attr("class", "relative");
-    this.#d3Content.html(this.#contentSummaryTemplate(this.dataValue));
+    this.#d3Content.html(this.#contentSummaryTemplate());
 
     const pie = d3
       .pie()
@@ -75,23 +76,17 @@ export default class extends Controller {
         this.#d3Svg
           .selectAll(".arc path")
           .attr("class", (d) => d.data.fill_color);
-        this.#d3ContentMemo.html(this.#contentSummaryTemplate(this.dataValue));
+        this.#d3ContentMemo.html(this.#contentSummaryTemplate());
       });
   }
 
-  #contentSummaryTemplate(data) {
-    const total = data.reduce((acc, cur) => acc + cur.value, 0);
-    const currency = data[0].currency;
-
-    return `${this.#currencyValue({
-      value: total,
-      currency,
-    })} <span class="text-xs">${this.labelValue}</span>`;
+  #contentSummaryTemplate() {
+    return `<span class="text-xl text-gray-900 font-medium">${this.totalValue}</span> <span class="text-xs">${this.labelValue}</span>`;
   }
 
   #contentDetailTemplate(datum) {
     return `
-      <span>${this.#currencyValue(datum)}</span>
+      <span class="text-xl text-gray-900 font-medium">${datum.formatted_value}</span>
       <div class="flex flex-row text-xs gap-2 items-center">
       <div class="w-[10px] h-[10px] rounded-full ${datum.bg_color}"></div>
         <span>${datum.label}</span>
@@ -100,21 +95,6 @@ export default class extends Controller {
     `;
   }
 
-  #currencyValue(datum) {
-    const formattedValue = Intl.NumberFormat(undefined, {
-      style: "currency",
-      currency: datum.currency,
-      currencyDisplay: "narrowSymbol",
-    }).format(datum.value);
-
-    const firstDigitIndex = formattedValue.search(/\d/);
-    const currencyPrefix = formattedValue.substring(0, firstDigitIndex);
-    const mainPart = formattedValue.substring(firstDigitIndex);
-    const [integerPart, fractionalPart] = mainPart.split(".");
-
-    return `<p class="text-gray-500 -space-x-0.5">${currencyPrefix}<span class="text-xl text-gray-900 font-medium">${integerPart}</span>.${fractionalPart}</p>`;
-  }
-
   get #radius() {
     return Math.min(this.#d3ViewboxWidth, this.#d3ViewboxHeight) / 2;
   }
diff --git a/app/views/pages/dashboard/_allocation_chart.html.erb b/app/views/pages/dashboard/_allocation_chart.html.erb
index 4c665af..380f3ce 100644
--- a/app/views/pages/dashboard/_allocation_chart.html.erb
+++ b/app/views/pages/dashboard/_allocation_chart.html.erb
@@ -11,7 +11,8 @@
           data-controller="pie-chart"
           class="w-full aspect-1"
           data-pie-chart-label-value="Total Assets"
-          data-pie-chart-data-value="<%= value_group_pie_data(account_groups[:assets]) %>">
+          data-pie-chart-data-value="<%= value_group_pie_data(account_groups[:assets]) %>"
+          data-pie-chart-total-value="<%= format_money(account_groups[:assets].sum, precision: 0) %>">
         </div>
       </div>
     </div>
@@ -21,7 +22,8 @@
           data-controller="pie-chart"
           class="w-full aspect-1"
           data-pie-chart-label-value="Total Debts"
-          data-pie-chart-data-value="<%= value_group_pie_data(account_groups[:liabilities]) %>">
+          data-pie-chart-data-value="<%= value_group_pie_data(account_groups[:liabilities]) %>"
+          data-pie-chart-total-value="<%= format_money(account_groups[:liabilities].sum, precision: 0) %>">
         </div>
       </div>
     </div>
```

AurelienConte: Okay, i have tested your proposal and it is sooo much simpler than mine, i'm going with it okay ?

zachgoll: Yep, looking good!  Just left a few comments.

app/helpers/value_groups_helper.rb
```diff
@@ -1,17 +1,14 @@
...
-          fill_color: accountable_fill_class(child.name)
-        }
-      end
-      .filter { |child| child[:value] > 0 }
-      .to_json
+    value_group.children.filter { |child| child.sum > 0 }.map do |child|
+      {
+        label: to_accountable_title(Accountable.from_type(child.name)),
+        percent_of_total: child.percent_of_total.round(1).to_f,
+        value: child.sum.amount.to_f,
```
zachgoll, line 7: We don't need `value` here anymore since we're passing in `formatted_value`

app/javascript/controllers/pie_chart_controller.js
```diff
@@ -101,18 +96,13 @@ export default class extends Controller {
...
-
-    return `<p class="text-gray-500 -space-x-0.5">${currencyPrefix}<span class="text-xl text-gray-900 font-medium">${integerPart}</span>.${fractionalPart}</p>`;
+    return `
+      <p class="text-gray-500 -space-x-0.5">
+        ${datum.currency_displayed_before_value === true ? datum.currency_symbol : ""}
+          <span class="text-xl text-gray-900 font-medium">${datum.value_str.main_part}</span>
+          ${datum.value_str.fractional_part ? "." + datum.value_str.fractional_part : ""}
+        ${datum.currency_displayed_before_value !== true ? datum.currency_symbol : ""}
+      </p>`;
   }
```
zachgoll, line 106: We should be able to delete this entire function

test/controllers/pie_chart_controller_test.rb
```diff
@@ -0,0 +1,22 @@
...
+    assert_amount_balance '48550.80'
+  end
+
+  private
+  def assert_amount_balance(expected_value)
+    value_element = find('div.absolute.inset-0.w-full.text-center p span.text-xl') # Fetch the pie chart text
+    entire_text = value_element.find(:xpath, '..').text # Get full text of the <p> element
+    extracted_value = entire_text.strip.gsub(/[^\d.]/, '').sub(/^0+/, '') # Removing non-digit characters except the decimal point
+    assert_equal expected_value, extracted_value, "The pie chart value is not displayed correctly"
+  end
```
zachgoll, line 20: I think we should test this functionality in a new system test called `test/system/dashboard_test.rb`, and for now, we can have a single test:

```rb
class DashboardTest < ApplicationSystemTestCase
  test "displays pie chart" do 
    # click on assets
    # assert formatted value
    # click on debts 
    # assert formatted value
  end
end
```

zachgoll (commented):
The code is looking good here.  We will just need to move the test and fix the lint errors (you can use `bin/erblint` to check this locally) and then we can get it merged!

AurelienConte:
Perfect, I'll start testing when I have some time

AurelienConte:
@zachgoll sorry but i think i can not find enough time to finish this ... can you finish this for me ?

zachgoll:
Going to close this out for now and will circle back when I finish some of the investment work I'm doing.

Issue #860, opened 2024-06-11
Transaction pagination should be in a "fixed" container for easier UX interactions


zachgoll:
## Feature Overview

![CleanShot 2024-06-11 at 09 15 07](https://github.com/maybe-finance/maybe/assets/16676157/7d6d137f-057b-4665-a47b-04c4163dc3d2)

Currently, the transaction pagination is not very helpful to the user because it requires them to scroll to the bottom of the screen every time they need to advance pages.

## Requirements

_If there is a missing / incorrect requirement, please leave a comment before starting work on this._

- [ ] Transactions list height is fixed and allows scroll
- [ ] User can select items per page
- [ ] When the user starts scrolling, the top/bottom border show
  - If this requirement cannot be done with pure CSS, let's keep it simple and just exclude the border by default

## Designs

Below are the designs you should follow while implementing this:

https://www.figma.com/design/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?node-id=3855-98

## Reminders

- Make sure to review our [contributing guidelines](https://github.com/maybe-finance/maybe/blob/main/CONTRIBUTING.md) before starting on an issue
- We do our best to define a clear spec for new features and fixes, but think of them as "suggestions", not "hard requirements".  We welcome ideas and suggestions!
    - If you see missing requirements to this issue, please leave a comment below explaining what is missing and why it is important.
    - If you see a requirement that you think is _incorrect_ or _not optimal_, please leave a comment explaining what you think needs to change below.

PR #873, opened 2024-06-15
feat: Transaction pagination Improvements

karankiri:
Issue: https://github.com/maybe-finance/maybe/issues/860

Completes below requirements:
 - [x] Transactions list height is fixed and allows scroll
 - [x] User can select items per page
 - [ ] When the user starts scrolling, the top/bottom border show
      - [x] If this requirement cannot be done with pure CSS, let's keep it simple and just exclude the border by default

Demo Video:
https://drive.google.com/file/d/1eT6V83gjo0m9hR69TreYroPZw30ErNFI/view?usp=sharing


app/views/transactions/index.html.erb
```diff
@@ -3,39 +3,45 @@
...
         <%= render "selection_bar" %>
       </div>
-
-      <div class="grid grid-cols-12 bg-gray-25 rounded-xl px-5 py-3 text-xs uppercase font-medium text-gray-500 items-center mb-4">
-        <div class="pl-0.5 col-span-4 flex items-center gap-4">
-          <%= check_box_tag "selection_transaction",
-                            class: "maybe-checkbox maybe-checkbox--light",
-                            data: { action: "bulk-select#togglePageSelection" } %>
-          <p class="col-span-4">transaction</p>
+      <div class="h-[calc(100dvh-370px)] overflow-y-auto px-4 border-y border-alpha-black-100">
```
zachgoll, line 16: In regards to the calculated height, I think an improvement we can make to this layout is to use flex columns to dynamically size this.  This will require a few small tweaks to the overall layout, but I think it will be nice to have everything relative.  Here's the diff that I used to get it working:

```diff
diff --git a/app/views/layouts/with_sidebar.html.erb b/app/views/layouts/with_sidebar.html.erb
index f014dd6..ca824d4 100644
--- a/app/views/layouts/with_sidebar.html.erb
+++ b/app/views/layouts/with_sidebar.html.erb
@@ -7,7 +7,7 @@
         <%= render "layouts/sidebar" %>
       <% end %>
     </div>
-    <main class="grow px-20 pt-6 pb-32 h-full overflow-y-auto">
+    <main class="grow px-20 py-6 h-full overflow-y-auto">
       <%= yield %>
     </main>
   </div>
diff --git a/app/views/transactions/index.html.erb b/app/views/transactions/index.html.erb
index 5fe827d..ff40b02 100644
--- a/app/views/transactions/index.html.erb
+++ b/app/views/transactions/index.html.erb
@@ -1,20 +1,19 @@
-<div class="space-y-4">
+<div class="space-y-4 h-full flex flex-col overflow-y-auto">
   <%= render "header" %>
 
   <%= render partial: "transactions/summary", locals: { totals: @totals } %>
 
-  <div id="transactions" data-controller="bulk-select" data-bulk-select-resource-value="<%= t(".transaction") %>" class="bg-white rounded-xl border border-alpha-black-25 shadow-xs pb-4">
+  <div id="transactions" data-controller="bulk-select" data-bulk-select-resource-value="<%= t(".transaction") %>" class="grow overflow-y-auto flex flex-col bg-white rounded-xl border border-alpha-black-25 shadow-xs pb-4">
 
     <div class="p-4 pb-0">
       <%= render partial: "transactions/searches/search", locals: { transactions: @transactions } %>
     </div>
 
-
     <% if @transactions.present? %>
       <div hidden id="transaction-selection-bar" data-bulk-select-target="selectionBar">
         <%= render "selection_bar" %>
       </div>
-      <div class="h-[calc(100dvh-370px)] overflow-y-auto px-4 border-y border-alpha-black-100">
+      <div class="grow overflow-y-auto px-4 border-alpha-black-100">
         <div class="grid grid-cols-12 bg-gray-25 rounded-xl px-5 py-3 text-xs uppercase font-medium text-gray-500 items-center mb-4">
           <div class="pl-0.5 col-span-4 flex items-center gap-4">
             <%= check_box_tag "selection_transaction",
```

app/views/transactions/_pagination.html.erb
```diff
@@ -1,39 +1,48 @@
...
-      <div class="inline-flex items-center px-3 py-3 text-sm font-medium hover:border-gray-300">
-        <%= lucide_icon("chevron-right", class: "w-5 h-5 text-gray-200") %>
-      </div>
-    <% end %>
+  <% per_page = params[:per_page] || "50" %>
+  <div class="flex items-center gap-4">
+  <p class="text-sm text-gray-500">Rows per page</p>
+  <%= form_with url: transactions_path, method: :get, class: "flex items-center gap-4", data: { controller: "auto-submit-form" } do %>
+    <%= render partial: "pagination_per_page_selector", locals: { value: per_page } %>
+  <% end %>
```
zachgoll, line 46: As a result of the other changes that I mentioned, the custom select here will have "floating" issues where it won't be able to break out of the bounding div.

In the future, we'll be building out a design system for components like custom selects to ensure that these behaviors are handled appropriately, but right now, we've moved towards the path of using native HTML form elements until we've got this design system in place.

No way to know that on your end so no worries at all.  Here's a tested and completed replacement form with native controls that you can drop in here to get it all working:

```erb
  <%= form_with url: transactions_path,
                builder: ActionView::Helpers::FormBuilder,
                method: :get,
                class: "flex items-center gap-4",
                data: { controller: "auto-submit-form" } do |f| %>
    <%= f.label :per_page, t(".rows_per_page"), class: "text-sm text-gray-500" %>
    <%= f.select :per_page,
                 options_for_select(["10", "20", "30", "50"], params[:per_page]),
                 {},
                 class: "py-1.5 pr-8 text-sm text-gray-900 font-medium border border-gray-200 rounded-lg focus:border-gray-900 focus:ring-gray-900 focus-visible:ring-gray-900",
                 data: { "auto-submit-form-target": "auto" } %>
  <% end %>
```

zachgoll: Just an FYI, you'll need to add the `.rows_per_page` key to the `config/locales/views/transactions/en.yml` locale file as "Rows per page"

zachgoll (commented):
Thanks for tackling this!  Functionality works great.  Just added a few comments with some tweaks around form controls.

karankiri:
Thanks for the review @zachgoll , will fix it today

app/views/transactions/_pagination.html.erb
karankiri: Replace custom select with browser select and move the text to localization file

app/views/transactions/index.html.erb
```diff
@@ -1,41 +1,46 @@
...
         </div>
-
-        <p class="col-span-3 pl-4">category</p>
-        <p class="col-span-3">account</p>
-        <p class="col-span-2 justify-self-end">amount</p>
-      </div>
-      <div class="space-y-6">
-        <% group_transactions_by_date(@transactions).each do |date, group| %>
-          <%= render partial: "date_group", locals: { date:, group: } %>
-        <% end %>
```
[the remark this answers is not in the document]
karankiri: @zachgoll fixed it

karankiri:
@zachgoll I've fixed the template. Still one of the unit test is failing due to per page implementation. 


```
test "can navigate to paginated result" do
    get transactions_url(page: 2)
    assert_response :success

    @recent_transactions[10, 10].each do |transaction|
      dom_id_str = "#" + dom_id(transaction)
      puts "Expected DOM ID: #{dom_id_str}"
      assert_dom "#" + dom_id(transaction), count: 1
    end
  end

```
I tried to debug it and found that  test controller `test/controllers/transactions_controller_test.rb` and actual controller `app/controllers/transactions_controller.rb` are returning different data.

@zachgoll if you can point me in the correct direction, i can fix it. Thanks

zachgoll:
@karankiri I think this test failure could be potentially related to some of the recent changes with the transfers.  

I think we may need to update the logic to look like this:

```
  setup do
    sign_in @user = users(:family_admin)
    @transaction = transactions(:checking_one)
    @recent_transactions = @user.family.transactions.ordered.limit(20).to_a
  end

  test "can navigate to paginated result" do
    get transactions_url(page: 2)
    assert_response :success

    @recent_transactions[10, 10].select { |t| t.transfer_id == nil }.each do |transaction|
      assert_dom "#" + dom_id(transaction), count: 1
    end
  end
```

The reason it was failing is because transfer rows will include exactly 2 transactions (hidden), which are not picked up by DOM assertions.

karankiri:
@zachgoll Thanks, Updated unit test, Now it seems to be working

zachgoll (commented):
@karankiri nice, looks like we have just a few more UI bugs to work through here prior to merging.

- Borders need to be removed (see [this code suggestion](https://github.com/maybe-finance/maybe/pull/873#discussion_r1642949075))
- When transactions don't occupy the full space OR filters are added, the pagination controls seem to lose their position


https://github.com/maybe-finance/maybe/assets/16676157/a65c145d-b603-46bf-b3e8-059588475462

karankiri:
@zachgoll To clarify the second point

> When transactions don't occupy the full space OR filters are added, the pagination controls seem to lose their position

The current/previous logic is that only when page > 1 then we're showing the pagination component. Should we now show the pagination in all the case?

zachgoll:
> @zachgoll To clarify the second point
> 
> > When transactions don't occupy the full space OR filters are added, the pagination controls seem to lose their position
> 
> The current/previous logic is that only when page > 1 then we're showing the pagination component. Should we now show the pagination in all the case?

I think we can keep the current logic of hiding the pagination controls when there is only 1 page to show.

We'll just need to tweak the styles a bit so when there isn't a full page of results, the empty space at the bottom doesn't show.  

It should match this frame:

https://www.figma.com/design/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?node-id=2029-890&t=hjF0MURznALHMZyL-4

karankiri:
@zachgoll Fixed it now, Please checkout here https://drive.google.com/file/d/18ZIvnfm2iDOiRyaDARMp6ILVQ_pvd0Gf/view?usp=sharing

zachgoll (approved):
Looks good now!  Thanks for the work on this, much easier to use this feature now :)

PR #876, opened 2024-06-17
docs: add Zeabur as another one-click deploy option

MichaelYuhe:


zachgoll:
@MichaelYuhe thanks for the submission!  Unfortunately, we're not accepting any new deployment options for our "official" guides at this time.  The main reason for this is so that our small team can fully support each deployment option.

I'm going to close this out, but this would be a great type of thing to put in a [Feature Request Discussion](https://github.com/maybe-finance/maybe/discussions/categories/feature-requests) so that other community members can vote on it to become an official deployment option and so they can see the instructions you've curated if they want to host on this "unofficial" deployment option.

PR #889, opened 2024-06-19
feat: add language switch functionality and Chinese

vsme:
https://github.com/maybe-finance/maybe/assets/18651467/5f817203-6887-41e3-9179-545ce4c30de3

1. Added i18n multilingual switching feature
2. Added translations for Chinese


zachgoll:
@vsme this is excellent work and really appreciate the contribution.

I'm a little hesitant to merge this into `main` right now given the speed at which these translation files are changing (we're adding 20+ new translation keys on a weekly basis right now).

I would love to here suggestions from the community on this—what is the best way to maintain a large set of translations?  As a primarily English-speaking core team, we will have no way to keep these updated accurately and it will add significant overhead to development.

From what I can tell, there are a few ways we could handle this:

- Forks of codebase (temporary) - currently, this seems like the best interim solution, but won't be viable long term as we release a hosted version of the app
- AI-powered translations - we may be able to set something up in CI that automatically generates a set of locales using AI translations
- 3rd party services

Interested to hear some thoughts on this.

vsme:
> @vsme this is excellent work and really appreciate the contribution.
> 
> I'm a little hesitant to merge this into `main` right now given the speed at which these translation files are changing (we're adding 20+ new translation keys on a weekly basis right now).
> 
> I would love to here suggestions from the community on this—what is the best way to maintain a large set of translations? As a primarily English-speaking core team, we will have no way to keep these updated accurately and it will add significant overhead to development.
> 
> From what I can tell, there are a few ways we could handle this:
> 
> * Forks of codebase (temporary) - currently, this seems like the best interim solution, but won't be viable long term as we release a hosted version of the app
> * AI-powered translations - we may be able to set something up in CI that automatically generates a set of locales using AI translations
> * 3rd party services
> 
> Interested to hear some thoughts on this.

Hi @zachgoll ,

Thank you for your message and for considering the different options for maintaining our translations.

I believe that AI translations might not always be accurate and may not capture the nuances of native expressions. If it's possible, I am willing to take responsibility for the Chinese translations to ensure their accuracy and cultural relevance.

vsme:
Initially, AI-powered translations are feasible, but human intervention will still be necessary for any inaccuracies.

jakubkottnauer:
I have used https://crowdin.com/pricing which I was very happy with. Should be free for open source projects

zachgoll:
@vsme appreciate it!  And yes, AI will only get us 80-90% of the way there.  We'll likely need a service or group of contributors to handle the translations.

The main reason we're holding off on this right now is because on a weekly basis, we're adding, editing, and reorganizing translations in such large quantities that it would be impossible to effectively manage and keep up with all the required changes.

We will need to get the app to a point where these translations are not being changed as often before we can build an official process around this.

@jakubkottnauer thanks for the suggestion, will definitely consider this when we reach the point of officially supporting various locales!

therealtimex:
I would definitely bet on AI-powered translations.

vsme:
@zachgoll I have added a `bin/translate-missing-keys` file that can be used in `CI`. I will integrate the translations later. I think Google Translate will be sufficient for now.

vsme:
I have completed the translation feature. You can choose to use either Google Translate or OpenAI by simply adding the environment variables and configuring it in CI. [Reference link](https://github.com/glebm/i18n-tasks?tab=readme-ov-file#translate-missing-keys). This way, every commit will supplement new translations and remove unnecessary ones without affecting the daily development experience. The main goal is to provide significant convenience for non-English speakers, allowing everyone to enjoy this program.

cuicui-V5:
Great work

zachgoll:
@vsme this is great work and appreciate the contribution!

Maybe is 100% committed to providing an internationalized app (it is on our roadmap), but we are currently too early in the development process to accept these changes into the `main` branch.

We are still making _significant_ updates to translation files and overall codebase structure right now, so we're sticking with a single set of translations to simplify and speed up feature development.

I'm going to close this out for now.  When we're ready to accept translations, we will make an announcement and will likely be able to re-open and merge this.

xrcq:
Why doesn't it work at all when I switch languages? My version is v0.2.0-alpha

PR #891, opened 2024-06-19
Ensure correct form's hidden input for selectedIds

ziraqyoung:
When adding transactions for bulk edit/delete, clicking on “Delete” or “Edit” icons  creates hidden form inputs for the selected transactions.

If the modal is closed and a selected transaction(s) is deselected , its hidden form input is already added and wont be removed upon deselecting hence will be accidentally deleted (or updated) on submit.

This PR addresses that by reseting the `paramName`, before creating hidden form inputs.

https://github.com/maybe-finance/maybe/assets/41262193/a3a899dd-4cdd-41bb-990e-dd37cb50a899

### DIFF
--- a/app/javascript/controllers/bulk_select_controller.js
+++ b/app/javascript/controllers/bulk_select_controller.js
@@ -66,6 +66,8 @@ export default class extends Controller {
   }
 
   #addHiddenFormInputsForSelectedIds(form, paramName, transactionIds) {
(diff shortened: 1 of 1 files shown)


zachgoll (approved):
Really good catch here, and thanks for the fix!  That's definitely not a fun bug to deal with as a user :)

PR #903, opened 2024-06-21
Allow transfers based on transactions in different currencies

jakubkottnauer:
This PR add supports for creating transfers from transactions in different currencies.

Screenshot 2024-06-21 at 21 16 22

### DIFF
--- a/app/models/account/transfer.rb
+++ b/app/models/account/transfer.rb
@@ -1,7 +1,8 @@
 class Account::Transfer < ApplicationRecord
   has_many :transactions, dependent: :nullify
 
-  validate :transaction_count, :from_different_accounts, :net_zero_flows, :all_transactions_marked
+  validate :net_zero_flows, if: :single_currency_transfer?
+  validate :transaction_count, :from_different_accounts, :all_transactions_marked
 
   def inflow_transaction
     transactions.find { |t| t.inflow? }
@@ -32,6 +33,10 @@ class Account::Transfer < ApplicationRecord
 
   private
 
+    def single_currency_transfer?
+      transactions.map(&:currency).uniq.size == 1
+    end
+
     def transaction_count
       unless transactions.size == 2
         errors.add :transactions, "must have exactly 2 transactions"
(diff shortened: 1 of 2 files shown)


app/models/account/transfer.rb
```diff
@@ -44,9 +44,25 @@ def from_different_accounts
...
+      if !same_currency
+        rate = ExchangeRate.find_rate_or_fetch from: first.currency, to: second.currency, date: first.date
+        if rate.nil?
+          errors.add :transactions, "must have an exchange rate between currencies"
+          return
+        end
+
+        converted_amount = first.amount * rate.rate
+        allowed_error = 1
+        unless (converted_amount + second.amount).abs < allowed_error
```
jakubkottnauer, line 62: Since we're comparing decimals and even the exchange rate that the banks used for conversion will probably be different than what Maybe uses, we need to allow for the converted amounts to be slightly different, hence `allowed_error`. Not really sure which value would be most reasonable here and 1 seems "good enough"

zachgoll: Just curious—with a multi-currency transfer, do banks always produce exactly 2 transactions?  Or is there a 3rd intermediate transaction for the currency conversion?

I'm wondering if instead of trying to add a tolerance and check for the presence of exchange rates, we just make the following rule:

```rb
validate :net_zero_flows if single_currency_transfer?

private
  def single_currency_transfer?
    transactions.map(&:currency).uniq == 1
  end
```

At this stage, we still have a very minimal UI for matching transactions that will eventually be enhanced with things like:

- Matching multiple transfers in one flow (multi-step)
- Auto-matching based on account type (i.e. outflow from "checking" and inflow to "credit card" would represent a credit card payment in most cases)

I think given how minimal this is right now, we should be _less restrictive_ in our conditions and only enforce things we know to be true without needing additional assumptions like a tolerance for currency conversion.  With the solution above we'll avoid quite a bit of complexity and the user will still be able to "optimistically" assign transfers.

jakubkottnauer: From what I've seen, it always ends up being 2 transactions. But my experience is limited to 5-6 banks in the EU so I wouldn't dare generalising.

I like your idea about playing it safe and simply skipping the net-zero validation for multi-currency for now 👍 I'll make the change

jakubkottnauer: @zachgoll updated 😄

zachgoll: That's fine, I was just curious as I don't have any personal experience with the multi-currency accounts.  So anything helps!

Thanks for the update.  I think this is a good starting approach that we can iterate on as this matching flow gets more complex (especially when we get into auto-matching provider transactions during the sync process)!

PR #907, opened 2024-06-22
Format money with currency on Trend view

### LINKED ISSUE #906: Bug: Valuation history table displays changes in USD regardless of currency setting
**Describe the bug**
The valuations table now display the changes in USD even when it is not the configured currency.

**To Reproduce**
1. Have a non-USD account
2. Add some valuation changes

**Expected behavior**
Changes displayed in actual currency (as before).

**Screenshots / Recordings**
If applicable, add screenshots or short video recordings to help show the bug in more detail.

Screenshot 2024-06-22 at 14 52 43

**Additional context**

joehoyle:
Fixes #906

### DIFF
--- a/app/views/account/valuations/_valuation.html.erb
+++ b/app/views/account/valuations/_valuation.html.erb
@@ -21,7 +21,7 @@
       <% if valuation.trend.direction.flat? %>
         <%= tag.span t(".no_change"), class: "text-gray-500" %>
       <% else %>
-        <%= tag.span format_money(valuation.trend.value) %>
+        <%= tag.span format_money(Money.new(valuation.trend.value, valuation.value_money.currency)) %>
         <%= tag.span "(#{valuation.trend.percent}%)" %>
       <% end %>
     </div>


app/views/account/valuations/_valuation.html.erb
```diff
@@ -21,7 +21,7 @@
       <% if valuation.trend.direction.flat? %>
         <%= tag.span t(".no_change"), class: "text-gray-500" %>
       <% else %>
-        <%= tag.span format_money(valuation.trend.value) %>
+        <%= tag.span format_money(Money.new(valuation.trend.value, valuation.value_money.currency)) %>
```
zachgoll, line 24: This will work for nearly all cases, but I think we can address this a bit further upstream and avoid having to do these additional conversions:

https://github.com/maybe-finance/maybe/blob/a5212f0f5e8fda87c9aed63e6d4d42424e9a7cd9/app/models/account/valuation.rb#L46-L51

If I'm not mistaken, I believe we can update to this and things will work fine:

```rb
def create_trend
  TimeSeries::Trend.new \
    current: self.value_money,
    previous: previous_valuation&.value_money,
    favorable_direction: account.favorable_direction
end
```

joehoyle: yes, when I tried to do this I, `current` and `previous` need to be the same type, I think that's what was the immediate issue. Is `previous_valuation&.value_money` maybe scalar?

zachgoll:
This should be fixed via #923, closing out.

PR #917, opened 2024-06-24
Fix #910

### LINKED ISSUE #910: Bug: Multi-currency Support
**Describe the bug**
I live in a country where local accounts do not support international transactions. Because of this, I have a local bank account and a domiciliary one under one financial institution. I would like to be able to manage these account separately. Currently, the app just adds the value of one account to the other even when they are not the same currency type. 

**To Reproduce**
1. Go to 'Settings > General (Accounts)'
2. Create a new Institution: Let's call it 'A'
3. Add a new account to this institution with the current balance as 1 and set the currency as USD (You can call it whatever)
4. Add another account to the same institution with the current balance as 1 and set the currency to anything other than USD
5. The total amount would show as $2

**Expected behavior**
What I would like to see here is the total amount on the dashboard should account for both currencies. In this case, I am expecting to see: "$1, ₦1", for instance. 

**Screenshots / Recordings**

image

**Additional context**
I understand that this might be a difficult problem to solve, but I have set up my current banking structure on the app and things are very inaccurate because it isn't taking into account the type of currency used when creating an account and using that to calculate the sum total.

tonyvince:
Closes #910


app/helpers/accounts_helper.rb
```diff
@@ -37,4 +37,11 @@ def class_mapping(accountable_type)
       "Vehicle" => { text: "text-pink-500", bg: "bg-pink-500", bg_transparent: "bg-pink-500/10", fill: "fill-pink-500", hex: "#F23E94" }
     }.fetch(accountable_type, { text: "text-gray-500", bg: "bg-gray-500", bg_transparent: "bg-gray-500/10", fill: "fill-gray-500", hex: "#737373" })
   end
+
+  def format_accounts_balance(accounts, options = {})
+    accounts.group_by(&:currency)
+            .transform_values { |acc| acc.sum(&:balance_money) }
+            .map { |_currency, balance| balance.to_s }
+            .join(", ")
+  end
```
zachgoll, line 46: I like this solution!  We're already doing something very similar here:

https://github.com/maybe-finance/maybe/blob/a5212f0f5e8fda87c9aed63e6d4d42424e9a7cd9/app/views/transactions/_date_group.html.erb#L13-L19

Can we consolidate both of these into one?

tonyvince: Done!

app/helpers/application_helper.rb
```diff
@@ -122,4 +122,11 @@ def format_money_without_symbol(number_or_money, options = {})
     options.reverse_merge!(money.default_format_options)
     ActiveSupport::NumberHelper.number_to_delimited(money.amount.round(options[:precision] || 0), { delimiter: options[:delimiter], separator: options[:separator] })
   end
+
+  def format_amount_by_curreny(collection:, money_method:, seperator: " | ", negate: false, options: {})
```
zachgoll, line 126:
```suggestion
  def totals_by_currency(collection:, money_method:, separator: " | ", negate: false)
```

Thoughts on this?  This is doing more than just "formatting", so I think this might describe the output a little more clearly?

tonyvince: Agree!

zachgoll (approved):
Looks good!

Issue #861, opened 2024-06-11
Bug: During transaction import, when user returns back to "mappings" step, all columns have reset in the forms


zachgoll:
**Where did this bug occur?**

- [ ] Local development
- [x] Self hosted app (i.e. Docker)

**Describe the bug**

- Column mappings reset when user returns to "configure" step
- Additionally, for imports with large number of rows, the "Save" request for the column mappings is very slow

While this does not cause any immediate problems, it is an inconvenience to the user if they want to "re-map" their columns.

**To Reproduce**

1. Go to transaction import
2. Paste a CSV
3. Map your columns
4. Manually go back to the "Configure" step, and you should see that all the form fields have reset to the same column 

**Expected behavior**

The "configure" step should retain the same mappings the user set originally.

**Screenshots / Recordings**
If applicable, add screenshots or short video recordings to help show the bug in more detail.

https://github.com/maybe-finance/maybe/assets/16676157/9eb38b9b-2e22-4f01-841d-8f55b7a9b5cc

**Additional context**

PR #918, opened 2024-06-24
Fix issue #861: Correct header selection logic in get_selected_header_for_field method

igorcarvalhh:
**Fix issue #861**

The get_selected_header_for_field method was incorrectly using the entire field object instead of the field key to dig into the column_mappings hash. This caused an error when trying to retrieve the selected header for a field.


zachgoll (approved):
Awesome, thanks for the fix!

PR #922, opened 2024-06-26
Fix #921

### LINKED ISSUE #921: Bug: Unable to edit transaction
**Describe the bug**
When I try to edit a transaction, the edit form appears. As soon as I edit a value, the edit form disappear and the change is not saved.

**To Reproduce**
1. Go to 'Transactions' or an account transactions history
2. Click on the description of a transaction
3. Edit a value (doesn't matter which one)
4. See error, the edit form is closed without saving

**Expected behavior**
When a value is edited, the form should stay opened and the change should be saved.
*OR*
When a value is edited, the form should stay opened and the change is not saved until we click on a 'save' button (which doesn't exist at the moment)

**Additional context**
Self-hosted instance behind reverse-proxy using docker-compose file

tonyvince:
Fix #921


app/controllers/account/transactions_controller.rb
```diff
@@ -15,7 +15,7 @@ def update
     @transaction.update! transaction_params
     @transaction.sync_account_later
 
-    redirect_back_or_to account_transaction_url(@transaction.account, @transaction), notice: t(".success")
+    render turbo_stream: turbo_stream.replace(@transaction, partial: "transaction", locals: { transaction: @transaction })
```
zachgoll, line 18: Could be wrong here, but I think we'd need to provide an HTML only response for this?

```rb
respond_to do |format|
  format.html { redirect_back_or_to account_transaction_path(@account, @transaction) }
  format.turbo_stream { render turbo_stream: turbo_stream.replace(@transaction) }
end
```

zachgoll (commented):
Looks good, thanks for the fix!

app/controllers/account/transactions_controller.rb
tonyvince: Done!

PR #937, opened 2024-07-02
Fix decimal display for euro currency

zachgoll:
CleanShot 2024-07-02 at 07 24 43

### DIFF
--- a/app/views/shared/_value_heading.html.erb
+++ b/app/views/shared/_value_heading.html.erb
@@ -3,13 +3,7 @@
   <p class="text-sm text-gray-500 font-medium"><%= label %></p>
   <p class="text-gray-900 -space-x-0.5">
     <% if value.is_a?(Money) %>
-      <span class="text-gray-500"><%= value.currency.symbol %></span>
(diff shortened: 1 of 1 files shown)


app/views/shared/_value_heading.html.erb
```diff
@@ -3,13 +3,7 @@
   <p class="text-sm text-gray-500 font-medium"><%= label %></p>
   <p class="text-gray-900 -space-x-0.5">
     <% if value.is_a?(Money) %>
-      <span class="text-gray-500"><%= value.currency.symbol %></span>
-      <span class="<%= size == "lg" ? "text-xl" : "text-lg" %> font-medium"><%= format_money_without_symbol value, precision: 0 %></span>
-      <%- if value.currency.default_precision.positive? -%>
-        <span class="text-gray-500">
-          <%= value.currency.separator %><%= value.cents_str %>
-        </span>
-      <% end %>
```
zachgoll, line 12: See [this comment](https://github.com/maybe-finance/maybe/pull/844#discussion_r1661086241) for more detail behind this change.

Due to the vast number of currencies, this approach will simplify display of currency values significantly.

PR #957, opened 2024-07-08
Add error handling for AccountsController#create

### LINKED ISSUE #956: Bug:  Validation error occurs when adding depository
**Describe the bug**

Validation error occurs when adding depository.

**To Reproduce**

I posted a reproduction video.

**Expected behavior**
Validation error does not occur.

**Screenshots / Recordings**

https://github.com/maybe-finance/maybe/assets/49807271/a95cc3a6-6a8d-4af0-92d5-be49e18c8fd7

**Additional context**

tonyvince:
Fix #956

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -3,7 +3,6 @@ class AccountsController < ApplicationController
 
   include Filterable
   before_action :set_account, only: %i[ edit show destroy sync update ]
-  after_action :sync_account, only: :create
 
   def index
     @institutions = Current.family.institutions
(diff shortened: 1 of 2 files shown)


tonyvince:
During error handling, I noticed that the error message was just `Entries is invalid` instead of the expected `Date has already been taken`. That means validation errors from the child are not bubbling up to the parent object. There is an open issue around this in rails https://github.com/rails/rails/issues/47334

zachgoll (approved):
Thanks for tracking this one down.  I think it's fine for now having the "Entry invalid" message rather than the appropriate "Date taken" message.  Agree with the form validation changes.

PR #967, opened 2024-07-09
Add missing migrations for good_job 4x

### LINKED ISSUE #966: Bug: ERROR:  column good_job_processes.lock_type does not exist at character 416.
Pulled the latest image v0.1.0-alpha.9, the log keep showing the following error.ERROR:  column good_job_processes.lock_type does not exist at character 416. The app still runs but my total net worth balance dropped to one of my account balance instead of the sum of all my depository, checking the other two accounts I am also seeing an error message "An error has occurred during the sync." . 

**Expected behavior**
Net worth balance should be the sum of all depository minus debt.

**Screenshots / Recordings**
Screenshot 2024-07-08 at 9 38 20 PM
Screenshot 2024-07-08 at 10 17 58 PM

tonyvince:
Fix #966


zachgoll (approved):
Awesome, thanks for this fix!

Issue #975, opened 2024-07-12
Add generic frontend validation for better user experience


MagnusHJensen:
**PLEASE READ before opening an issue:**

- Is this a feature request?  Please [open a feature request discussion](https://github.com/maybe-finance/maybe/discussions/new?category=feature-requests).
- Do you need help or have a question?  Please [open a discussion](https://github.com/maybe-finance/maybe/discussions/new/choose) or [join our Discord](https://link.maybe.co/discord) and post to the "help" channel.

----------------------

**Is this issue related to a problem?  Please describe.**
It's a split off issue/task from #799, which limits the scope to adding a frontend validation controller, and not adding it to the entire application.

**Describe the work that needs to be done to address this issue**
Should be extensible to allow for various ways of validating different frontend things in different ways.
- Ex: Non empty input
- Min or max input size.

**Additional context**

Issue #799, opened 2024-05-23
Enable File Upload for CSV transaction imports


zachgoll:
![Image](https://github.com/maybe-finance/maybe/assets/16676157/623b6fa5-4e62-480d-ab91-a14ab2b1a63e)

## Feature Overview

This is a _follow up feature_ to the CSV imports that were already implemented in #708

In the original feature, only "copy/paste" was enabled.  This issue's purpose is to add the ability for a user to directly upload a File.

## Requirements

_If there is a missing / incorrect requirement, please leave a comment before starting work on this._

- [ ] User sees 2 tabs on the "load" step—"Upload CSV" or "Copy & Paste"
- [ ] Copy/paste tab should use existing implementation
- [ ] File upload should allow user to click or drag to upload a CSV file
- [ ] File upload status (likely via ActionCable)
- [ ] All errors handled (see designs) - many of these are already handled by validations on the `Import` model

## Implementation Suggestions

_Below are some ideas for implementation to get you started.  Use your best judgment here—if there's a better way to do things, go for it!_

- We don't necessarily need to store the file that is uploaded.  We simply need to get the CSV string and save it to `raw_csv_str` column on `Import`
  - We may eventually run into performance issues with this, but we'll address that in a later issue if needed.

## Designs

Below are the designs you should follow while implementing this:

https://www.figma.com/design/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?node-id=3656-18086&t=t84SVrxXV2CSiCM7-0

## Reminders

- Make sure to review our [contributing guidelines](https://github.com/maybe-finance/maybe/blob/main/CONTRIBUTING.md) before starting on an issue
- We do our best to define a clear spec for new features and fixes, but think of them as "suggestions", not "hard requirements".  We welcome ideas and suggestions!
    - If you see missing requirements to this issue, please leave a comment below explaining what is missing and why it is important.
    - If you see a requirement that you think is _incorrect_ or _not optimal_, please leave a comment explaining what you think needs to change below.

joehoyle:
Related: when importing a CSV it seems there's no option to skip category / tags, i.e. select "no" field to map those to.

zachgoll:
@joehoyle yeah I had run into this with a few of my own imports too.  Created an issue here that we can cover this in:

https://github.com/orgs/maybe-finance/projects/14/views/1?pane=issue&itemId=65794196

joehoyle:
[...]

PR #977, opened 2024-07-12
Feature/add frontend validation for import flow

MagnusHJensen:
Closes: #975 

This adds a simple validation controller that we can start to use, and allows for expandability by adding more targets in the controller.

## Media

https://github.com/user-attachments/assets/0a4bff77-ad2b-475d-9d15-9690d7e414d2


zachgoll (commented):
Left a comment [here](https://github.com/maybe-finance/maybe/issues/799#issuecomment-2225688507) about the reasoning why we're generally sticking with native validations.

That said, since this is ready to go I'd be open to merging it as I do agree it improves the user experience.

I think we could probably simplify this a bit.  What about something like this?

```js
  static targets = ["required", "submit"]

  connect() {
    this.#validate()
    this.element.addEventListener("input", this.#handleFormChange)
  }

  disconnect() {
    this.element.removeEventListener("input", this.#handleFormChange)
  }

  #handleFormChange = () => {
    this.#validate()
  }

  #validate() {
    const isFormValid = this.requiredTargets.every(target => target.value.trim() !== "")
    this.submitTarget.disabled = !isFormValid
  }
}
```

This has a couple benefits:

- We can delegate the styling logic directly to the view via the Tailwind `disabled:` selector and remove global styles
- We can remove a bit of boilerplate code in general

The view would then be:

```erb
  <%= form_with model: @import, url: load_import_path(@import), data: { controller: "validation" } do |form| %>
    <div>
      <%= form.text_area :raw_csv_str,
                         rows: 10,
                         required: true,
                         "data-validation-target": "required",
                         placeholder: "Paste your CSV file contents here",
                         class: "rounded-md w-full border text-sm border-alpha-black-100 bg-white placeholder:text-gray-400 p-4" %>
    </div>

    <%= form.submit t(".next"), "data-validation-target": "submit", class: "disabled:bg-gray-50 disabled:text-gray-400 disabled:cursor-not-allowed px-4 py-2 block w-full rounded-lg bg-gray-900 text-white text-sm font-medium cursor-pointer hover:bg-gray-700", data: { turbo_confirm: (@import.raw_csv_str? ? { title: t(".confirm_title"), body: t(".confirm_body"), accept: t(".confirm_accept") } : nil) } %>
  <% end %>
```

zachgoll:
Closing out as I think we should [hold off on these client side validations](https://github.com/maybe-finance/maybe/issues/799#issuecomment-2225688507) for the time being to optimize for speed of development.

PR #973, opened 2024-07-11
[WIP] Add csv file upload for transaction import

tonyvince:
Proof of concept for to close #799

PR #978, opened 2024-07-12
Feature/allow for csv upload in import flow

MagnusHJensen:
Closes: #799 

This PR enables CSV upload in the load flow of imports.

It adds a dropzone controller, that through an XMLHttpRequest monitors the upload status, which can give you a progress bar.

It also handles wrong file formats and displays it in the dropzone.
it also uploads the parsed CSV and then redirects you back, where you can manually edit the content of the uploaded CSV while still having the success dropzone.

It is also translated and in accordance to the design spec, which can be found here: 
https://www.figma.com/design/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?node-id=3656-18086&t=t84SVrxXV2CSiCM7-0

## Media

https://github.com/user-attachments/assets/c05f8050-be9f-4196-94ca-3f700588a2cf


app/controllers/imports_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class ImportsController < ApplicationController
   before_action :set_import, except: %i[ index new create ]
+  protect_from_forgery with: :null_session, include: %i[ upload_csv ]
```
MagnusHJensen, line 5: I don't think this is the best practice, however was unsure how to make it work from a JS http request.

app/controllers/imports_controller.rb
```diff
@@ -47,6 +48,24 @@ def load_csv
...
+    raw_csv = params[:file].read
+
+    begin
+      csv = Import::Csv.new(raw_csv)
+      if csv.valid?
+        render turbo_stream: turbo_stream.action(
+          :update_input, "raw_csv_text_area", raw_csv
+        )
+      else
+        render json: { message: "CSV contents is not valid" }, status: :unprocessable_entity
```
MagnusHJensen, line 61: The errors are not translated, since they are only logged in the console here, which could be picked up by a service like Sentry.

The dropzone does translate errors as seen in the translation file.

zachgoll: No worries, think this is okay!

app/controllers/imports_controller.rb
```diff
@@ -47,6 +48,24 @@ def load_csv
...
+      end
+    rescue CSV::MalformedCSVError => error
+      file_extension = params[:file].path.split(".").last
+      render json: { message: "Expected file format CSV, but recieved #{file_extension}", error: error }, status: :bad_request
+    end
+  end
```
zachgoll, line 67: This Turbo Stream solution definitely works, but I think we should try to build this flow without any dependencies on Streams to start as [outlined in the Turbo handbook](https://turbo.hotwired.dev/handbook/streams#progressively-enhance-when-necessary):

> It’s good practice to start your interaction design without Turbo Streams. Make the entire application work as it would if Turbo Streams were not available, then layer them on as a level-up. This means you won’t come to rely on the updates for flows that need to work in native applications or elsewhere without them.

I'm thinking we should build this endpoint a bit closer to what @tonyvince had in #973:

```rb
def upload_csv
  file = import_params[:raw_csv_str]
  @import.raw_csv_str = file.read
  CSV.parse(@import.raw_csv_str)
  if @import.save
    redirect_to configure_import_path(@import), notice: t(".import_loaded")
  else
    flash.now[:error] = @import.errors.full_messages.to_sentence
    render :load, status: :unprocessable_entity
  end
rescue CSV::MalformedCSVError, ArgumentError => error
  flash.now[:error] = error.message
  render :load, status: :unprocessable_entity
end
```

I know some of the designs dictate client-side interactions, but I'm 100% fine deviating a little bit from them for now so that we can achieve the simplest possible flow that's easy to test.

tonyvince:
Screenshot 2024-07-15 at 10 24 05
Awesome job setting up this PR <3 
I think what we have in #973  is closer to the designs. If you don't want to make many changes, maybe we can at least copy the sample file download part?

MagnusHJensen:
> Awesome job setting up this PR ❤️ I think what we have in #973 is closer to the designs. If you don't want to make many changes, maybe we can at least copy the [sample file download part](https://github.com/maybe-finance/maybe/pull/973/files#diff-84ad6e219ec134bdb059b41d19d0017c3d528af110e7e76d1a3f6c4c73c85dcfR38)?
> 
> Screenshot 2024-07-15 at 10 24 05

Hey,

I'm going to move the download part away from turbo stream and just into progressing after a download like you did.

In regards to the table, see this comment: https://github.com/maybe-finance/maybe/issues/799#issuecomment-2214126257 where it was stated that mismatch between designs and application can and will occur as of right now, so the table already in the application is correct, and should not be changed, that is what I understand from it.

tonyvince:
> so the table already in the application is correct, and should not be changed

I was referring to the sample CSV file download part. [See](https://github.com/maybe-finance/maybe/pull/973/files#diff-84ad6e219ec134bdb059b41d19d0017c3d528af110e7e76d1a3f6c4c73c85dcfR31-R44)

MagnusHJensen:
Ah okay, feel free to go with @tonyvince's suggestion, since it's already made and closer to the wanted outcome.

PR #988, opened 2024-07-16
Sanitize input for ilike in Account::Entry.search

### LINKED ISSUE #987: Bug: Handling percent signs in filters
**Describe the bug**
The percent sign in the filter may or may not be a wildcard.

**To Reproduce**

please watch the recording

**Expected behavior**
I don't think it's common to treat percent signs as wildcards.

**Screenshots / Recordings**

https://github.com/user-attachments/assets/2dd89b1c-572a-4955-a03d-b42451ef6942

tonyvince:
Fixes #987

### DIFF
--- a/app/models/account/entry.rb
+++ b/app/models/account/entry.rb
@@ -137,7 +137,7 @@ class Account::Entry < ApplicationRecord
 
     def search(params)
       query = all
-      query = query.where("account_entries.name ILIKE ?", "%#{params[:search]}%") if params[:search].present?
+      query = query.where("account_entries.name ILIKE ?", "%#{sanitize_sql_like(params[:search])}%") if params[:search].present?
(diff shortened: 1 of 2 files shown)


tonyvince:
A [trigram index](https://thoughtbot.com/blog/choosing-the-right-database-index-type#pattern-matching-with-likeilike-gin-index) on `account_entries.name` might be beneficial

zachgoll (approved):

zachgoll:
> A [trigram index](https://thoughtbot.com/blog/choosing-the-right-database-index-type#pattern-matching-with-likeilike-gin-index) on `account_entries.name` might be beneficial

Yeah I think this along with a few other indexes could be helpful to the search process.  I've kept it pretty minimal and unoptimized for now though since we'll eventually be introducing a "global" search and my guess is that we'll have to rethink some of it anyways.

PR #989, opened 2024-07-16
More composable forms

zachgoll:
Did a quick run-through of the codebase to make form building a bit more composable.

Previously, `ApplicationFormBuilder` applied to all form instances and had started to get a little challenging to override when needed.

This PR keeps the core strategy of `ApplicationFormBuilder` in place, but attempts to separate out responsibilities a bit more clearly:

- `styled_form_with` - a helper that allows us to "opt-in" to the form builder (rather than using it as the default for every form)
- `StyledFormBuilder` - this form builder should be responsible for styling form labels + inputs.  It should not attempt to style the _structure_ of any form (e.g. spacing between inputs, layout of inputs, etc.)
- `form_helpers` - I have moved most of the input helpers from the form builder to the helpers file so they can be used both inside and outside of the context of the form builder.

This PR also has a slight re-write of the "Money field" to give forms a bit more flexibility in how they deal with money inputs (currency can be removed, hidden, or directly integrated into the input).


app/controllers/accounts_controller.rb
```diff
@@ -19,6 +19,7 @@ def summary
   end
 
   def list
+    render layout: false
```
zachgoll, line 22: Bug that has been here for a while—turbo frame was linking to itself since the sidebar (part of the layout) calls this.

app/javascript/controllers/select_controller.js
zachgoll, line 1: Previously introduced to enable custom select inputs.

Removing it as we are sticking with Rails + HTML native inputs for the time being until we get a proper design system with Lookbook + ViewComponent setup.

PR #997, opened 2024-07-17
Gereman translation from Models

Alpha63:
Adding frist Step german translation.


zachgoll:
@Alpha63 really appreciate the thought here, but we are not currently accepting translations.  I explained the rationale behind this more in [this comment](https://github.com/maybe-finance/maybe/pull/889#issuecomment-2186701723) and we will let the community know when we're ready to accept these!

PR #998, opened 2024-07-18
Sync notifications and troubleshooting guides

zachgoll:
This PR integrates our notification system directly into the account sync process, and scaffolds out a "troubleshooting" directory of markdown guides that can be rendered in the app and maintained directly in this repo.

https://github.com/user-attachments/assets/122a77d3-9029-453d-83e6-03e59e442f14

## Notifications

I've simplified our notification partial to only deal with 3 possible types, always with a simple string message.  More complex notifications with links and CTAs should be rendered persistently:

- `alert` - warnings and errors, will show red text
- `notice` - success notifications
- `processing` - persistent notifications to indicate a background job in progress

Furthermore, each flash key can pass either a string or array of strings to be appended to the notifications tray:

```rb
flash[:notice] = ["first notification", "second notification"]
flash[:notice] = "one notification"
```

## Troubleshooting guides

This PR scaffolds out a basic system for maintaining troubleshooting guides for the app.

- All guides will be stored under `docs/help` in this repository so that we can accept contributions from users dealing with various provider and account sync errors in the future
- Each guide must have a `title` and `slug`
- In the future, we'll store specific error codes on the `Account::Sync` model to know which guide to render.  Right now, there is a generic placeholder guide rendering for all sync errors.


app/views/shared/_notification.html.erb
```diff
@@ -1,45 +1,45 @@
-<%# locals: (type: "success", content: { title: '', body: ''}, action: { label:'' , url:'' }, options: { auto_dismiss: true }) -%>
+<%# locals: (message:, type: "notice", id: nil, **_opts) %>
```
zachgoll, line 1: **_opts are needed here since turbo rails automatically passes the model object on any `broadcast_<verb>_to` methods, and we construct this partial to send via streams sometimes.

PR #1000, opened 2024-07-19
Basic Portfolio Views

zachgoll:
A simple, starter UI for investment portfolios.  Lots of placeholders and simplistic calculations here that will be the starting point for more nuanced portfolio management moving forward.

https://github.com/user-attachments/assets/57ade850-4473-4870-83d6-cb0d7de30d99


app/models/account/holding.rb
```diff
@@ -1,6 +1,42 @@
...
+  def weight
+    portfolio_value = account.holdings.current.sum(&:amount)
+    portfolio_value.zero? ? 1 : amount / portfolio_value * 100
+  end
+
+  # Basic approximation of cost-basis
+  def avg_cost
+    avg_cost = account.holdings.for(security).where("date <= ?", date).average(:price)
+    Money.new(avg_cost, currency)
+  end
```
zachgoll, line 27: Very simplistic approximation of "cost basis" of a holding.

In the future, we'll introduce other methods such as:

```ruby
def weighted_avg_cost
end

def fifo_cost
end
```

zachgoll:
Still have a few todo items on this as it is highly reliant on the user having a stock price provider setup with the app.

We'll need some sort of fallback UI if that provider is not configured correctly so the user knows what the problem is:

- [ ] Provide a fallback UI (and warning with troubleshooting guide) for when a user creates a portfolio without having a stock price provider configured yet
- [ ] Instead of rendering the current day holdings, render the "most recent" day of calculated holdings (so user doesn't see holdings disappear if accounts haven't synced recently)
- [ ] Integrate Synth stock price endpoint with flow

PR #1003, opened 2024-07-20
Handle CSV files without headers while importing

### LINKED ISSUE #999: ignore first row when importing a CSV
When you try to import a csv either from a file or by pasting the text it always ignores the first row and does not import it as a transaction.
This means that when you import many transactions from several files you always lose 1 transaction at a time.
If you paste or upload a file with a single row, it ignores it and does not show it.

**To Reproduce**
Go to transactions, click import, select CSV and upload a file with, for example, 10 transactions and it only detects 9.

tonyvince:
Fix #999

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -46,6 +46,15 @@ class Import < ApplicationRecord
     update! normalized_csv_str: updated_csv.to_s
   end
 
+  def add_headers_if_missing
+    return if raw_csv_str.blank?
+    first_raw = CSV.new(raw_csv_str).shift
+
+    unless header_row?(first_raw)
+      self.raw_csv_str = first_raw.join(",") + "\n" + raw_csv_str
+    end
+  end
+
   # Type-specific methods (potential STI inheritance in future when more import types added)
   def publish
     update!(status: "importing")
@@ -73,6 +82,14 @@ class Import < ApplicationRecord
   end
 
   private
+
+    def header_row?(row)
+      row.none? { |col| numeric?(col) }
+    end
+
+    def numeric?(string)
+      true if Float(string) rescue false
+    end
 
     def get_normalized_csv_with_validation
       return nil if normalized_csv_str.nil?
--- a/app/controllers/imports_controller.rb
+++ b/app/controllers/imports_controller.rb
@@ -41,6 +41,7 @@ class ImportsController < ApplicationController
   def upload_csv
     begin
       @import.raw_csv_str = import_params[:raw_csv_str].read
+      @import.add_headers_if_missing
     rescue NoMethodError
       flash.now[:alert] = "Please select a file to upload"
       render :load, status: :unprocessable_entity and return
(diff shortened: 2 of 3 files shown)


app/models/import.rb
```diff
@@ -46,6 +46,15 @@ def update_csv!(row_idx:, col_idx:, value:)
     update! normalized_csv_str: updated_csv.to_s
   end
 
+  def add_headers_if_missing
+    return if raw_csv_str.blank?
+    first_raw = CSV.new(raw_csv_str).shift
+
+    unless header_row?(first_raw)
+      self.raw_csv_str = first_raw.join(",") + "\n" + raw_csv_str
```
tonyvince, line 54: header raw is made same as the first raw. Adding a default raw like  "date,name,category,tag,amount" might confuse the user in the next configuration step

app/models/import.rb
```diff
@@ -74,6 +83,14 @@ def expected_fields
 
   private
 
+    def header_row?(row)
+      row.none? { |col| numeric?(col) }
```
tonyvince, line 87: The idea is to assume no header row is present in the CSV file if the first row contains any numeric value. Better suggestions are welcome

app/models/import.rb
tonyvince: The trade-off of this approach is that while it makes the configuration step easier for the user, it makes debugging harder. Someone looking at the database might think the import missed a row because there is no obvious header.

zachgoll: Yeah, as I'm looking at this, I'm wondering if this trade-off is worth it.  When I had originally thought about a solution here I had not considered how brittle it might be to determine whether the first row is a header or a value.

I think this could get really tricky if a user provides a CSV where the transaction names and categories come first.

Would love to hear your thoughts on this, but I'm thinking we'd be better off requiring a user to provide headers for now so that our debugging can remain simple.

We could still provide a simple validation to ensure there are at least 2 headers (for our required data, amount fields), but probably not worth the hassle past that?

app/models/import.rb
zachgoll: See my comment above, will resolve this one.

app/models/import.rb
tonyvince: >  I'm thinking we'd be better off requiring a user to provide headers for now so that our debugging can remain simple.

We would still need to determine whether the first raw is a header. Assuming banks provide CSV export in different languages/formats the only thing that could work is checking if at least one column is numeric or not.

zachgoll: @tonyvince are you saying that we'd still need to determine whether the first row is a header so that we can support both types of CSVs?

My view is that we should just assume all CSVs have headers and let the user fix any issues that arise from it (similar to how we're currently doing)

Long term I can see the use case for adding more automation around this, but I don't think it's a necessary feature right now.

tonyvince: ah ok. Makes sense

PR #1004, opened 2024-07-20
Fix form labels

### LINKED ISSUE #1001: Bug: Form Text Labels Display Incorrect Text on Account Creation Page
**Describe the bug**
On the account creation page, the text labels for the input fields are displaying the text "true" instead of the correct labels (e.g., "Email", "Password").

**To Reproduce**
1. Go to the account creation page.
2. Observe the text labels for the input fields.
3. Notice that the text labels are "true" instead of the expected labels.

**Expected behavior**

The input fields should have the labels "Email", "Password" and "Confirm Password".

**Screenshots / Recordings**
image

**Additional context**
1. Running Maybe through docker. 
2. Updated the application recently

tonyvince:
Fix #1001

### DIFF
--- a/app/helpers/styled_form_builder.rb
+++ b/app/helpers/styled_form_builder.rb
@@ -50,6 +50,7 @@ class StyledFormBuilder < ActionView::Helpers::FormBuilder
     end
 
     def label_html(method, options)
+      return label(method, class: "form-field__label") if options[:label] == true
       options[:label] ? label(method, options[:label], class: "form-field__label") : "".html_safe
     end
 end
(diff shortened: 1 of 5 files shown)


zachgoll (commented):
Really nice solution here!

I like this as our fallback, but I think we need to provide an explicit label in order to leverage I18n translations correct?  Or is there a more automated way to handle that via model translations that you know of?

tonyvince:
> I think we need to provide an explicit label in order to leverage I18n translations correct? Or is there a more automated way to handle that via model translations that you know of?

Not sure I understand what you mean fully but `label: true` as we have [here](https://github.com/maybe-finance/maybe/blob/main/app/views/accounts/accountables/_investment.html.erb#L1) is the [automaded](https://api.rubyonrails.org/v7.1.3.4/classes/ActionView/Helpers/FormHelper.html#method-i-label) way of handling them via model translation. A custom label like `label: t(".email")` will still work with this solution

zachgoll:
@tonyvince my bad, I see what you mean now!  I had missed that piece of logic when updating the forms.

PR #1017, opened 2024-07-22
Set last_login_at only on login instead of every single action

tonyvince:
Currently, every request write to the `users` table updating `last_login_at`. Not sure this adds any value. This PR 
limits writing `users.last_login_at` to `POST /session` and `POST /registration` requests only.

Screenshot 2024-07-22 at 12 45 00

### DIFF
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -1,0 +1,18 @@
+require "test_helper"
+
+class SessionsControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @user = users(:family_admin)
+  end
+
+  test "can sign in" do
+    post session_url, params: { email: @user.email, password: "password" }
+    assert_redirected_to root_url
+  end
+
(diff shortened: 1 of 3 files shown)


zachgoll (approved):
I agree with this approach.

For some additional context around this—the original logic was put in a while back when we were attempting to "auto-sync" the family if the user hadn't synced for the current day.  This was required to make that work, but I removed it a few weeks back as it was creating too much complexity.

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,18 @@
...
+    post session_url, params: { email: @user.email, password: "password" }
+    assert_redirected_to root_url
+  end
+
+  test "sets last_login_at on successful login" do
+    assert_changes -> { @user.reload.last_login_at }, from: nil do
+      post session_url, params: { email: @user.email, password: "password" }
+    end
+  end
+end
```
zachgoll, line 18: Appreciate you adding these in!

Issue #1019, opened 2024-07-23
Improvement: Format money symbol in correct position


zachgoll:
**Is this issue related to a problem?  Please describe.**

Currently, money formatting does not properly position the currency symbol:

![image](https://github.com/user-attachments/assets/ae783faa-9eb0-422e-98b0-dc4c3d267ad2)

This should be: `0,00 kr`

**Describe the work that needs to be done to address this issue**

We're using the built-in Rails [number_to_currency](https://edgeapi.rubyonrails.org/classes/ActiveSupport/NumberHelper.html#method-i-number_to_currency) helper to format money, which receives some default options provided by the `Money` class:

https://github.com/maybe-finance/maybe/blob/f62c5e43c30ecdeffbe6898be63b1b94c34a5f91/lib/money.rb#L89-L96

I _think_ we should be able to simply add the following option to fix this:

```
default_format_options = {
  # ... 
  format: currency.default_format
}
```

These default options are passed directly to the Rails helper.

PR #1020, opened 2024-07-25
Fix curency format

JuliusMieliauskas:
Fixes issue: #1019

Screenshot from 2024-07-25 11-21-37

As you can see dollar sign is on the left of the number and 'kr' is on the right.


zachgoll (approved):
Awesome!

imist:
Wow! That was quick! :thumbsup:

PR #1024, opened 2024-07-26
Fix: Omit layout for turbo frames with custom sidebar layout

pranavbabu:
This PR folows this hint for turbo frame usage [https://github.com/hotwired/turbo-rails?tab=readme-ov-file#a-note-on-custom-layouts](url) and implements custom method for with sidebar layout. This change will ommit rendering layout when turbo frame request is used. It improves performance by returning much smaller html as example for rendering New account modal with layout server retruns 165 KB (request time: 235ms) of html with this change server returns 1.6 KB (request time: 35ms).

### DIFF
--- a/app/controllers/account/entries_controller.rb
+++ b/app/controllers/account/entries_controller.rb
@@ -1,5 +1,5 @@
 class Account::EntriesController < ApplicationController
-  layout "with_sidebar"
+  layout :with_sidebar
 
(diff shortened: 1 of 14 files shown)


zachgoll (approved):
Nice idea here, especially with the expensive calculations that are needed for our sidebar layout.

PR #1039, opened 2024-08-01
Add security prices provider (Synth integration)

zachgoll:
A basic implementation of security price fetching from our "default" data provider, Synth.

A lot of the code here is intentionally left with some duplication as we implement more and more Synth endpoints and incorporate more providers.  Eventually we'll need to consolidate and refactor provider logic.

Major changes here include:

- Move to using `ticker` rather than `isin` to identify a security internally
- Add `fetch_security_prices` method to `Synth` with pagination support for fetching multiple prices at once (1 Synth credit used per "group" rather than per securities price)
- Incorporate live prices into Holdings sync algorithm

### DIFF
--- a/db/migrate/20240731191344_change_primary_identifier_for_security.rb
+++ b/db/migrate/20240731191344_change_primary_identifier_for_security.rb
@@ -1,0 +1,7 @@
+class ChangePrimaryIdentifierForSecurity < ActiveRecord::Migration[7.2]
+  def change
+    rename_column :securities, :symbol, :ticker
+    remove_column :securities, :isin, :string
+    rename_column :security_prices, :isin, :ticker
+  end
+end
--- a/test/interfaces/exchange_rate_provider_interface_test.rb
+++ b/test/interfaces/exchange_rate_provider_interface_test.rb
@@ -9,7 +9,7 @@ module ExchangeRateProviderInterfaceTest
 
   test "exchange rate provider response contract" do
     VCR.use_cassette "synth/exchange_rate" do
-      response = @subject.fetch_exchange_rate from: "USD", to: "MXN", date: Date.current
+      response = @subject.fetch_exchange_rate from: "USD", to: "MXN", date: Date.iso8601("2024-08-01")
 
       assert_respond_to response, :rate
       assert_respond_to response, :success?
--- a/test/interfaces/exchange_rate_provider_interface_test.rb
+++ b/test/interfaces/exchange_rate_provider_interface_test.rb
@@ -8,7 +8,7 @@ module ExchangeRateProviderInterfaceTest
   end
 
   test "exchange rate provider response contract" do
-    accounting_for_http_calls do
+    VCR.use_cassette "synth/exchange_rate" do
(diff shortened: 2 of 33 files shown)


db/migrate/20240731191344_change_primary_identifier_for_security.rb
```diff
@@ -0,0 +1,7 @@
+class ChangePrimaryIdentifierForSecurity < ActiveRecord::Migration[7.2]
+  def change
+    rename_column :securities, :symbol, :ticker
+    remove_column :securities, :isin, :string
+    rename_column :security_prices, :isin, :ticker
+  end
+end
```
zachgoll, line 7: This is largely a simplification of how we identify securities internally.  My original thinking was that `isin` codes would provide globally unique identification for securities and help us avoid duplicates, but after doing some research on various data providers, this is not viable.

For example, Plaid provides an `isin` code for each security, while Polygon (a popular stock price provider) doesn't provide an isin code at all.  Instead, it provides a FIGI code, which is the new globally unique identifier standard for all types of financial instruments that is not subject to change when a company reclassifies, splits, moves exchanges, etc. (isin can change in rare cases).

While a "ticker" symbol isn't globally unique, it fits our domain model well.  Each ticker represents a company + exchange pair, which results in a unique price.

test/interfaces/exchange_rate_provider_interface_test.rb
```diff
@@ -8,20 +8,13 @@ module ExchangeRateProviderInterfaceTest
...
 
       assert_respond_to response, :rate
       assert_respond_to response, :success?
       assert_respond_to response, :error
       assert_respond_to response, :raw_response
     end
   end
-
-  private
-    def accounting_for_http_calls
```
josefarias, line 22: @zachgoll great work here! In case you're wondering (you may have already figured this out) this was put in place to have a single point of extension to stub responses for all providers, since these tests will eventually be shared across them.

All good removing, and all good to never bring it back. Just realized it may not be obvious so thought I'd point it out.

zachgoll: @josefarias ahh, good callout!  Was this the direction you were thinking?

```rb
def accounting_for_http_calls
  cassettes = [
   { name: 'synth/exchange_rate' },
   { name: 'other/exchange_rate' }
  ]
  
  VCR.use_cassettes(cassettes) do
    yield
  end
end
```

josefarias: Yes! Something like that

db/migrate/20240731191344_change_primary_identifier_for_security.rb
Schyzophrenic: The main issue is that tickers are for specific companies. All ETF and active funds will have an ISIN instead. 
Tickers are useful but ISIN are the standard for looking up values of an investment and are widely available on the Key Investor Information or popular web sites such as morningstar. 
Without the ISIN, how will you add investments which are not companies?

PR #1046, opened 2024-08-02
Show cash + holdings value for investment account view

zachgoll:
https://github.com/user-attachments/assets/01e15430-ea4e-435d-ad8f-8a53f4eb5a25

This is a very rudimentary implementation of showing an investment account's total value by combining the market value of the account's holdings with the cash balance on the account.

There are likely a handful of nuances to this and this will require several follow-ups to solve for all the edge cases while attempting to derive a time series of an investment account's value.  The purpose of this PR is to sketch out a basic implementation to work from.

### DIFF
--- a/app/controllers/account/cashes_controller.rb
+++ b/app/controllers/account/cashes_controller.rb
@@ -1,0 +1,14 @@
+class Account::CashesController < ApplicationController
+  layout :with_sidebar
+
+  before_action :set_account
+
+  def index
+  end
+
+  private
+
+    def set_account
+      @account = Current.family.accounts.find(params[:account_id])
+    end
+end
--- a/app/views/account/entries/entryables/transaction/_transaction.html.erb
+++ b/app/views/account/entries/entryables/transaction/_transaction.html.erb
@@ -1,8 +1,9 @@
 <%# locals: (entry:, selectable: true, editable: true, short: false, show_tags: false, **opts) %>
 <% transaction, account = entry.account_transaction, entry.account %>
+<% is_investment_transfer = entry.account.investment? && entry.transfer.present? %>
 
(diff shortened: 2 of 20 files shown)


app/controllers/account/cashes_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def index
+  end
+
+  private
+
+    def set_account
+      @account = Current.family.accounts.find(params[:account_id])
+    end
+end
```
zachgoll, line 14: This controller is a "specialization" of `Account::HoldingsController`.  The "cash" section of an investment account will show:

1. Brokerage Cash - this is equal to `@account.balance` and represents the "brokerage cash" or "uninvested cash" being held in the investment account.  While brokerage cash is not technically a "holding", I've created a helper to simulate a `Holding` for displaying this row
  a. Sales of stock, dividends, interest accrued, and deposits (transfers from external account) increase brokerage cash
  b. Withdrawals from the account decrease brokerage cash
2. Cash Equivalents - in the future, we may show a list of cash-equivalent holdings here (e.g. USDC, money market funds, etc.)

app/views/account/entries/entryables/transaction/_transaction.html.erb
```diff
@@ -51,6 +52,12 @@
     <% end %>
   </div>
 
+  <% if is_investment_transfer %>
+    <div class="col-span-5 text-right">
+      <%= tag.p entry.inflow? ? t(".deposit") : t(".withdrawal") %>
+    </div>
+  <% end %>
```
zachgoll, line 59: This partial is starting to mix a lot of logic related to the display of transfers.   It will likely need a refactor soon for readability and future flexibility, but leaving it as-is until we've nailed down all of the account views so all requirements are known prior to reworking.

Issue #1037, opened 2024-07-31
Bug: Unique Constraint Violation Error When Creating Transaction in Different Currency


realsama:
I encountered an error when creating a transaction in a different currency than the account. The error message is as follows:

```sh
PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "index_exchange_rates_on_base_converted_date_unique"
DETAIL: Key (from_currency, to_currency, date)=(NGN, USD, 2024-07-31) already exists.
```

Steps to reproduce:

Create a transaction in a currency different from the account currency.
- Observe the error mentioned above.
- It seems like there might be an issue with handling exchange rates or currency conversion. Could you please look into this?

Thank you for your help!

<img width="1321" alt="Screenshot 2024-07-31 at 20 46 02" src="https://github.com/user-attachments/assets/49711777-7ba4-4f5e-8d68-a7e63a37f23e">

osaroadade:
To add more context to this:

If you add a new transaction, this error would leave and currency syncing would continue.

zachgoll:
Done via #1069

PR #1069, opened 2024-08-08
Fetch exchange rates in bulk from synth

### LINKED ISSUE #1047: Make bulk request for exchange rates to Synth API
**Describe the bug**
See title.

**To Reproduce**
1. Set up project without synth API key first
2. Try to convert a currency on the App and get the expected error message
3. Then adding API key + restart server
4. Open the web app
5. Noticed the 700+ API calls

**Screenshots / Recordings**
image

tonyvince:
Fix #1047


app/models/provider/synth.rb
```diff
@@ -57,12 +57,40 @@ def fetch_exchange_rate(from:, to:, date:)
     end
   end
 
+  def fetch_exchange_rate_for_date_range(from:, to:, date_start:, date_end:)
```
zachgoll, line 60: Mostly style here, but could we keep this consistent with the `fetch_security_prices` method stub?
```suggestion
  def fetch_exchange_rates(from:, to:, start_date:, end_date:)
```

app/models/exchange_rate/provided.rb
```diff
@@ -6,12 +6,31 @@ module ExchangeRate::Provided
...
+            rate = ExchangeRate.new \
+              from_currency: from,
+              to_currency: to,
+              date: exchange_rate.dig(:date).to_date,
+              rate: exchange_rate.dig(:rate)
+
+            rate.save! if cache
+            rate
+          rescue ActiveRecord::RecordNotUnique
+            next
```
zachgoll, line 29: I think this should fix #1037 as well correct?

zachgoll (commented):
Awesome work here!

app/models/exchange_rate/provided.rb
tonyvince: Yes!

app/models/provider/synth.rb
tonyvince: Done.

PR #1075, opened 2024-08-09
Deposit, Withdrawal, and Interest Transactions for Investment View

zachgoll:
Initial concept of a dynamic trade / transaction form.  

`Account::EntryBuilder` is the entrypoint where a flattened params hash can be passed in with a `type` and then the relevant trade, transaction, or transfer is created based on this type.

Likely will need to rework some of this as we add more types, but wanted to get this in so users can start building up portfolios.

### DIFF
--- a/app/javascript/controllers/trade_form_controller.js
+++ b/app/javascript/controllers/trade_form_controller.js
@@ -1,0 +1,64 @@
+import {Controller} from "@hotwired/stimulus"
+
+const TRADE_TYPES = {
+  BUY: "buy",
+  SELL: "sell",
+  TRANSFER_IN: "transfer_in",
+  TRANSFER_OUT: "transfer_out",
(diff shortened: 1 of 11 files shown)


app/javascript/controllers/trade_form_controller.js
```diff
@@ -0,0 +1,64 @@
...
+  get fieldTargets() {
+    return {
+      ticker: this.tickerInputTarget,
+      amount: this.amountInputTarget,
+      transferAccount: this.transferAccountInputTarget,
+      qty: this.qtyInputTarget,
+      price: this.priceInputTarget
+    }
+  }
+}
```
zachgoll, line 64: This is a very rudimentary controller, highly coupled to the trade forms.  Ideally we'll abstract this a bit in the future

PR #1076, opened 2024-08-10
Refactor: Use native error i18n lookup

pedrocarmona:
This pull request refactors the existing error handling system to implement native internationalization (i18n) for error messages.

Rails [guides](https://guides.rubyonrails.org/i18n.html#error-message-scopes) documents the lookup order for a validation of name is not empty on the user model:
```
activerecord.errors.models.user.attributes.name.blank
activerecord.errors.models.user.blank
activerecord.errors.messages.blank
errors.attributes.name.blank
errors.messages.blank
```

I had to add magic comment hints for [i18n-tasks to recognized](https://github.com/glebm/i18n-tasks?tab=readme-ov-file#fine-tuning) that this translations are used.
I have not perfomed changes in lib/money.rb because i18n-tasks would not peek into that file.

### DIFF
--- a/app/controllers/settings/profiles_controller.rb
+++ b/app/controllers/settings/profiles_controller.rb
@@ -17,7 +17,7 @@ class Settings::ProfilesController < SettingsController
     if Current.user.update(user_params_with_family)
       redirect_to settings_profile_path, notice: t(".success")
     else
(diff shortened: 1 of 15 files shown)


zachgoll (approved):
Really nice work here, thanks for going through and updating this!

PR #1085, opened 2024-08-13
Fix: i18n symbol typo

pedrocarmona:
Hi @zachgoll,

I missed a colon in the previous PR #1076. 
Now I added tests to that part. 
Maybe the value validation could also be dropped because, as we can see the the tests, the trend validation comes first, so the only way to test that missing colon is to mock trend. On the other hand, code could change in the future and that validation does not hurt.

Sorry and thank you,
Pedro


zachgoll (commented):
No worries, thanks for the update. 

I agree the validation could be dropped.  There are a few areas of TimeSeries and Trend classes that I think we could rework a bit; just haven't gotten around to it due to higher priorities

PR #1086, opened 2024-08-13
Fix for invalid accountable data

zachgoll:
A non-destructive, one-time fix for self hosters who currently have invalid account data that is causing the app to crash


zachgoll:
@jakubkottnauer @jermorin this look / run okay for you?

zachgoll:
@jermorin awesome, thanks for confirming!

PR #1095, opened 2024-08-16
add pagination to account transactions list

code-constructor:
The transaction view in account/transactions had no pagination. This caused a problem where a large list of transactions took a long time to load.


zachgoll (approved):
Nice, thanks for the addition!

PR #1096, opened 2024-08-16
Add support for different column separator in csv import logic

code-constructor:
As we all know, CSV is not always "comma separated values". It is often separated by other characters such as ";" or "|". This PR opens the model layer to support multiple column separators. The implementation is now able to import files with "," and ";" separators. More may come.

Important to know: The normalized representation `normalized_csv_str` is still separated by comma (",").

I would also like to implement the view support. But I don't want to do anything without a little guidance on what would be a good solution. Maybe just a selector below the upload field or a hidden section with "advanced import options".


zachgoll:
@code-constructor I think this all looks good and I'm definitely supportive of adding it in since a lot of non-US imports will have `;` as the separator.

In terms of the UI, what do you think of grabbing this information on the very first step?  

CleanShot 2024-08-16 at 11 47 17@2x

code-constructor:
Did an update of the Pull-Request. I've implemented your idea :+1:. Thanks for the really fast response. Was a pleasure.

app/views/imports/_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= styled_form_with model: @import do |form| %>
   <div class="mb-4">
```
zachgoll, line 2:
```suggestion
  <div class="mb-4 space-y-3">
```

zachgoll (commented):
Awesome, very nice addition!  Just added a little suggestion to fix some form spacing.  Otherwise looks and works great.

app/views/imports/_form.html.erb
zachgoll: CleanShot 2024-08-16 at 12 39 22@2x

code-constructor:
Committed 👍

PR #1099, opened 2024-08-17
Refactor: Allow other import files

pedrocarmona:
This PR does not change existing functionality.

It refactors the code to be more generic: renaming `raw_csv_str` to `raw_file_str` allows for future different filetypes (json, pdf, etc)

Additionally, it also fixes the click to open the system file picker, when clicking in the dropzone of imports.


zachgoll (approved):
Nice, I think these changes make sense!

PR #1111, opened 2024-08-20
Fix file selector opening twice

### LINKED ISSUE #1110: Bug:
**Describe the bug**
The transactions file upload menu opens a second time when clicking cancel

**To Reproduce**
- Go to import transactions, CSV file upload page
- Click on the file drag and drop zone. File selector opens
- Click cancel

**Actual behavior**
- File selector dialogue closes and a new file selector menu appears 

**Expected behavior**
- File selector dialogue closes and a new one does not appear

tonyvince:
Close #1110

### DIFF
--- a/app/javascript/controllers/import_upload_controller.js
+++ b/app/javascript/controllers/import_upload_controller.js
@@ -45,6 +45,7 @@ export default class extends Controller {
   }
 
   click() {
+    this.preventDefault();
     this.inputTarget.click();
   }
 
--- a/app/javascript/controllers/import_upload_controller.js
+++ b/app/javascript/controllers/import_upload_controller.js
@@ -44,8 +44,7 @@ export default class extends Controller {
     }
   }
 
-  click() {
-    this.preventDefault();
+  openFilePicker() {
     this.inputTarget.click();
   }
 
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/import_upload_controller.js
```diff
@@ -45,6 +45,7 @@ export default class extends Controller {
   }
 
   click() {
+    this.preventDefault();
     this.inputTarget.click();
   }
```
zachgoll, line 50: I missed this in a prior PR, but I think rather than preventing the default behavior here, we should just rename this method to something that doesn't conflict with the global namespace per the docs:

https://stimulus.hotwired.dev/reference/actions#naming-conventions

tonyvince: [c9b8a49](https://github.com/maybe-finance/maybe/pull/1111/commits/c9b8a49f1aadc057d961ff71756f8f354fff048d)

zachgoll: I may be missing something here, but are we using this method anywhere now?  The import flow seems to work fine for me without this at all.

zachgoll:
@tonyvince I'd like to get this in for today's release, so I've opened #1119 and have credited you there with my suggestion above.  Tried pushing to this branch but was not able to due to permissions.  Will close this out.

PR #1113, opened 2024-08-20
Fix holding name error

zachgoll:


Procfile.dev
```diff
@@ -1,3 +1,3 @@
-web: bin/rails server -b 0.0.0.0
+web: ${DEBUG:+rdbg -O -n -c --} bin/rails server -b 0.0.0.0
```
zachgoll, line 1: Optionally, run `DEBUG=true bin/dev` to start the dev server with a debugging port open, which VSCode (or another editor) can use to hook into the `debug` gem (installed with project already)

For example, with Ruby LSP plugin in VSCode, use this launch setting to attach to the running dev process:

```json
    {
      "type": "ruby_lsp",
      "name": "Attach debugger",
      "request": "attach"
    }
```

PR #1117, opened 2024-08-23
Basic Vehicle View

zachgoll:
This PR introduces a basic Vehicle account page.  A user can fill in property details such as:

- Make + Model
- Car year
- Mileage

We will use this information in the future to automatically estimate vehicle values from data aggregators.

https://github.com/user-attachments/assets/a2c8ed75-bea4-49fb-9a53-fec2817d7b18

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -29,6 +29,8 @@ module AccountsHelper
     case account.accountable_type
     when "Property"
       properties_path
+    when "Vehicle"
+      vehicles_path
     else
       accounts_path
     end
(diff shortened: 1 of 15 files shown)


app/helpers/accounts_helper.rb
```diff
@@ -38,6 +40,8 @@ def edit_account_form_url(account)
     case account.accountable_type
     when "Property"
       property_path(account)
+    when "Vehicle"
+      vehicle_path(account)
```
zachgoll, line 44: We will eventually not need this.  This is an interim helper so that while we're developing each view, we don't have to do one big reorg of files and views.

PR #1118, opened 2024-08-23
Rubocop updates

zachgoll:
No functional code changes here; just formatting and linting due to changing a few Rubocop rules and running `rubocop -a` on the codebase.

- Added `.editorconfig` with a minimal set of rules (no tabs, 2 spaces)
- Added some Rubocop rules around indentation

In general, we'll stay away from adding too many Rubocop rules, but enforcing 2 spaces is helpful

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,12 +1,12 @@
-# Omakase Ruby styling for Rails
-inherit_gem: { rubocop-rails-omakase: rubocop.yml }
+inherit_gem:
+  rubocop-rails-omakase: rubocop.yml
+  
+Layout/IndentationWidth:
+  Enabled: true
 
-# Overwrite or add rules to create your own house style
-#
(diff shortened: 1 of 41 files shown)


.rubocop.yml
```diff
@@ -1,12 +1,15 @@
...
\ No newline at end of file
+Layout/IndentationStyle:
+  EnforcedStyle: spaces
+  IndentationWidth: 2
+
+Layout/IndentationConsistency:
+  Enabled: true
+
+Layout/SpaceInsidePercentLiteralDelimiters:
+  Enabled: true
```
zachgoll, line 15: The Rails omakase config disables this one and prefers `%i[ items with space ]`.

That said, there is no Cop rule that can enforce this.  Given the OSS nature of this project, consistency is the most important thing; hence adding this rule in.

PR #1120, opened 2024-08-23
Add ability to import pdfs from financial institutions

pedrocarmona:
Adds ability to parse pdf files from banks and credit cards

- Is behind a feature flag because its not clear that handling pdfs would be safe
- Also storing regexes from users into database is not that safe.
- Added a requirement: active record encryption to be set. Because of potencial PII in pdfs

Anyway, it looks like this:

### DIFF
--- a/test/fixtures/pdf_regexes.yml
+++ b/test/fixtures/pdf_regexes.yml
@@ -1,0 +1,7 @@
+awesome_bank_pdf_regex:
+  family: dylan_family
+  name: Awesome Bank Pdf Regex
+  transaction_line_regex_str: '(?<date>\w{3}\s\d\d?\,\s\d{4})\s+(?<name>.+?(?=\s{4}\s*))\s+\S(?<amount>(\d?\d?\d\,)*(\d?\d?\d\.\d\d))\s+\S(?<balance>(\d?\d?\d\,)*(\d?\d?\d\.\d\d))'
+  metadata_regex_str: '.*Account transactions from (?<start_date>\w+\s\d\d?\,\s\d{4}) to (?<end_date>\w+\s\d\d?\,\s\d{4}).*'
+  pdf_transaction_date_format: "%b %-d, %Y"
+  pdf_range_date_format: "%B %-d,%Y"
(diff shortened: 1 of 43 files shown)


test/fixtures/pdf_regexes.yml
```diff
@@ -0,0 +1,7 @@
+awesome_bank_pdf_regex:
```
pedrocarmona, line 1: when testing in development you can use this regexes + the transactions.pdf located in the fixtures folder

zachgoll (commented):
@pedrocarmona thanks for this submission, really awesome work here!

I read through the code and while this is definitely a path we might explore at some point, I think we're a little _early_ to be introducing this to the product.

Below I've left some rationale behind why I think we need to wait on this:

**Roadmap / Product Management**

In general, our highest priority right now from a roadmap standpoint is to reach stability for the product and release our first hosted version of the application.  Because of this, we have to be very mindful of the features we're adding and the complexity we're introducing since we (as a company) will have to fully support them on our hosted version.  

While a PDF import feature is definitely something we'll want to explore down the road, we'll need to dedicate quite a bit of time and resources to ensure that the feature is well tested and ready for general release in our hosted version.  

While we're still working through the early days of this, we've put together some details on [project management here](https://github.com/maybe-finance/maybe/wiki/Project-Management-and-Release-Cycle) which gives a bit of insight into how we're thinking about introducing and developing new features like this.

**Plugin ecosystem**

Eventually, I think we'll need to introduce some sort of plugin ecosystem to the self hosted version of Maybe which would enable contributions like this to be 100% self contained and available for users who would like to "opt-in" to them.  

This would be a great candidate for a "plugin" since many users would be overwhelmed by adding regex's to their imports.  This falls under the "superuser" category of features which we'll want to support, but also keep isolated (via plugins) from the core functionality that we offer on the hosted version.

Hopefully this gives a bit of insight behind how we're thinking about all this.  Let me know if you have any questions!

pedrocarmona:
Hi @zachgoll,

Thank you for taking the time to review my pull request, and detailed response.

I agree that many users would be overwhelmed by Regular Expressions.
I look forward for the "Plugin ecosystem".

Best regards,
Pedro

PR #1121, opened 2024-08-24
Run PostgreSQL in a local docker container

dgrcode:
This PR adds minimal Docker setup to get PostgreSQL running inside a container.

The goal is to simplify the setup process, so anyone can do `make local`, `bin/setup`, `bin/dev` and get started with any code changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,15 @@
+.PHONY: .setup
+
+COMPOSE_PROJECT_NAME ?= maybe
+LOCAL_DETACHED ?= true
+
+local: .setup
+	COMPOSE_PROJECT_NAME=$(COMPOSE_PROJECT_NAME) \
+		docker compose -f docker/docker-compose-local.yml up $(if $(filter true,$(LOCAL_DETACHED)),-d)
+
+stop_local:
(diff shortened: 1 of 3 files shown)


zachgoll:
@dgrcode thanks for the submission!  Definitely can see how this would benefit some cross-platform devs, but I think in general, we're pretty happy with our local dev setup documentation and options.

We typically recommend Devcontainers for anyone looking to develop locally with Docker.  I know this is a slightly more agnostic method for spinning up the app, but since we have lots of self hosting users who aren't developers and who reference the `docker-compose.example.yml` file in the root of this repo, I'm afraid adding another compose file might cause some confusion among this group of users.

Would be happy to add a note/link in the wiki to this PR for users looking to automate their setup a bit more though!

dgrcode:
Thanks for the fast reply! I agree it could be confusing for some users.

I've never used Devcontainers before and completely missed the `.devcontainer` folder, so I added this docker configuration to develop locally. I thought it could be useful to contribute, but from what I see Devcontainers should remove the need for locally running containers. I'll give it a go some time this week and see how it goes.

PR #1125, opened 2024-08-25
Fix unable to create Deposit entries in investment portfolio

### LINKED ISSUE #1123: Bug: Depositing into Investment Account Requires Transfer Account, but should be Optional
**Describe the bug**
When submitting a deposit-type transaction for an investment account, a deposit without an associated transfer account will error and not create the transaction, even though the UI says that the transfer account field in the form is optional.

**To Reproduce**
1. Create an account and an investment portfolio.
2. Click the investment portfolio, then click "Transactions" in the portfolio view.
3. Click "New Transaction"
4. Select "Deposit" for the type
5. Enter in any date and amount, but leave the transfer account field unchanged.
6. Click "Add Transaction." The page will reload, but the transaction will not be added. If you open browser devtools on the network tab, you will see the request returned a "422 Unprocessable Content" error.

**Expected behavior**
The deposit transaction should create without the user required to select a transfer account. Or, if the behavior is correct and the UI is wrong, the "transfer account" field should not be marked as optional.

**Screenshots / Recordings**
Example deposit that does not work:

Deposit View

The error in question:

Deposit Error

**Additional context**
Being able to deposit without a transfer account is an important feature. I'm personally using this for my 401(k), which gets deposited from my paycheck rather than from an account. The only way I can make this work is either adding the money in a different portfolio type, like a Depository, and then transferring from there, or by creating the transaction as "interest" rather than "deposit," neither of which are ideal.

tonyvince:
Fix #1123

### DIFF
--- a/app/models/account/transaction_builder.rb
+++ b/app/models/account/transaction_builder.rb
@@ -21,7 +21,7 @@ class Account::TransactionBuilder
     end
 
     def create_transfer
-      return create_unlinked_transfer(account.id, signed_amount) unless transfer_account_id
+      return create_unlinked_transfer(account.id, signed_amount) if transfer_account_id.blank?
 
       from_account_id = type == "transfer_in" ? transfer_account_id : account.id
       to_account_id = type == "transfer_in" ? account.id : transfer_account_id
--- a/test/system/transactions_test.rb
+++ b/test/system/transactions_test.rb
@@ -150,6 +150,24 @@ class TransactionsTest < ApplicationSystemTestCase
     assert_selection_count(0)
   end
 
+
+  test "can create deposit transaction for investment account" do
+    investment_account = accounts(:investment)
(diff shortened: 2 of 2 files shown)


app/models/account/transaction_builder.rb
```diff
@@ -21,7 +21,7 @@ def transfer?
     end
 
     def create_transfer
-      return create_unlinked_transfer(account.id, signed_amount) unless transfer_account_id
+      return create_unlinked_transfer(account.id, signed_amount) if transfer_account_id.blank?
```
tonyvince, line 24: in this case the value of `transfer_account_id` is an empty string ""

zachgoll: nice catch!

zachgoll (approved):
Thanks for the quick fix on this!

PR #1133, opened 2024-08-26
Fix account transaction form resetting amount to 0

### LINKED ISSUE #1132: Bug: Transactions update to zero
**Describe the bug**
When editing a transaction, it will update the value to zero.
Specially if there is text in the Notes section.

**To Reproduce**
1. Go to click any transaction
2. Type something into Notes
3. It closes the side bar abruptly and set the transaction value to zero

**Screenshots / Recordings**
[screen capture.webm](https://github.com/user-attachments/assets/92359a6d-b213-48c7-9774-e291b7b0fb0a)

**Additional context**
I'm self hosting and I've been using Maybe for a while, and this behaviour was introduced in version 13.
I'm not sure if i'm the only one with this bug, or more people have it.

zachgoll:
Fixes #1132

### DIFF
--- a/app/controllers/account/transactions_controller.rb
+++ b/app/controllers/account/transactions_controller.rb
@@ -12,8 +12,7 @@ class Account::TransactionsController < ApplicationController
   end
 
   def update
-    @entry.update!(entry_params.merge(amount: amount))
-    @entry.sync_account_later
+    @entry.update!(entry_params)
 
     respond_to do |format|
       format.html { redirect_to account_entry_path(@account, @entry), notice: t(".success") }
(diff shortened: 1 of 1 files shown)


app/controllers/account/transactions_controller.rb
```diff
@@ -12,8 +12,7 @@ def index
   end
 
   def update
-    @entry.update!(entry_params.merge(amount: amount))
-    @entry.sync_account_later
```
zachgoll, line 16: Removed the `sync_later` because this triggers a refresh of the entire page (via the Sync broadcasts) which gets annoying for updating transaction details.  This will cause out-of-sync data if the user changes the amount or date, but will be updated on the next sync.

Eventually we'll want to get more granular with _when_ we trigger the re-sync, but for now this feels like a decent tradeoff.

app/controllers/account/transactions_controller.rb
```diff
@@ -12,8 +12,7 @@ def index
   end
 
   def update
-    @entry.update!(entry_params.merge(amount: amount))
```
zachgoll, line 15: Merging in `amount` caused the scenario when we didn't include the param, it was calling `nil.to_d`, which resulted in 0

PR #1148, opened 2024-09-02
Do not show registration link when REQUIRE_INVITE_CODE=true

### LINKED ISSUE #1140: Bug: Option to disable register link on landing page
Hello,

**Description**
It would be nice to add an option in settings to disable the register link `...create an account` on the landing page. This will ensure there are no other account on our self-hosted website while it being public and will greatly improve the security of the budget app and make it usable only by ourselves.

**Expected behavior**
To be able to toggle `...create an account` on and off depending on the needs.

tonyvince:
closes #1140


zachgoll (approved):
Good idea!

PR #1167, opened 2024-09-10
Fix: When transaction drawer closed, turbo frame renders below main content

### LINKED ISSUE #1165: Bug: When transaction drawer closed, turbo frame renders below main content
**Describe the bug**

https://github.com/user-attachments/assets/9883e2ef-9274-4cf4-82bb-759c304937e4

**Expected behavior**

- When a drawer is closed, no content should show anymore

jestinjoshi:
Fixes #1165 

Explanation: HTML dialog should not be given CSS display property as it must be handled by the browser for hiding/showing the dialog


zachgoll (approved):
Good catch!

PR #1175, opened 2024-09-13
Support multi-currency transfers

### LINKED ISSUE #1139: Bug: Currency Inconsistency in Transfer Transactions
**Describe the bug**
When initiating a transfer between two EUR accounts, the system fails to infer or allow the selection of the transaction currency correctly. Instead, it defaults to USD, causing incorrect deductions from non-USD accounts.

**To Reproduce**
1. Navigate to the Transactions section
2. Click on "Add Transactions"
3. Select "Transfer" as the transaction type
4. Observe that the amount field does not have a currency selection option
5. Attempt to transfer any amount from a non-USD account (e.g., EUR account)

**Expected behavior**
The system should either:
1. Automatically infer and use the transaction currency based on the source account's currency, or
2. Provide an option for the user to select the transaction currency

**Additional Notes**
- This issue appears to affect all non-USD accounts, not just EUR accounts
- The problem is consistent across all transfer attempts from non-USD accounts

zachgoll:
Fixes #1139

### DIFF
--- a/app/models/account/transfer.rb
+++ b/app/models/account/transfer.rb
@@ -46,7 +46,7 @@ class Account::Transfer < ApplicationRecord
     def build_from_accounts(from_account, to_account, date:, amount:, currency:, name:)
       outflow = from_account.entries.build \
         amount: amount.abs,
-        currency: currency,
+        currency: from_account.currency,
         date: date,
         name: name,
         marked_as_transfer: true,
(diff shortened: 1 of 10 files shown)


app/models/account/transfer.rb
```diff
@@ -72,27 +72,23 @@ def single_currency_transfer?
 
     def transaction_count
       unless entries.size == 2
-        # i18n-tasks-use t('activerecord.errors.models.account/transfer.attributes.entries.must_have_exactly_2_entries')
```
zachgoll, line 75: These introduce too much maintenance overhead at this stage.  Removing and ignoring.

PR #1179, opened 2024-09-14
Add error handling for vehicle and property account creation

### LINKED ISSUE #1178: Bug: Blank Current Balance in New Property or Vehicle Causing Crashes
**Describe the bug**
When creating a new Property or Vehicle, if the 'Current Balance' field is left blank, the app crashes. The blank entry is not being handled properly.

**To Reproduce**

1. Open add new Property or add new Vehicle form.
2. Leave the 'Current Balance' field empty.
3. Click on Add.

**Expected behavior**

image

**Screenshots / Recordings**

image

tonyvince:
Close #1178

### DIFF
--- a/app/helpers/forms_helper.rb
+++ b/app/helpers/forms_helper.rb
@@ -34,7 +34,8 @@ module FormsHelper
       default_currency: options[:default_currency] || "USD",
       disable_currency: options[:disable_currency] || false,
       hide_currency: options[:hide_currency] || false,
-      label:         options[:label] || "Amount"
+      label:         options[:label] || "Amount",
+      required:      options[:required] || false
     }
   end
 
--- a/test/controllers/properties_controller_test.rb
+++ b/test/controllers/properties_controller_test.rb
@@ -46,33 +46,6 @@ class PropertiesControllerTest < ActionDispatch::IntegrationTest
     assert_enqueued_with(job: AccountSyncJob)
   end
 
-  test "handles erors when creating property" do
-    assert_no_difference [ "Account.count", "Property.count", "Account::Valuation.count", "Account::Entry.count" ] do
-      post properties_path, params: {
-        account: {
-          name: "Property",
-          currency: "USD",
-          accountable_type: "Property",
-          accountable_attributes: {
-            year_built: 2002,
-            area_value: 1000,
-            area_unit: "sqft",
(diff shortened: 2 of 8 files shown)


zachgoll (commented):
Could this be solved easier if we just apply client-side form validations?

I think we can add a `required` attribute on these fields and let the browser do all of this for us.

test/controllers/properties_controller_test.rb
```diff
@@ -46,6 +46,33 @@ class PropertiesControllerTest < ActionDispatch::IntegrationTest
...
+              country: "US",
+              postal_code: "90001"
+            }
+          }
+        }
+      }
+      assert_redirected_to accounts_path
+      assert_equal "Balance can't be blank and Entries is invalid", flash[:alert]
+    end
+  end
```
zachgoll, line 74: I think we can still remove all of these tests alongside the `rescue` handlers in the controller. 

The form validation should handle all the main requirements directly in the browser.  If a user somehow finds a way to side-step these, I'd rather the system raise an error so that we can capture it in observability systems like Sentry (rather than the user seeing an error message and the system acting as if nothing went wrong).

Following this logic, I think our existing `create` action in `accounts_controller` should probably _not_ have the `rescue` block either (as I'm sure that was the pattern you followed here).  In other words, we could probably remove this as well:

https://github.com/maybe-finance/maybe/blob/fd4011126422023228a9295b2d9b30bb1b9402cb/app/controllers/accounts_controller.rb#L61-L62

app/helpers/forms_helper.rb
```diff
@@ -34,7 +34,8 @@ def money_with_currency_field(form, money_method, options = {})
       default_currency: options[:default_currency] || "USD",
       disable_currency: options[:disable_currency] || false,
       hide_currency: options[:hide_currency] || false,
-      label:         options[:label] || "Amount"
+      label:         options[:label] || "Amount",
+      required:      options[:required] || false
```
zachgoll, line 38: We'll need to update this improper usage for this to flow through correctly:

https://github.com/maybe-finance/maybe/blob/fd4011126422023228a9295b2d9b30bb1b9402cb/app/views/accounts/_form.html.erb#L8

See `required: "required"` should be -> `required: true`

zachgoll (approved):

PR #1189, opened 2024-09-16
Finish remaining transaction filters

zachgoll:
We introduced transaction filtering a while back, but a few of the filters were not ready to be implemented.

This PR completes the remaining "coming soon" transaction filters so users have full control over their transactions dashboard.

### DIFF
--- a/app/models/account/entry.rb
+++ b/app/models/account/entry.rb
@@ -158,6 +158,17 @@ class Account::Entry < ApplicationRecord
         end
       end
 
+      if params[:amount].present? && params[:amount_operator].present?
+        case params[:amount_operator]
+        when "equal"
+          query = query.where("ABS(ABS(account_entries.amount) - ?) <= 0.01", params[:amount].to_f.abs)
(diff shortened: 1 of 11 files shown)


app/models/account/entry.rb
```diff
@@ -148,6 +148,27 @@ def search(params)
...
+      if params[:amount].present? && params[:amount_operator].present?
+        case params[:amount_operator]
+        when "equal"
+          query = query.where("ABS(ABS(account_entries.amount) - ?) <= 0.01", params[:amount].to_f.abs)
+        when "less"
+          query = query.where("ABS(account_entries.amount) < ?", params[:amount].to_f.abs)
+        when "greater"
+          query = query.where("ABS(account_entries.amount) > ?", params[:amount].to_f.abs)
+        end
+      end
```
zachgoll, line 170: If we have to add more filters in the future, this is definitely a refactor candidate, but given this is the final set of filters for a while, no need right now.

PR #1195, opened 2024-09-20
Refactor CSV Import Controller (#1192)

pedrocarmona:
Improves usability and responsiveness on the clean step of importing. 
Also, uses a json column instead of a normalized csv for faster updates.

## Before: Each change triggers a full page refresh

https://github.com/user-attachments/assets/60efc358-bdb1-4f82-8def-6b748b2a2796

## After: Each change triggers a turbo request, the cursor stays in the next cell (no page refresh)

https://github.com/user-attachments/assets/adbd8fe2-d43f-45df-9b24-205d054ab0af

Details:

- Create a new database import_rows:
  - that references import_id,
  - has an index row index within import,
  - contains a json column fields.
- Add a new model (Import::Row).
  - use active record store to programatically set any number of columns (:date, :name, :category, :tags, :amount)
  - set validations: should validate required fields, and data types (dates, bigdecimal)
- Move updating a cell into own controller (Import::RowsController)
  - change views to call new controller using turbo stream format
  - on update, run validations on the cell, and update cell validation status
  - on update, if all validations are passing, then should present a next button, to allow to step into the confirmation step
- Also some small changes to column mappings


zachgoll:
@pedrocarmona I see some changes coming in so I'll hold off on any comments for now.

When this is ready for review, feel free to request one directly from me and I'll take a look!

pedrocarmona:
Hi @zachgoll, took me a while, I hope its good now...
I am not able to request review using github ui (maybe its restricted to maintainers).

Thank you for reviewing!

Pedro

app/models/import.rb
```diff
@@ -1,14 +1,20 @@
...
   validate :raw_file_must_be_parsable
   validates :col_sep, inclusion: { in: Csv::COL_SEP_LIST }
 
-  before_save :initialize_csv, if: :should_initialize_csv?
+  before_save :recreate_rows, if: :should_recreate_rows?
 
   enum :status, { pending: "pending", complete: "complete", importing: "importing", failed: "failed" }, validate: true
 
-  store_accessor :column_mappings, :define_column_mapping_keys
+  store :column_mappings, accessors: FIELDS, suffix: :column_mapping
```
zachgoll, line 17: I think we'll want to leverage `store_accessor` here since we're using Postgres.  This allows us to offload all the serialization/deserialization to the DB rather than doing it in Rails.  

From the docs:

> NOTE: If you are using structured database data types (e.g. PostgreSQL hstore/json, or MySQL 5.7+ json) there is no need for the serialization provided by [.store](https://api.rubyonrails.org/classes/ActiveRecord/Store/ClassMethods.html#method-i-store). Simply use [.store_accessor](https://api.rubyonrails.org/classes/ActiveRecord/Store/ClassMethods.html#method-i-store_accessor) instead to generate the accessor methods. Be aware that these columns use a string keyed hash and do not allow access using a symbol.

https://api.rubyonrails.org/classes/ActiveRecord/Store.html

app/controllers/import/rows_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  before_action :set_import
+  before_action :set_row_data, only: %i[update]
+
+  def update
+    @row.assign_attributes(row_params)
+    @row.save(validate: false)
+
+    respond_to do |format|
+      format.html { redirect_to clean_import_path(@import) }
+      format.turbo_stream { render "imports/rows/update" }
```
zachgoll, line 11: I'm a bit hesitant to rely solely on Turbo streams to change the behavior of the update here.  I think we're putting too much responsibility in the stream itself and veering away from the [Turbo conventions](https://turbo.hotwired.dev/handbook/streams#progressively-enhance-when-necessary):

> It’s good practice to start your interaction design without Turbo Streams. Make the entire application work as it would if Turbo Streams were not available, then layer them on as a level-up.

Alternatively, what I think we could do is:

- Create an endpoint that returns an individual cell's data 
- Each cell in the form will be its own turbo frame
- On update, we redirect to the controller endpoint that renders the individual cell data

I think this will allow us to reduce the amount of code here and create a more straightforward flow.

zachgoll (commented):
Good work here!  

Overall, I 100% agree with your idea to split the updates to each row to its own model/controller.  This is a much easier model to reason about.

Like I mentioned in the [discussion](https://github.com/maybe-finance/maybe/discussions/1192#discussioncomment-10698644), I've got a pretty big overhaul coming in the next few days on the imports feature, which makes quite a few changes around all the column mappings and structure of the feature.  

I'm going to wait to merge this until I've got my PR open and we can see how much overlap there is between the two.  I'll let you know when it is open so you can review it and we can decide if/how we can incorporate this PR.

zachgoll:
@pedrocarmona so I did quite a bit of a refactor (more than I had originally thought) in #1209 which uses a "Rows" based approach and significantly extends the ability to configure imports.  Given how many changes are in #1209 I unfortunately think it's best we close this out as I don't see a good way to integrate.

If you have any suggestions around my implementation given the work you did here, let me know and I will take a look.

PR #1210, opened 2024-09-27
Fix: Escape button not being handled on settings pages

jestinjoshi:
### Describe the bug:
When you click `Escape` key while on settings page nothing happens

### To Reproduce:
1. Open Settings
2. Press `Escape` key

### Expected Behavior:
Settings should close on `Escape` key press.

### Alternate Solution:
Currently the `Escape` key takes you back one page in history. So if you are on another page in Settings, pressing `Escape` will simply take you back to the previous page and not outside Settings. Using `window.history.go()` or `window.history.pushState()` might be a better way to handle this scenario.

### DIFF
--- a/app/views/settings/_nav.html.erb
+++ b/app/views/settings/_nav.html.erb
@@ -4,9 +4,9 @@
       <%= lucide_icon "chevron-left", class: "w-5 h-5 text-gray-500" %>
       <span>Back</span>
     <% end %>
-    <span data-controller="hotkey" data-hotkey="Escape" data-action="hotkey#navigateBack" class="uppercase bg-gray-100 rounded-sm px-1 py-0.5 text-xs text-gray-500 shadow-sm ml-1">
+    <button data-controller="hotkey" data-hotkey="Escape" data-action="hotkey#navigateBack" class="uppercase bg-gray-100 rounded-sm px-1 py-0.5 text-xs text-gray-500 shadow-sm ml-1 pointer-events-none">
       esc
-    </span>
+    </button>
   </div>
   <nav class="space-y-4">
     <section class="space-y-2">


zachgoll:
@jestinjoshi thanks for diving in and finding this!  Definitely a valuable fix to add.

I think what we could do here is build a super basic controller concern that stores the "return to" path in a server-side session, and then both the back button and the ESC key will use this to know where to return to.

So we'd have something like this included in `ApplicationController`:

```rb
class ApplicationController < ActionController::Base
  include StoreLocation
end
```

```rb
module StoreLocation
  extend ActiveSupport::Concern

  included do
    helper_method :previous_path
    before_action :store_return_to
    after_action :clear_previous_path
  end

  def previous_path
    session[:return_to] || fallback_path
  end

  private

    def store_return_to
      if params[:return_to].present?
        session[:return_to] = params[:return_to]
      end
    end

    def clear_previous_path
      if request.fullpath == session[:return_to]
        session.delete(:return_to)
      end
    end

    def fallback_path
      root_path
    end
end
```

Then in our navbar, we'd trigger this by passing the return path:

```erb
<%= link_to "Go to settings (sample link)", settings_profile_path(return_to: request.fullpath) %>
```

And finally, our ESC will use the provided helper:

```erb
<%= link_to "Back", previous_path %>
```

jestinjoshi:
@zachgoll the changes you suggested worked like a charm! Now, we can return to the exact page we navigated from instead of always going back to the homepage. Also, no matter where you are in the settings page, clicking the back button or pressing escape exits the settings entirely, which is a great touch.

However, I noticed that when entering the settings page, a `return_to` query param gets added. This seems to be the reason why the CI test is failing. I think we'll need to refactor the test. Let me know what you think.

zachgoll:
@jestinjoshi awesome!  And yes, I agree with you.  Let's refactor that test to only check the URL path (excluding query params) which should fix it!

zachgoll (approved):
Great work!  Works well on my end.

PR #1216, opened 2024-09-30
Sort currencies by name as a second order

alagos:
The currency sorting is very chaotic only relying on the "priority"
Screenshot 2024-09-30 at 6 30 10 PM
so I've also set the "name" as a second order parameter
Screenshot 2024-09-30 at 6 30 32 PM

### DIFF
--- a/app/helpers/forms_helper.rb
+++ b/app/helpers/forms_helper.rb
@@ -23,7 +23,7 @@ module FormsHelper
   end
 
   def currencies_for_select
-    Money::Currency.all_instances.sort_by(&:priority)
+    Money::Currency.all_instances.sort_by { |currency| [ currency.priority, currency.name ] }
   end
 
   private


alagos:
BTW, related with the second screenshot, why does the "Belarusian Ruble" has [a higher priority](https://github.com/maybe-finance/maybe/blob/main/config/currencies.yml#L87) compared with other currencies?

zachgoll:
@alagos good idea, I think adding the name as a secondary sort is helpful.

In regards to the priorities of these currencies—there has not been a ton of work done around this yet and most of the priorities are a best guess.

We'd be glad to accept any contribution that provides an updated priorities list given there is reasonable data to back up those changes.  For now, we're holding off on changes as we don't have a ton of data to go off and we'd just be guessing.

alagos:
Right, I'm not sure that I have in mind a reliable source of currency priorities (actually, I think is out of the scope of what I was thinking to fix here).

ssjr:
I set up the project yesterday and started using it locally.
Set the default currency in http://localhost:3000/settings/preferences to "BRL". When I went to add accounts in http://localhost:3000/accounts the default currency was USD and I had to manually change it every time when creating an account.
image

image

Could it be like
```ruby
  def currencies_for_select
    Money::Currency.all_instances.sort_by do |currency|
      family_currency = Current.family&.currency == currency.iso_code ? 0 : 1

      [ family_currency, currency.priority, currency.name ]
    end
  end
```

zachgoll:
@ssjr this is fixed now:

https://github.com/maybe-finance/maybe/pull/1209/commits/e8eb4b0eb964c1ecf83fbaef51e46de2221695c1

PR #1231, opened 2024-10-02
Add formatting for EUR locales

zachgoll:
This PR sets up some basic "exceptions" for formatting currencies.

As of now, we format currencies in the UI based on the following logic:

1. We maintain a list of currencies with default formatting metadata in [currencies.yml](https://github.com/maybe-finance/maybe/blob/main/config/currencies.yml)
2. The `Money` and `Money::Currency` classes read this list and provide a method, `format_options` 
3. We have a global helper, `format_currency` which uses Rails `number_to_currency(format_options)` to format the currency

This works in 95% of cases, but there are some cases where a single currency like `EUR` should be formatted differently based on the locale of the user.

#1226 enabled setting locales for each user.  We are now able to pass the configured locale directly to the `format_options` method on the `Money` class to "override" the default formatting.

For example, `25123.98` in the `EUR` currency will be formatted as follows based on the user's locale:

- `€25,123.98` - for `en`
- `€25.123,98` - for `de`

This PR will certainly not cover all exceptional cases, but sets up an easy way to add these via `locale_options`:

```rb
def locale_options(locale)
  case [ currency.iso_code, locale.to_sym ]
  when [ "EUR", :nl ], [ "EUR", :pt ]
    { delimiter: ".", separator: ",", format: "%u %n" }
  when [ "EUR", :en ], [ "EUR", :en_IE ]
    { delimiter: ",", separator: "." }
  else
    {}
  end
end
```

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -138,13 +138,13 @@ module ApplicationHelper
 
   def format_money(number_or_money, options = {})
     money = Money.new(number_or_money)
-    options.reverse_merge!(money.format_options(Current.family.locale))
+    options.reverse_merge!(money.format_options(I18n.locale))
     number_to_currency(money.amount, options)
   end
 
   def format_money_without_symbol(number_or_money, options = {})
     money = Money.new(number_or_money)
-    options.reverse_merge!(money.format_options(Current.family.locale))
+    options.reverse_merge!(money.format_options(I18n.locale))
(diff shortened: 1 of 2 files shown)


lib/money.rb
```diff
@@ -54,29 +54,6 @@ def exchange_to(other_currency, date: Date.current, fallback_rate: nil)
...
-    format_str = "%.#{precision}f"
-    amount_str = format_str % amount
-    parts = amount_str.split(currency.separator)
-
-    if parts.length < 2
-      ""
-    else
-      parts.last.ljust(precision, "0")
-    end
-  end
```
zachgoll, line 67: legacy, not being used anywhere

PR #1233, opened 2024-10-03
Use DB for auth sessions

zachgoll:
Previously, we were using cookie-based sessions for authentication.  This has security limitations around invalidation as described in the [Rails guides here](https://guides.rubyonrails.org/security.html#session-storage).

This PR does a bit of a clean up around authentication views in general.  In addition, it creates a `sessions` table that will allow for:

- Session invalidation on logout
- Multiple devices to have sessions at once
- Better visibility into logins


zachgoll:
H/T to @ppfeister for calling out the session invalidation issue here!

PR #1262, opened 2024-10-07
fix: default value if user's name isn't set

ajmeese7:
Upon account creation, accessing the home page will just say `Welcome back, ` with no data, since the user is not able to offer this during the sign-up flow. This PR defaults to calling the user "friend" if their first name isn't set, which makes it feel more complete and personable.


zachgoll:
@ajmeese7 good idea!  Would you mind adding an I18n key for this?  Something like:

```rb
<%= Current.user.first_name.present? ? t(".greeting", name: Current.user.first_name ) : t(".fallback_greeting") %>
```

zachgoll:
@ajmeese7 looks good, you'll just have to run `i18n-tasks normalize` to get those tests passing and then we can merge.

ajmeese7:
I don't have the development environment set up locally, my machine isn't configured with Ruby or a suitable database at the moment. Can you run it for me?

PR #1269, opened 2024-10-08
Billing

Shpigford:


app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,65 @@
+class WebhooksController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: [ :stripe ]
+  skip_authentication
+
```
zachgoll, line 4: Think we need to skip csrf protection here:

```rb
protect_from_forgery except: :stripe
```

https://docs.stripe.com/webhooks?verify=check-signatures-library#csrf-protection

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+    rescue JSON::ParserError
+      render json: { error: "Invalid payload" }, status: :bad_request
+      return
+    rescue Stripe::SignatureVerificationError
+      render json: { error: "Invalid signature" }, status: :bad_request
+      return
+    rescue => e
+      Rails.logger.error "Error processing Stripe webhook: #{e.message}"
+      render json: { error: "Internal server error" }, status: :internal_server_error
+      return
```
zachgoll, line 33: We'll probably want to throw all this handling in a background job at some point, so I'd probably remove this "catch all" handler and let Sentry capture anything that reaches this point.  That way this endpoint remains solely responsible for parsing the payload and verifying the signature.

app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  private
+
+    def handle_subscription_event(event)
+      subscription = event.data.object
+      family = Family.find_by(stripe_customer_id: subscription.customer)
+
+      if family
+        family.update(
+          stripe_plan_id: subscription.plan.id,
```
zachgoll, line 47: Is `plan` available on `Subscription`?  I see that you can dig into `subscription.items[0].plan`

app/controllers/webhooks_controller.rb
zachgoll: Ahh wait, that's already covered by skipping `:verify_authenticity_token`, hadn't released that could be used as an alternative to `protect_from_forgery`

app/controllers/webhooks_controller.rb
Shpigford: It is indeed! Don't recall where I found it, but definitely works. 🙂

PR #1274, opened 2024-10-09
Minor improvements to categories & changelog pages

arsenstorm:
Hey there!

I’m testing the hosted version of Maybe, and a few things bothered me regarding the UI.

The grey border around the badges for the categories feels *wrong*. I’ve updated this to match the colour of the badge instead.

Before:

Screenshot 2024-10-09 at 21 38 30

After:

Screenshot 2024-10-09 at 21 38 55

It also feels a bit wrong to see the name of the changelog update instead of the person who created the release.

Before:

Screenshot 2024-10-09 at 21 42 09

After:

Screenshot 2024-10-09 at 21 41 31

### DIFF
--- a/app/models/provider/github.rb
+++ b/app/models/provider/github.rb
@@ -47,6 +47,8 @@ class Provider::Github
         if release
           {
             avatar: release.author.avatar_url,
+            # this is the username, it would be nice to get the full name
+            username: release.author.login,
             name: release.name,
             published_at: release.published_at,
             body: Octokit.markdown(release.body, mode: "gfm", context: repo)
--- a/app/views/categories/_badge.html.erb
+++ b/app/views/categories/_badge.html.erb
@@ -2,9 +2,10 @@
 <% category ||= null_category %>
 
 <div>
-  <span class="flex items-center gap-1 text-sm font-medium rounded-full px-1.5 py-1 border border-alpha-black-25"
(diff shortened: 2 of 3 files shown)


zachgoll:
@arsenstorm thanks for the updates! 

@justinfar will have to approve these as I'm not the best with design :)

arsenstorm:
Awesome; thank you!

justinfar:
Hey @arsenstorm thanks so much for looking into this

Re changing the name to show the person who created the release, 100% agree

Re the border color, initially I had the `alpha-black/25` that blended with the background which gave it a darkened category background color which is exactly what you're doing here (albeit slightly more subdued) 

Figma handles borders/strokes differently in comparison to actual CSS so this probably got a bit lost in translation. 

So overall agree on the change, only thing I'd change is the to make the border slightly lighter and perhaps change from 25% to 30% over here:

```ruby
border-color: color-mix(in srgb, <%= category.color %> 25%, white);
``` 

If you have other UI bits that you think could be improved feel free to let us know :)

arsenstorm:
Ah I see

I think switching 25% -> 30% does make it slightly nicer

Another option is to make it borderless which would align better with the tags and merchants page; however, I personally feel like the badges on these pages would benefit more with a border (or a slightly darker background)

PR #1280, opened 2024-10-10
fix: use correct delimiter on credit card zero values

gariasf:
After recent fixes, the credit card view displays 0 by default on Minimum Payment and Annual Fee. However, the format is wrong. On zero values, the two digits after the delimiter are the decimals, and this is the current behaviour on other sections, like the dashboard.

For example, this is correct:
image
But this is not:
image

This PR resolves this issue, the credit card Minimum Payment and Annual Fee now display the correct delimiter on empty values:
image

I haven't seen any issues related to this, so I can't link one.

First time contributing and first time using ruby, I'll be happy to hear about your feedback.

Cheers.


zachgoll (approved):
Looks good!

PR #1281, opened 2024-10-10
fix: amend inputs on loan, c.c., vehicle, and property partials

gariasf:
It's currently possible to input text on some number values for loans, credit cards, vehicles, and property. When happens and the form is submitted, it is gracefully handled by setting the value to zero. 
This PR proposes that those inputs are number inputs, with appropriate min and step values. % values have a step of 0.1 and both % and currency inputs have a min of 0. This assumes that negative interest rates are VERY rare in personal finance. With that said, I am not sure if it makes sense to put in a negative balance, so I'll be expecting to hear your thoughts on that.
Ideally, these inputs should display the currency symbol or percentage symbol, but that isn't such a trivial change, so I left it out for now.

Looking forward to your feedback. Cheers!

### DIFF
--- a/app/views/accounts/accountables/_credit_card.html.erb
+++ b/app/views/accounts/accountables/_credit_card.html.erb
@@ -9,7 +9,7 @@
 
       <div class="flex items-center gap-2">
         <%= credit_card_form.number_field :minimum_payment, label: t(".minimum_payment"), placeholder: t(".minimum_payment_placeholder"), min: 0 %>
-        <%= credit_card_form.number_field :apr, label: t(".apr"), placeholder: t(".apr_placeholder"), min: 0, step: 0.1 %>
+        <%= credit_card_form.number_field :apr, label: t(".apr"), placeholder: t(".apr_placeholder"), min: 0, step: 0.01 %>
       </div>
 
       <div class="flex items-center gap-2">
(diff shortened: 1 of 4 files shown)


app/views/accounts/accountables/_credit_card.html.erb
```diff
@@ -4,17 +4,17 @@
...
       <div class="flex items-center gap-2">
-        <%= credit_card_form.text_field :available_credit, label: t(".available_credit"), placeholder: t(".available_credit_placeholder") %>
+        <%= credit_card_form.number_field :available_credit, label: t(".available_credit"), placeholder: t(".available_credit_placeholder"), min: 0 %>
       </div>
 
       <div class="flex items-center gap-2">
-        <%= credit_card_form.text_field :minimum_payment, label: t(".minimum_payment"), placeholder: t(".minimum_payment_placeholder") %>
-        <%= credit_card_form.text_field :apr, label: t(".apr"), placeholder: t(".apr_placeholder") %>
+        <%= credit_card_form.number_field :minimum_payment, label: t(".minimum_payment"), placeholder: t(".minimum_payment_placeholder"), min: 0  %>
+        <%= credit_card_form.number_field :apr, label: t(".apr"), placeholder: t(".apr_placeholder"), min: 0, step: 0.1 %>
```
zachgoll, line 12:
```suggestion
        <%= credit_card_form.number_field :apr, label: t(".apr"), placeholder: t(".apr_placeholder"), min: 0, step: 0.01 %>
```

zachgoll:
@gariasf those two suggestions I made should fix the system test failures

zachgoll (approved):
Awesome, looks great!

PR #1286, opened 2024-10-10
Add additional subtypes and allow for None

ahatzz11:
- Add additional investment subtypes
- Allow for "None" option since any editing defaults to "Brokerage" right now.
- Small reorder to put the 401k options next to each other

_I've been in software for a decade but have never touched ruby, so using this as a slow learning opportunity_

### DIFF
--- a/app/models/investment.rb
+++ b/app/models/investment.rb
@@ -2,16 +2,19 @@ class Investment < ApplicationRecord
   include Accountable
 
   SUBTYPES = [
+    [ "None", nil ],
     [ "Brokerage", "brokerage" ],
     [ "Pension", "pension" ],
     [ "Retirement", "retirement" ],
     [ "401(k)", "401k" ],
-    [ "529 plan", "529_plan" ],
+    [ "Roth 401k", "roth_401k" ],
+    [ "529 Plan", "529_plan" ],
     [ "Health Savings Account", "hsa" ],
     [ "Mutual Fund", "mutual_fund" ],
+    [ "Traditional IRA", "traditional_ira" ],
     [ "Roth IRA", "roth_ira" ],
-    [ "Roth 401k", "roth_401k" ],
-    [ "Angel", "angel" ]
+    [ "Angel", "angel" ],
+    [ "Cryptocurrency", "cryptocurrency" ]
   ].freeze
 
   def value


zachgoll:
@ahatzz11 thanks for the update.  Agree w/the organization here.

For handling the nil value, I think we just need to remove that `selected: ""` option in the select input.  Here's my diff:

```diff
diff --git a/app/views/accounts/accountables/_investment.html.erb b/app/views/accounts/accountables/_investment.html.erb
index d403d99..9d97ced 100644
--- a/app/views/accounts/accountables/_investment.html.erb
+++ b/app/views/accounts/accountables/_investment.html.erb
@@ -1 +1 @@
-<%= f.select :subtype, options_for_select(Investment::SUBTYPES, selected: ""), { label: true } %>
+<%= f.select :subtype, options_for_select(Investment::SUBTYPES), { label: true, prompt: t(".prompt") } %>
diff --git a/config/locales/views/accounts/en.yml b/config/locales/views/accounts/en.yml
index b0b6421..56b386a 100644
--- a/config/locales/views/accounts/en.yml
+++ b/config/locales/views/accounts/en.yml
@@ -5,6 +5,8 @@ en:
       has_issues: Issue detected.
       troubleshoot: Troubleshoot
     accountables:
+      investment:
+        prompt: Select subtype
       credit_card:
         annual_fee: Annual fee
         annual_fee_placeholder: '99'
```

ahatzz11:
> @ahatzz11 thanks for the update. Agree w/the organization here.
> 
> For handling the nil value, I think we just need to remove that `selected: ""` option in the select input. Here's my diff:
> 
> ...

Thanks for the suggestion! I've gone ahead and made that change as well. I'll make a bug report for the state issue, but I appreciate the help so far!

zachgoll:
@ahatzz11 I see what you're saying now.  I think I misread the first time around.

Rails forms _should_ be able to populate this value automatically without much additional code.  Let me take a deeper look at this and see what's going on.

zachgoll:
@ahatzz11 okay I figured out the issue.  To leverage some of the Rails form magic, we need to remove `options_for_select` and pass an array of options directly.  Looks like we have to do this in two places:

```erb
<%# _depository.html.erb %>

<%= f.select :subtype, [["Checking", "checking"], ["Savings", "savings"]], { label: "Type", prompt: "Select subtype" } %>
```

```erb
<%# _investment.html.erb %>

<%= f.select :subtype, Investment::SUBTYPES, { label: true, prompt: "Select subtype" } %>
```

This should solve all the issues here:

- Allows user to leave blank
- Properly pre-selects the value already stored on the account

ahatzz11:
@zachgoll Hey Zach, after playing with this for awhile I don't think that quite works (but if I'm doing something wrong please let me know). Here is what I see when running your code for the subtypes:

1. Create new account with Brokerage:
	image
2. Edit account, no "None" option (but "Brokerage" is selected):
	image

I went back and forth on a bunch of options but for for the sake of not overcomplicating this I think the following should be true:
- If no option is selected the helper prompt should be shown
- If a subtype is chosen at creation or added later it should be shown when the edit screen pops up
- There should be a "None" option for a user to select.

I did like how you removed the `options_for_select` and the associated `selected` though, that still seems to do the trick on using the current state. I took my original code and made that tweak from yours and this is now fulfilling all 3 of the bullets above:
```ruby
<%= f.select :subtype, Investment::SUBTYPES, { label: true, prompt: t(".prompt"), include_blank: t(".none") } %>
```

I think the `include_blank` option is required here to ensure there is a blank option for the user to chose for `nil`. They _could_ pick the "Select subtype" option, but I think that's kinda annoying. 

I couldn't find any nice way to have non-selectable helpertext on this ruby form, so this is where I landed:
image
image

I also did this for depository! Happy to make any other changes you think would be good here.

zachgoll (approved):
Yep, think these changes make sense!  Thanks for the help on this one.

Issue #1295, opened 2024-10-11
Bug: Private method in Stimulus controllers causes `SyntaxError: Private field must be declared in an enclosing class` when minified


zachgoll:
_Note: a temporary fix has been added for this issue by removing all private method syntax and using underscores instead, but it would be nice to identify the root cause of this_

**Describe the bug**

This bug only occurs in production when Stimulus controllers are minified.

For example, the linked controller produces the following error:

`SyntaxError: Private field '#updateView' must be declared in an enclosing class`

Here is the original file:

https://github.com/maybe-finance/maybe/blob/c5bf1db230122066571b018ed417d132e6e7e446/app/javascript/controllers/bulk_select_controller.js

And here is the minified source served in production environments:

```js
import{Controller}from "@hotwired/stimulus"
export default class extends Controller{static targets=["row","group","selectionBar","selectionBarText","bulkEditDrawerTitle"]
static values={resource:String,selectedIds:{type:Array,default:[]}}
connect(){document.addEventListener("turbo:load",this._updateView)
this._updateView()}
disconnect(){document.removeEventListener("turbo:load",this._updateView)}
bulkEditDrawerTitleTargetConnected(element){element.innerText=`Edit ${this.selectedIdsValue.length} ${this.#pluralizedResourceName()}`}
submitBulkRequest(e){const form=e.target.closest("form");const scope=e.params.scope
this.#addHiddenFormInputsForSelectedIds(form,`${scope}[entry_ids][]`,this.selectedIdsValue)
form.requestSubmit()}
togglePageSelection(e){if(e.target.checked){this.#selectAll()}else{this.deselectAll()}}
toggleGroupSelection(e){const group=this.groupTargets.find(group=>group.contains(e.target))
this.#rowsForGroup(group).forEach(row=>{if(e.target.checked){this.#addToSelection(row.dataset.id)}else{this.#removeFromSelection(row.dataset.id)}})}
toggleRowSelection(e){if(e.target.checked){this.#addToSelection(e.target.dataset.id)}else{this.#removeFromSelection(e.target.dataset.id)}}
deselectAll(){this.selectedIdsValue=[]
this.element.querySelectorAll('input[type="checkbox"]').forEach(el=>el.checked=false)}
[...]

PR #1296, opened 2024-10-11
fix stimulus minification on time series and bulk select controllers

gariasf:
This PR attempts to close https://github.com/maybe-finance/maybe/issues/1295.

I'm a bit unsure about it because the bulk select controller doesn't have private fields but rather private methods, and they are not placed directly after an object without a semicolon. However, it's worth a try.

The only catch here is that I wasn't able to validate this locally because I don't know how to build the assets with minification.


zachgoll:
@gariasf thanks for taking a go at this!  I think we should get #1299 (or something similar) merged which should fix and give us some confidence developing these controllers moving forward

gariasf:
Hey there! Sure, that makes sense. Closing.

Issue #1313, opened 2024-10-15
Improvement: Add tooltip to account "pill" on dashboard


zachgoll:
Currently, users have no way of identifying the accounts listed in the metrics on the dashboard.

We should add a tooltip over each on hover using `tooltip_controller.js`

![image](https://github.com/user-attachments/assets/ebf8093f-f152-449c-91d6-c1365feef33c)

PR #1315, opened 2024-10-16
add dashboard account pill tooltips

gariasf:
Fixes https://github.com/maybe-finance/maybe/issues/1313

### DIFF
--- a/app/views/shared/_text_tooltip.erb
+++ b/app/views/shared/_text_tooltip.erb
@@ -1,0 +1,5 @@
+<div role="tooltip" data-tooltip-target="tooltip" class="tooltip bg-gray-700 text-xs p-1 rounded">
+  <div class="text-white font-normal">
+    <%= tooltip_text %>
+  </div>
+</div>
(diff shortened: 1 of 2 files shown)


app/views/shared/_text_tooltip.erb
```diff
@@ -0,0 +1,5 @@
+<div role="tooltip" data-tooltip-target="tooltip" class="tooltip bg-gray-700 text-xs p-1 rounded">
```
zachgoll, line 1:
```suggestion
<div role="tooltip" data-tooltip-target="tooltip" class="tooltip bg-gray-700 text-sm px-1.5 py-1 rounded-md">
```

Just a few tweaks based on official specs.  My bad for not including Figma file!

BesBettison: Puzzle in this step, Comprehensive clear  but on moving steps ahead may hire wire,

BesBettison: My dad never input the figma files

PR #1323, opened 2024-10-17
Redirect upload step

### LINKED ISSUE #1322: Bug: View Import doesn't work as expected if the import file doesn't exist
**Describe the bug**
If the import was interrupted before the import file was loaded, `View Import` must be directed to the correct location.

**To Reproduce**
1. Go to Import index
2. Click on New Import
3. Don't upload a file
4. Quit from the import
5. Back to import list
6. Click View Import button from the list
7. Redirect to Configuration step instead Upload step

**Expected behavior**
`View button` should be redirect to Upload step if the import doesn't have a document.

**Screenshots / Recordings**

https://github.com/user-attachments/assets/8e9d062e-47a0-41e1-a61b-417d425f9a29

enderahmetyurt:
Fix #1322

### DIFF
--- a/app/controllers/imports_controller.rb
+++ b/app/controllers/imports_controller.rb
@@ -24,7 +24,7 @@ class ImportsController < ApplicationController
   end
 
   def show
-    redirect_to import_confirm_path(@import), alert: "Please finalize your mappings before proceeding." unless @import.publishable?
+    redirect_to import_upload_path(@import), alert: "Please finalize your mappings before proceeding." unless @import.publishable?
   end
 
   def destroy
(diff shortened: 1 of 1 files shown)


Shpigford:
I agree with the bug, but we likely need a bit more nuance to *where* this redirects.

There's an `uploaded?` method on the Import model that checks for an uploaded file so maybe we need some sort of if/else or case statement to check for the current step and redirect to the appropriate one?

enderahmetyurt:
Good point @Shpigford. How does it look now?

Shpigford:
@enderahmetyurt Works for me! Thanks!

PR #1325, opened 2024-10-17
Impersonation

Shpigford:
This is the first pass at an impersonation feature.

This is very much for the Maybe hosted version and purely to help us provide _secure_ support with an auditable trail of activity.

## Flow
1. User contacts support with an issue.
2. Support, which has a "super_admin" boolean enable, sends a request to the user to enable "impersonation mode"
3. User sees a new banner in their account asking them to approve or reject the request
4. Once User approves, Support now is able to "switch" to that user's account and see what they see.
5. All Support activity in the account is thoroughly logged (ImpersonationSessionLogs)
6. Both User and Support can end access at any point and Support will only keep access as long as is necessary to resolve the ticket. Future issues will require re-requesting access.

## Things to review
1. The code feels a bit unnecessarily complex and possibly redundant in places. Especially in the Authentication concern.
2. I want to be VERY thorough with security on this. Would obviously be a major problem for there to be some sort bug that allowed anyone to impersonate another account. So, need eyes on that.

CleanShot 2024-10-17 at 14 54 01@2x

CleanShot 2024-10-17 at 14 54 55@2x

CleanShot 2024-10-17 at 14 55 22@2x

CleanShot 2024-10-17 at 14 55 40@2x

CleanShot 2024-10-17 at 14 56 01@2x

CleanShot 2024-10-17 at 14 56 24@2x


Shpigford:
@zachgoll Basically the entire test suite is failing now and my eyes are crossing trying to nail down why. Some fresh eyes may be useful.

PR #1326, opened 2024-10-18
Add impersonation testing

zachgoll:
This PR adds a test suite for the impersonation feature along with a few changes around how the sessions are stored:

- All critical validation happens directly on `ImpersonationSession` model
- A `Session` stores the impersonation session and can have exactly 1 active `ImpersonationSession` at a time
- Access all sessions _through_ the `Current.true_user` object to ensure unauthorized users cannot hijack a session


db/migrate/20241009214601_add_super_admin_to_users.rb
```diff
@@ -1,5 +1,23 @@
...
+      dir.down do
+        execute <<-SQL
+          CREATE TYPE user_role AS ENUM ('admin', 'member');
+        SQL
+
+        change_column_default :users, :role, nil
+        change_column :users, :role, :user_role, using: 'role::user_role'
+        change_column_default :users, :role, 'member'
+      end
+    end
```
zachgoll, line 21: We have an existing `role` on the `User` table that I turned into a string (leverage ActiveRecord more, make more flexible) and added the `super_admin` option to the enum in the `User` model.

test/models/impersonation_session_test.rb
```diff
@@ -1,7 +1,40 @@
...
+  test "impersonation session must have different impersonator and impersonated" do
+    super_admin = users(:maybe_support_staff)
+
+    assert_raises(ActiveRecord::RecordInvalid) do
+      ImpersonationSession.create!(
+        impersonator: super_admin,
+        impersonated: super_admin
+      )
+    end
+  end
```
zachgoll, line 39: Tests to ensure super admins are the only users in the system able to initiate this functionality

app/models/session.rb
```diff
@@ -1,5 +1,9 @@
 class Session < ApplicationRecord
   belongs_to :user
+  belongs_to :active_impersonator_session,
+    -> { where(status: :in_progress) },
+    class_name: "ImpersonationSession",
+    optional: true
```
zachgoll, line 6: This will only be set for the `super_admin` session.  It represents the impersonation session that the Admin is _currently_ viewing.

db/migrate/20241009214601_add_super_admin_to_users.rb
Shpigford: My thought on the "super_admin" _column_ vs enum is that the "role" column is directly related to a user's actual Maybe account whereas "super_admin" is more of an internal function of administrating the product.

Maybe that's splitting hairs, though?

zachgoll: Long term if the "super admin" starts severely diverging in capabilities within the app, I'd probably look towards creating a brand new table, `SuperAdminUser` that doesn't have a `Family`, `Account`s, or other domain related associations / properties so it's sole purpose is to log in as a "superadmin" and do administrative tasks.  But since these current super admins are logging in with regular `User` accounts, I think it makes sense that they have an "elevated privilege role" within the app's features.

Shpigford: that's fair. in reality we'd likely just build an entirely separate admin UI and flow anyways as having support members be logged in to their own Maybe account is a bit of a weird flow anyways.

Shpigford (approved):
Overall looks good to me!

db/migrate/20241009214601_add_super_admin_to_users.rb
zachgoll: Yep, agree!

PR #1327, opened 2024-10-18
Fixes issue with mapping values during the transactions import

### LINKED ISSUE #1317: Bug: Incorrect numbers in transactions import summary
**Describe the bug**
At the final step of a transactions import, the summary showing how many new items will be added displays numbers that don’t match the options selected in the previous steps. 

I imported a CSV with 2 rows, mapped every category, tag, and account, but the summary shows that some of those items will be created.

If I go back to the mapping step, I can see that some of the selected options were cleared.

**To Reproduce**
1. Go to "Import"
2. Click on "Import transactions"
3. Paste rows separated by commas or upload a CSV
4. Click "Upload CSV"
5. Configure your import and click "Apply configuration"
6. Click "Next step" on the clean step
7. Select an existing category for each category in the list, then click "Next"
8. Select an existing tag for each tag in the list, then click "Next"
9. Select an existing account for each account in the list, then click "Next"

**Expected behavior**
I expected a zero next to Accounts, Categories, and Tags, but I see a number greater than 0.

**Screenshots / Recordings**
image
image
image
image

If I go back to Map step I see that some options have been cleared:
image
image

nicogaldamez:
### Why?

- Fixes #1317 

At the final step of a transactions import, the summary showing how many new items will be added displays numbers that don’t match the options selected in the previous steps.

I imported a CSV with 2 rows, mapped every category, tag, and account, but the summary shows that some of those items will be created.
If I go back to the mapping step, I can see that some of the selected options were cleared.

**The issue was caused by the debounce timeout in the auto-submit-form Stimulus controller. It waits 500ms to submit the form, so if you click the 'Next' button too quickly, the form doesn't get submitted, and the option isn't saved.**

### What?
I added a default debounce timeout that adjusts based on the element type, and also made it customizable by allowing a custom debounce timeout to be set through a data attribute

### What should we test?

Select an item during the mapping step when importing transactions and quickly move to the next step. At the final step of the import, you should now see a 0 next to Categories, Accounts, and Tags.

### DIFF
--- a/app/javascript/controllers/auto_submit_form_controller.js
+++ b/app/javascript/controllers/auto_submit_form_controller.js
@@ -24,10 +24,33 @@ export default class extends Controller {
     });
   }
 
-  handleInput = () => {
+  handleInput = (event) => {
+    const target = event.target
+
     clearTimeout(this.timeout);
     this.timeout = setTimeout(() => {
       this.element.requestSubmit();
-    }, 500);
+    }, this.#debounceTimeout(target));
   };
+
+  #debounceTimeout(element) {
+    if(element.dataset.autosubmitDebounceTimeout) {
+      return parseInt(element.dataset.autosubmitDebounceTimeout);
+    }
+
+    const type = element.type || element.tagName;
+
+    switch (type.toLowerCase()) {
+      case 'input':
+      case 'textarea':
+        return 500;
+        break;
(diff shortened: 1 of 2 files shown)


iwfyou (approved):

zachgoll (approved):
Thanks for tackling this one!

PR #1328, opened 2024-10-18
Basic account onboarding

zachgoll:
Allows a user to select the "mode" their account's are in:

https://github.com/user-attachments/assets/c81495ff-5eb0-4f74-82c4-1d9246c6d7f9

Also includes a few bug fixes around syncing.

### DIFF
--- a/app/views/accounts/accountables/investment/_tabs.html.erb
+++ b/app/views/accounts/accountables/investment/_tabs.html.erb
@@ -1,28 +1,34 @@
 <%# locals: (account:, selected_tab:) %>
 
-<% if account.entries.account_trades.any? || account.entries.account_transactions.any? %>
+<% if account.mode.nil? %>
+  <%= render "accounts/accountables/value_onboarding", account: account %>
+<% else %>
   <div class="flex gap-2 text-sm text-gray-900 font-medium mb-4">
-    <%= render "accounts/accountables/tab", account: account, key: "holdings", is_selected: selected_tab.in?([nil, "holdings"]) %>
-    <%= render "accounts/accountables/tab", account: account, key: "cash", is_selected: selected_tab == "cash" %>
-    <%= render "accounts/accountables/tab", account: account, key: "transactions", is_selected: selected_tab == "transactions" %>
+    <% if account.mode == "transactions" %>
+      <%= render "accounts/accountables/tab", account: account, key: "holdings", is_selected: selected_tab.in?([nil, "holdings"]) %>
+      <%= render "accounts/accountables/tab", account: account, key: "cash", is_selected: selected_tab == "cash" %>
+      <%= render "accounts/accountables/tab", account: account, key: "transactions", is_selected: selected_tab == "transactions" %>
+    <% end %>
   </div>
 
   <div class="min-h-[800px]">
-    <% case selected_tab %>
-    <% when nil, "holdings" %>
-      <%= turbo_frame_tag dom_id(account, :holdings), src: account_holdings_path(account) do %>
-        <%= render "account/entries/loading" %>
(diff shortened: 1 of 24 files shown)


app/views/accounts/accountables/investment/_tabs.html.erb
```diff
@@ -1,28 +1,34 @@
...
+          <%= render "account/entries/loading" %>
+        <% end %>
+      <% when "transactions" %>
+        <%= turbo_frame_tag dom_id(account, :trades), src: account_trades_path(account) do %>
+          <%= render "account/entries/loading" %>
+        <% end %>
       <% end %>
+    <% else %>
+      <%= render "accounts/accountables/valuations", account: account %>
     <% end %>
```
zachgoll, line 32: Will need to come back around and clean a lot of this markup up.  Lots of duplication across the accountable views

test/models/account/balance/syncer_test.rb
```diff
@@ -47,6 +47,17 @@ class Account::Balance::SyncerTest < ActiveSupport::TestCase
...
+  test "syncs account with valuations and transactions when transaction starts" do
+    new_account = families(:empty).accounts.create!(name: "Test Account", balance: 1000, currency: "USD", accountable: Depository.new)
+    create_transaction(account: new_account, date: 2.days.ago.to_date, amount: 250)
+    create_valuation(account: new_account, date: Date.current, amount: 1000)
+
+    run_sync_for(new_account)
+
+    assert_equal 1000, new_account.balance
+    assert_equal [ 1250, 1000, 1000, 1000 ], new_account.balances.chronological.map(&:balance)
+  end
```
zachgoll, line 59: Validates bug when there is a transaction at an earlier date than the first valuation

Muzosh:
I'm confused. What if I want to use both the way it was before? I.e. adding transactions for personal transaction logging, but also update value based on current's date account appreciation?

Muzosh:
I would also argue that this splitting accounts into two modes is even more confusing. Let's say I created the account above, with "today's balance" 1 CZK. I then change my mind and convert this account from balance mode to transactions mode and add some transactions on the same day.

The value is not updated because the "today's balance" is still overwriting it.

I guess this is intented (although I could again argue that I added transaction **after** I updated today's balance on the same day) but now user has even lower chance of figuring out what is going on since he thinks his account is in "transactions only" mode.

Muzosh:
Also if the account is in "balance only" mode, I can still successfully import transactions. However, after import it gets immediately overwritten by new balance entry, that has today's date and same value as previous entry.

ttkjakub:
I was just thinking about the same thing today. I believe a solution here might be introducing a third value tracking mode called "Advanced" which would be a combination of Balance and Transaction tracking mode (= basically the old behavior).

What do you think @zachgoll?

jakubkottnauer:
I was just thinking about the same thing today. I believe a solution here might be introducing a third value tracking mode called "Advanced" which would be a combination of Balance and Transaction tracking mode (= basically the old behavior).

What do you think @zachgoll?

zachgoll:
> I was just thinking about the same thing today. I believe a solution here might be introducing a third value tracking mode called "Advanced" which would be a combination of Balance and Transaction tracking mode (= basically the old behavior).
> 
> 
> 
> What do you think @zachgoll?

This is in the works already!  It will just be a general "Activity View"

zachgoll:
Working on an update that will look like this, which we sourced quite a bit of feedback on in Discord:

image

Muzosh:
@zachgoll I just updated my instance and I'm super satisfied with this activity window. It solves all the issues I have written above :) Good job!

PR #1333, opened 2024-10-21
Refactor: Extract Reused String Literals to Constants

Anish-Chanda:
Just a quick cleanup based on some SonarQube suggestions—moved a bunch of repeated string literals to constants. Not the most exciting change, but it does help make the code a bit more manageable and avoids duplication.

No big functional changes here, just some housekeeping to keep things tidy :)

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -1,4 +1,5 @@
 class AccountsController < ApplicationController
+  SUCCESS = ".success".freeze
   layout :with_sidebar
 
   include Filterable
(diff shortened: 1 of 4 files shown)


Anish-Chanda:
Thanks to my class teammates for their input: @elim655 , @ACDrafahl ...

iwfyou (approved):

zachgoll:
Hey @Anish-Chanda, thanks for the contribution!

In this case, I don't think we should merge this because this is adding a layer of indirection to the code that I think actually causes it to be more confusing.

With string literals, the main areas of the codebase that I'd be most concerned about is for "magic number" type values.  A great example of this is this code:

https://github.com/maybe-finance/maybe/blob/da7f19d5ab533a9275f1fd4a3681a29f5ab3175e/app/models/account/entry.rb#L15

In this case, we're extracting an arbitrary "minimum supported date" into a method (could have also been a constant) so that it's clear in this code that this is a special case value.

PR #1351, opened 2024-10-22
Stock Exchanges with seed

Shpigford:
Creating the foundation for a few other features by adding stock exchange data.

This includes adding an "exchanges" seed (along with YML file).

This is needed during bin/setup, so want to go ahead and get it in the codebase for future usage.

The seed file also gets run in the migration file.


Shpigford:
@zachgoll Only question for me here is should the "exchange" seed file get processed as part of the migration so that self-hosters immediately get setup without having to do anything?

db/schema.rb
```diff
@@ -535,7 +557,7 @@
     t.datetime "updated_at", null: false
     t.string "last_prompted_upgrade_commit_sha"
     t.string "last_alerted_upgrade_commit_sha"
-    t.string "role", default: "member", null: false
+    t.enum "role", default: "member", null: false, enum_type: "user_role"
```
zachgoll, line 560: This appears to be from the recent impersonations migration.  I think you'll need to run a `db:reset` to clear this out since this _should_ be a `t.string` now.

Shpigford: I had to add a new migration to just fix it for good. I think the problem from before was that the migration file from the original impersonations was editing after it'd already been run on my end. So we just ended up with a jumbled mess.

Running `db:reset` didn't fix it unfortunately.

_Should_ be addressed now, though.

Issue #1354, opened 2024-10-23
Feature: filter on uncategorized transactions


zachgoll:
### Discussed in https://github.com/maybe-finance/maybe/discussions/1298

<div type='discussions-op-text'>

<sup>Originally posted by **Pierricktnl** October 12, 2024</sup>
### Describe the feature

Add the possibility to filter on "**Uncategorized**" transactions
<img width="574" alt="Screenshot 2024-10-12 at 11 02 32" src="https://github.com/user-attachments/assets/bc0b1443-aea3-4d81-a0f6-ac5e0768bfec">

### Why is this feature important?

 So the user can see the transactions that need a category (act like a work list). 

### Additional context, screenshots, and relevant links

_No response_</div>

zachgoll:
This could be very useful if a user is looking for all categorized transactions so they can bulk-assign a new category to them.

PR #1359, opened 2024-10-24
Feature | Filter on uncategorized transactions

bruno-costanzo:
### Why?
- Related feature request: [Feature: filter on uncategorized transactions #1354](https://github.com/maybe-finance/maybe/issues/1354)
In the transaction index, in the filter by category, it is not possible to filter those transactions that do not have any assigned category.

### What?
- On the filter view side, the Uncategorized category was added among the options.
- In the model, the query is updated to be able to filter transactions without category.

### Test
When filtering only on uncategorized transactions, these should appear and those with a category should not appear. When adding a category, both those without and those with a category should appear.

### Screenshot
image
image

co-author @nicogaldamez

### DIFF
--- a/app/helpers/categories_helper.rb
+++ b/app/helpers/categories_helper.rb
@@ -6,6 +6,6 @@ module CategoriesHelper
   end
 
   def family_categories
-    [null_category].concat(Current.family.categories.alphabetically)
+    [ null_category ].concat(Current.family.categories.alphabetically)
   end
 end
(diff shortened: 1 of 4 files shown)


zachgoll (approved):
Thanks for the contribution @bruno-costanzo!  Everything looks good to me.

You'll just need to fix those lint errors and then we can merge: `bin/rubocop -f github -a`

bruno-costanzo:
done  @zachgoll

PR #1363, opened 2024-10-24
Stock imports

Shpigford:
Sets the foundation for adding Marketstack as a data provider for non-US stocks.

This is primarily focused on populating the Securities table with all global stocks in the hosted version, but lays the groundwork for pulling prices as well as letting self-hosters only import specific exchanges.


app/models/provider/marketstack.rb
```diff
@@ -0,0 +1,143 @@
...
+    )
+  rescue StandardError => error
+    SecurityPriceResponse.new(
+      success?: false,
+      error: error,
+      raw_response: error
+    )
+  end
+
+  def fetch_all_tickers
```
zachgoll, line 36: Looks like we might be able to consolidate into a `fetch_tickers(exchange_mic: nil)` method that covers both `fetch_all_tickers` and `fetch_exchange_tickers`?

```rb
def fetch_tickers(exchange_mic: nil)
  endpoint = "#{base_url}/tickers"
  endpoint = endpoint + "?exchange=#{exchange_mic}" if exchange_mic.present?

  tickers = paginate(endpoint) do |body|
    # ...
  end
end
```

And then over in importer, would reduce to:

```rb
securities = @provider.fetch_tickers(exchange_mic: @exchange)&.tickers
```

app/jobs/securities_import_job.rb
```diff
@@ -0,0 +1,10 @@
+class SecuritiesImportJob < ApplicationJob
+  queue_as :default
+
+  def perform(exchange_mic = nil)
```
zachgoll, line 4: From a user perspective, my guess is that we're wanting to import by country code correct?

If that's the case, would `country_code = nil` be a better param to pass?  That way, we could do something like this in the job and make it 1 job per user:

```rb
def perform(country_code = nil)
  exchanges = StockExchange.in_country(country_code)
  market_stack_client = Provider::Marketstack.new(ENV["MARKETSTACK_API_KEY"])

  exchanges.each do |exchange|
    importer = Security::Importer.new(market_stack_client, exchange.mic)
    importer.import
  end
end
```

And then just a simple scope on `StockExchange`:

```rb
class StockExchange < ApplicationRecord
  scope :in_country, ->(country_code) { where(country_code: country_code) }
end
```

zachgoll (commented):
Setup looks good to me!

app/models/provider/marketstack.rb
Shpigford: Good catch1

app/jobs/securities_import_job.rb
Shpigford: Good call. Keeps us from having to arbitrarily map mic codes as well.

Shpigford:
@zachgoll Any other changes you'd want made before this gets pushed?

Next PR will be actually setting up appropriate importing based on hosted vs self-hosted. So this on its own doesn't actually make any data additions (just db schema).

zachgoll:
@Shpigford yeah I think we're all set for this first step here

PR #1364, opened 2024-10-24
Add good job dashboard with auth

zachgoll:
Gives super admin access to jobs dashboard for hosted version of the app.


Shpigford:
Tested locally and this worked as intended.

PR #1385, opened 2024-10-28
Do not include income transactions in liability accounts for savings rate

### LINKED ISSUE #1383: Bug: Liability accounts are included in "Savings Rate" insight card on dashboard
**Describe the bug**

On the dashboard, liability accounts like credit cards are being grouped into the "savings rate" insight card, which is always invalid since credit cannot count towards savings.

**To Reproduce**
1. Add liability account like a credit card
2. Add an income transaction
3. Go to dashboard
4. See account included in savings rate card

**Expected behavior**

We should exclude all liability accounts from the "savings rate" insight card.

tonyvince:
Fix #1383


app/models/account/entry.rb
```diff
@@ -17,6 +17,7 @@ class Account::Entry < ApplicationRecord
   scope :chronological, -> { order(:date, :created_at) }
   scope :reverse_chronological, -> { order(date: :desc, created_at: :desc) }
   scope :without_transfers, -> { where(marked_as_transfer: false) }
+  scope :without_income_transactions_in_liability_accounts, -> { joins(:account).where(amount: 0..).where.not(accounts: { classification: Account.classifications[:liability] }).distinct }
```
zachgoll, line 20: I think we're on the right track here, but I wouldn't make a scope for this given how specific it is to the `snapshot_transactions` method.

Also, "Income" would be anything _less than_ 0, so I _think_ we need to flip that condition here

zachgoll (approved):
Nice, I think this should fix it.  We'll likely have other little edge cases pop up but I think this is a good step forward.

PR #1388, opened 2024-10-29
First pass at security price reference

Shpigford:


Shpigford:
@zachgoll No tests have been updated, but wanting a sanity check from you when you get a chance.

app/models/security.rb
```diff
@@ -2,6 +2,7 @@ class Security < ApplicationRecord
   before_save :upcase_ticker
 
   has_many :trades, dependent: :nullify, class_name: "Account::Trade"
+  has_many :prices, dependent: :destroy, class_name: "Security::Price"
```
zachgoll, line 5: Namespacing _should_ handle this to where we don't need a `class_name` for `prices`

app/models/security/price/provided.rb
```diff
@@ -6,17 +6,16 @@ module Security::Price::Provided
   class_methods do
     private
 
-      def fetch_price_from_provider(ticker:, date:, cache: false)
+      def fetch_price_from_provider(security:, date:, cache: false)
         return nil unless security_prices_provider.present?
 
         response = security_prices_provider.fetch_security_prices \
-          ticker: ticker,
```
zachgoll, line 13: I know this is probably in the works, but will need `security.ticker` here

db/migrate/20241029125406_add_reference_to_security_prices.rb
```diff
@@ -0,0 +1,5 @@
+class AddReferenceToSecurityPrices < ActiveRecord::Migration[7.2]
+  def change
+    add_reference :security_prices, :security, foreign_key: true, type: :uuid
```
zachgoll, line 3: I'd also remove `ticker` from the `security_prices` table to avoid any confusion.  We'll also want to link any existing prices to their relevant security; something like (pseudocode below):

```rb
Security::Price.each do |sp|
  security = Security.find_by(ticker: sp.ticker)
  sp.security = security
  sp.save!
end
```

Shpigford: Good call on linking. I'm goign to save removing the ticker until after this is deployed/migrated so we have the ticker data in case something goes wrong with the data migration.

zachgoll (approved):
Ran everything locally and works good.

Just a quick diff on `generator.rb` so `rake demo_data:reset` works:

```diff
maybe $ git diff app/models/demo/generator.rb 
diff --git a/app/models/demo/generator.rb b/app/models/demo/generator.rb
index ba210b2..b56ef1b 100644
--- a/app/models/demo/generator.rb
+++ b/app/models/demo/generator.rb
@@ -176,12 +176,12 @@ class Demo::Generator
 
     def load_securities!
       # Create an unknown security to simulate edge cases
-      Security.create! ticker: "UNKNOWN", name: "Unknown Demo Stock"
+      Security.create! ticker: "UNKNOWN", name: "Unknown Demo Stock", exchange_mic: "UNKNOWN"
 
       securities = [
-        { ticker: "AAPL", name: "Apple Inc.", reference_price: 210 },
-        { ticker: "TM", name: "Toyota Motor Corporation", reference_price: 202 },
-        { ticker: "MSFT", name: "Microsoft Corporation", reference_price: 455 }
+        { ticker: "AAPL", exchange_mic: "NASDAQ", name: "Apple Inc.", reference_price: 210 },
+        { ticker: "TM", exchange_mic: "NYSE", name: "Toyota Motor Corporation", reference_price: 202 },
+        { ticker: "MSFT", exchange_mic: "NASDAQ", name: "Microsoft Corporation", reference_price: 455 }
       ]
 
       securities.each do |security_attributes|
@@ -193,7 +193,7 @@ class Demo::Generator
           low_price = reference - 20
           high_price = reference + 20
           Security::Price.create! \
-            ticker: security.ticker,
+            security: security,
             date: date,
             price: Faker::Number.positive(from: low_price, to: high_price)
         end
```

PR #1391, opened 2024-10-29
Bug: Edited transactions on the home page get a checkbox

### LINKED ISSUE #1386: Bug: Edited transactions on the home page get a checkbox
**Describe the bug**
On the home page there is a list of recent transactions. None of these transactions have a checkbox. If a transaction is edited it then gets a checkbox, which doesn't have any options once selected

**To Reproduce**
1. Have a few transactions
2. Go to the home page - https://app.maybe.co/
3. Edit a transaction directly on this page by clicking on it (sidebar edit comes up)
4. See checkbox on that edited transaction

**Expected behavior**
I would expect consistency in one of two ways:
- All transactions have a checkbox (assuming there is an action you can take from this page, which doesn't seem to be the case currently)
- No transactions have a checkbox, even after an edit.

Hosted

Chrome 130, MacOS 14

**Screenshots / Recordings**
<p>
image
image
image
</p>

**Additional context**

nicogaldamez:
### Why?

* Fixes #1386 
On the home page there is a list of recent transactions. None of these transactions have a checkbox. If a transaction is edited it then gets a checkbox, which doesn't have any options once selected

### What?

* I added a parameter to the transaction form to specify whether the corresponding row returned by the Turbo Stream response should include a checkbox or not.

### What should we test?
* When editing a transaction from the recent transactions list, the updated row should no longer have a checkbox.
* When editing a transaction from the transactions list, the updated row should still have a checkbox.

### DIFF
--- a/app/controllers/account/transactions_controller.rb
+++ b/app/controllers/account/transactions_controller.rb
@@ -16,7 +16,9 @@ class Account::TransactionsController < ApplicationController
 
     respond_to do |format|
       format.html { redirect_to account_entry_path(@account, @entry), notice: t(".success") }
-      format.turbo_stream { render turbo_stream: turbo_stream.replace(@entry) }
+      format.turbo_stream {
+        render turbo_stream: turbo_stream.replace(@entry, partial: @entry.to_partial_path, locals: { entry: @entry, **partial_options })
+      }
     end
   end
 
@@ -53,5 +55,15 @@ class Account::TransactionsController < ApplicationController
                 permitted_params[:amount] = amount_value
               end
             end
+    end
+
+    def partial_options
+      params
+        .require(:partial_options)
+        .permit(:selectable)
+        .to_h
+        .each_with_object({}) do |(k, v), acc|
+          acc[k.to_sym] = v
+        end
     end
 end
--- a/app/views/account/transactions/_transaction.html.erb
+++ b/app/views/account/transactions/_transaction.html.erb
@@ -1,6 +1,8 @@
 <%# locals: (entry:, selectable: true, editable: true, short: false, show_tags: false, **opts) %>
 <% transaction, account = entry.account_transaction, entry.account %>
 
+<% selectable = ActiveModel::Type::Boolean.new.cast(selectable) %>
+
 <div class="grid grid-cols-12 items-center text-gray-900 text-sm font-medium p-4">
   <% name_col_span = unconfirmed_transfer?(entry) ? "col-span-10" : short ? "col-span-6" : "col-span-4" %>
   <div class="pr-10 flex items-center gap-4 <%= name_col_span %>">
@@ -21,7 +23,7 @@
             <%= content_tag :p, entry.name %>
           <% else %>
             <%= link_to entry_name(entry),
-                        account_entry_path(account, entry),
+                        account_entry_path(account, entry, selectable:),
                         data: { turbo_frame: "drawer", turbo_prefetch: false },
                         class: "hover:underline hover:text-gray-800" %>
           <% end %>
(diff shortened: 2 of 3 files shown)


app/controllers/account/transactions_controller.rb
```diff
@@ -54,4 +56,14 @@ def entry_params
               end
             end
     end
+
+    def partial_options
+      params
+        .require(:partial_options)
+        .permit(:selectable)
```
nicogaldamez, line 63: I wasn’t sure if it was worth adding all the other locals that the transaction partial receives (editable, short, show_tags, etc.).

zachgoll (commented):
This definitely works, but I'm wondering if we're solving the wrong problem here?

To me, having to pass all of these complicated params conditionally based on the controller is a sign that we should probably be simplifying something in the flow.

What if we made it so that `editable: false` rows cannot open the edit drawer?  To me, that would be a reasonable tradeoff here for the sake of simplicity since users should generally be going to the account page or "Transactions" page to do lots of editing anyways:

```diff
maybe $ git diff app/views/account/transactions/_transaction.html.erb
diff --git a/app/views/account/transactions/_transaction.html.erb b/app/views/account/transactions/_transaction.html.erb
index 475ab25..4e033f9 100644
--- a/app/views/account/transactions/_transaction.html.erb
+++ b/app/views/account/transactions/_transaction.html.erb
@@ -17,7 +17,7 @@
         </div>
 
         <div class="truncate text-gray-900">
-          <% if entry.new_record? %>
+          <% if entry.new_record? || !editable %>
             <%= content_tag :p, entry.name %>
           <% else %>
             <%= link_to entry_name(entry),
```

Also totally open to hearing other ways we could simplify all of this (i.e. changing partials / controllers etc.)?

nicogaldamez:
>@zachgoll said:
This definitely works, but I'm wondering if we're solving the wrong problem here?
To me, having to pass all of these complicated params conditionally based on the controller is a sign that we should probably be simplifying something in the flow.
What if we made it so that editable: false rows cannot open the edit drawer? To me, that would be a reasonable tradeoff here for the sake of simplicity since users should generally be going to the account page or "Transactions" page to do lots of editing anyways:

I couldn't agree more with this. It felt really odd to pass all those complicated params, but I didn’t want to change all the existing logic. The problem I see with your approach is that—unless there’s a way I’m unaware of—the user won’t be able to see all the transaction details if the transaction isn’t editable.

What do you think about changing the show view to display the data while having a separate edit view with the form? We could then include conditional links based on the editable parameter.

zachgoll:
@nicogaldamez yeah, no worries!  Anything is fair game here; simplification improvements are always welcome, even if they touch a good bit of code :)

> What do you think about changing the show view to display the data while having a separate edit view with the form? We could then include conditional links based on the editable parameter.

My gut reaction is that this may result in quite a bit of code duplication for something that will _only_ be used on the dashboard.  In that case, I like the tiny diff I had suggested above.  But maybe I'm misunderstanding what you're suggesting here?

nicogaldamez:
@zachgoll you're probably right. We could go with your suggestion—it's simpler and straightforward.

My idea was to have something like this in `transactions/show.html.erb`.
image

And create a new file (`transactions/edit.html.erb`) with the current view with the form:
image

zachgoll:
@nicogaldamez yeah I think I'd probably prefer to not have the view duplication mainly for these reasons:

1. A fully "read-only" view would only be shown on the dashboard, and I can't foresee a scenario where we'd use it elsewhere
2. In the future, we'll need to support "partially editable" forms based on the account type (manual vs. connected)
3. Since `Account::Entry` is a delegated type, we'd have to duplicate this pattern across 3 separate views (`Account::Transaction`, `Account::Valuation`, `Account::Trade`)

I'd say we either go with my proposed solution or consider splitting out into more controllers that give fine-grained controls around the responses after editing

nicogaldamez:
I close this PR. The changes suggested were applied in https://github.com/maybe-finance/maybe/pull/1394.

Issue #1225, opened 2024-10-01
I18n (Internationalization) of the Maybe App


zachgoll:
The purpose of this issue is twofold: 

1. Act as a placeholder for a feature we intend to support _in the future_
2. Provide documentation around how we're thinking about I18n (localization) of the app

## What is I18n?

I18n (Internationalization) is a crucial aspect of our global personal finance app. It involves adapting the software to support various languages and regional differences without changing the codebase. This process is essential for creating a "native" feel for users worldwide.

### Key Components of I18n:

1. **Text Localization**: Translating all user-facing text in the app to the user's preferred language.
2. **Currency Formatting**: Displaying monetary values in the appropriate format for each locale.
3. **Date and Time Formatting**: Presenting dates and times according to local conventions.
4. **Number Formatting**: Adapting number representations (e.g., decimal separators, thousands separators) to match local standards.

### Challenges and Considerations:

- **Comprehensive Coverage**: Identifying and localizing all text strings throughout the app, including error messages, tooltips, and dynamic content.
- **Context Preservation**: Ensuring translations maintain the original meaning and context.
- **UI Adaptability**: Designing flexible layouts to accommodate varying text lengths in different languages.
- **Pluralization Rules**: Handling complex pluralization rules that differ across languages.
- **Maintenance**: Keeping translations up-to-date as the app evolves and new features are added.
- **Testing**: Implementing thorough testing processes to verify the correctness and consistency of localized content.

There's a lot more to it than meets the eye, so we thought it would be helpful to the community to walk through our near-term and long-term plans for full I18n support.

## What is Maybe's plan for I18n?

We are a small, English-speaking COSS startup.  Because of our team size and limited runway, we have to be conscious that we are spending our time on the high value personal finance features that help you manage your finances more effectively.

Below outlines our plan to support a variety of user locales. 

### Near-term I18n support

In the near term, we intend to support **multi-currency and date/time formatting** for the following locales, which aligns with the `I18n.available_locales` from the [i18n gem](https://github.com/ruby-i18n/i18n):

```
[...]

PR #1393, opened 2024-10-30
Auto naming of Transfer Transaction

### LINKED ISSUE #1384: Improvement: Auto-naming of transfer transactions
Right now, we're providing a "Description" field on the transfer form:

CleanShot 2024-10-28 at 10 02 00

Instead, we should _remove_ this field entirely and generate the name automatically on the backend when the transfer is created.

The name should be: `Transfer from {originator} to {receiver}` and should be based on the accounts that were selected.

This will:

1. Reduce input required from user
2. Be less generic, easier to understand

Harry-kp:
#### Why?
- Fixes #1384 

#### What?
- Removed the Description field from the transfer form.
- Removed redundant locale to fix the test
- Started saving Account::Entry name as  `Transfer from {from_account} to {to_account}` .Earlier it was saved as `Transfer from Originator to Receiver` or whatever user inputs in the Transfer form.

#### Probable Concerns
- After removing the description field from the transfer form, there is a slight UI shift when moving from expense form to the Transfer form, I am not fixing this in this PR though

https://github.com/user-attachments/assets/344791ae-16b0-4034-9cbc-4100b4837282
- This change might be needing a test case, though I am not so familiar with the testing part and also not able to find any instructions in the README.

### DIFF
--- a/app/models/account/entry.rb
+++ b/app/models/account/entry.rb
@@ -28,8 +28,6 @@ class Account::Entry < ApplicationRecord
       .where("er.rate IS NOT NULL OR account_entries.currency = ?", currency)
   }
 
-  before_save :assign_transfer_name, if: -> { marked_as_transfer && transfer.present? }
-
   def sync_account_later
     if destroyed?
       sync_start_date = previous_entry&.date
@@ -224,9 +222,5 @@ class Account::Entry < ApplicationRecord
         current: amount_money,
         previous: previous_entry&.amount_money,
         favorable_direction: account.favorable_direction
-    end
-
-    def assign_transfer_name
-      self.name = transfer.name
     end
 end
--- a/app/models/account/entry.rb
+++ b/app/models/account/entry.rb
@@ -28,6 +28,8 @@ class Account::Entry < ApplicationRecord
       .where("er.rate IS NOT NULL OR account_entries.currency = ?", currency)
   }
 
+  before_save :assign_transfer_name, if: -> { marked_as_transfer && transfer.present? }
+
   def sync_account_later
     if destroyed?
       sync_start_date = previous_entry&.date
@@ -222,5 +224,9 @@ class Account::Entry < ApplicationRecord
         current: amount_money,
         previous: previous_entry&.amount_money,
         favorable_direction: account.favorable_direction
+    end
+
+    def assign_transfer_name
+      self.name = transfer.name
     end
 end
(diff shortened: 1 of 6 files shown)


app/models/account/entry.rb
```diff
@@ -223,4 +225,8 @@ def create_trend
         previous: previous_entry&.amount_money,
         favorable_direction: account.favorable_direction
     end
+
+    def assign_transfer_name
+      self.name = transfer.name
+    end
```
zachgoll, line 231: I think we should move this logic directly to the transfers controller:

https://github.com/maybe-finance/maybe/blob/490f44589e846e5b5d2b6bf56ddb4382234f0067/app/controllers/account/transfers_controller.rb#L18

Furthermore, it may be a nice touch if we had generated names for each of the entries within the transfer:

https://github.com/maybe-finance/maybe/blob/490f44589e846e5b5d2b6bf56ddb4382234f0067/app/models/account/transfer.rb#L46-L64

For example, the "from" name could be something like `"Transfer to #{to_account.name}"` and the "to" name could be `"Transfer from #{from_account.name}"`

harrrykp: Yes , This makes sense. 
I believe if we’re assigning distinct names to both entries, then the `name` parameter in `build_from_accounts` function would be of no use, so removing it seems like the right approach to me

Also, Are we going to store the full name i.e Transfer from {from account} to {to account} elsewhere or this will going to be generated at runtime, like we do now.

Harry-kp: Yes , This makes sense.
I believe if we’re assigning distinct names to both entries, then the name parameter in build_from_accounts function would be of no use, so removing it seems like the right approach to me

Also, Are we going to store the full name i.e Transfer from {from account} to {to account} elsewhere or this will going to be generated at runtime, like we do now.

Harry-kp: Also, I've noticed that locales are already used in the project, so directly adding `"Transfer to #{to_account.name}"` might not be ideal. Would it be better to create a locale attribute under the `entry` folder in `en.yml`, or would that be overkill for this change?

zachgoll: @Harry-kp regarding locales, we're going to have to go back through a lot of the model stuff and make bulk updates once we start down the path of #1225 , so for this change, let's just hardcode it for now.

And in regards to the `name:` param, fully agree.

> Are we going to store the full name i.e Transfer from {from account} to {to account} elsewhere

That's my bad, I forgot that `Transfer` does not have a stored `name` field.  We can just keep generating at runtime as we're doing now.

Harry-kp: @zachgoll  Changes are done.

zachgoll (approved):
Awesome, looks great!

PR #1397, opened 2024-10-31
Family invites

Shpigford:
Adds the ability to invite family members to your account.

- [x] Figure out how this works in a self-hosted environment
- [x] Make sure email formatting looks passable
- [x] Update add member modal to match Figma design
- [x] Show/format pending invites in the member list
- [x] Write some basic tests


Shpigford:
@zachgoll Thoughts on the self-hosted version of this? Instead of email invites we could just have the admin add a user with a password (then it's their responsibility to give that to whomever they want)? Other option is just show them the invitation URL for them to copy/paste and share on their own?

app/controllers/onboardings_controller.rb
```diff
@@ -1,19 +1,15 @@
...
   before_action :set_user
-
-  def show
-  end
-
-  def profile
-  end
-
-  def preferences
-  end
```
zachgoll, line 13: Can we keep these?  I know it's a bit redundant but if we've got a route for them I think having the action (even if empty) provides good documentation for future devs

app/controllers/registrations_controller.rb
```diff
@@ -4,36 +4,51 @@ class RegistrationsController < ApplicationController
...
   private
 
+    def load_invitation
+      token = params[:invitation] || params.dig(:user, :invitation)
+      @invitation = Invitation.pending.find_by!(token: token)
+    end
+
+    def invitation_token?
+      params[:invitation].present? || params.dig(:user, :invitation).present?
+    end
```
zachgoll, line 44: Think we can simplify this a bit and remove `invitation_token?`:

```rb
before_action :set_invitation

def set_invitation
  token = params[:invitation] || user_params(:invitation)
  @invitation = Invitation.pending.find_by(token: token)
end
```

zachgoll (commented):
Implementation looks good. 

In regards to self hosters, I think this should work fine

- If a self hoster wants to invite a member to their family, they use this flow
- If a self hoster wants to allow someone else to sign up for their instance, they use the invite codes flow

CleanShot 2024-10-31 at 08 44 48

zachgoll:
@Shpigford I guess the one consideration for self hosting that may not work with the email config is users running things on a local network.

Rather than having to worry about SMTP stuff, I think I like your idea of just generating a URL with a token that they can copy/paste and send to whoever they want.  Since we've got an expiry on the `Invitation` that seems reasonable.

PR #1407, opened 2024-11-02
Rails 8 upgrade

3zcurdia:
# Changelog

https://edgeguides.rubyonrails.org/8_0_release_notes.html


zachgoll:
@3zcurdia appreciate the setup here, but I think we're a bit early to do this upgrade.  We will eventually be moving towards Rails 8, but would like to stick with the stable `7.2` for a bit longer.

PR #1427, opened 2024-11-06
Financial AI

Shpigford:
This is the start on integrating AI in to the app.

It's still very early and right now the major focus is on security and choice (making sure it's 100% opt-in).


zdehasek:
Are you planning enable Ollama too ?

Shpigford:
> Are you planning enable using Ollama too ?

No immediate plans.

app/models/message.rb
```diff
@@ -0,0 +1,4 @@
+class Message < ApplicationRecord
```
chadwilken, line 1: Is it worth making the `role` and `status` an enum? Rails supports string enum values now.

config/initializers/openai.rb
```diff
@@ -0,0 +1,6 @@
+if ENV.fetch("OPENAI_ACCESS_TOKEN", nil).present?
+  OpenAI.configure do |config|
+    config.access_token = ENV.fetch("OPENAI_ACCESS_TOKEN")
+    config.log_errors = true if Rails.env.development?
```
chadwilken, line 4: You can do the following to simplify
```suggestion
    config.log_errors = Rails.env.development?
```

app/views/chats/_message.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    <% else %>
+      <div class="w-8 h-8">
+        <%= render "shared/circle_logo", name: "AI", hex: "#22CCEE", size: "full" %>
+      </div>
+    <% end %>
+  </div>
+  <div class="flex-1">
+    <div class="inline-block max-w-full">
+      <div class="text-sm prose text-gray-900" id="message_content_<%= message.id %>">
+        <% if message.content == '...' %>
```
chadwilken, line 16: You could move some of the logic to the model so the magic `'...'` string logic isn't in several different places. Could be something like `message.pending_message?` or something like that 🤷🏻.

app/views/chats/_message.html.erb
```diff
@@ -0,0 +1,41 @@
...
+              <div class="dot"></div>
+              <div class="dot"></div>
+              <div class="dot"></div>
+            </div>
+            <div class="text-xs text-gray-400">
+              <% if message.status == 'pending' %>
+                Determing the best route to answer your question.
+              <% elsif message.status == 'data' %>
+                Grabbing the necessary information to answer your question.
+              <% elsif message.status == 'processing' %>
```
chadwilken, line 28: These become predicate methods if you use an enum.

app/views/layouts/with_sidebar.html.erb
```diff
@@ -8,13 +8,20 @@
       <% end %>
     </div>
 
-    <main class="grow px-20 py-6 h-full <%= require_upgrade? ? "relative overflow-hidden" : "overflow-y-auto" %>">
+    <main class="grow py-6 h-full <%= require_upgrade? ? "relative overflow-hidden" : "overflow-y-auto" %>">
```
chadwilken, line 11: You can use the `class_names` helper here:
```suggestion
    <main class="<%= class_names("grow py-6 h-full", {"relative overflow-hidden" => require_upgrade?, "overflow-y-auto" => !require_upgrade?}) %>">
```

app/views/shared/_circle_logo.html.erb
```diff
@@ -9,5 +9,6 @@
 
 <%= tag.div style: mixed_hex_styles(hex),
             class: [size_classes[size], "flex shrink-0 items-center justify-center rounded-full"] do %>
-  <%= tag.span name[0].upcase, class: ["font-medium", size == "sm" ? "text-xs" : "text-sm"] %>
+  <%= tag.span name.length <= 2 ? name.upcase : name[0].upcase, 
+               class: ["font-medium", size == "sm" ? "text-xs" : "text-sm"] %>
```
chadwilken, line 13: Could use `class_names` here if you wanted to as well.

app/views/chats/show.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= turbo_frame_tag "chat_content" do %>
+  <div class="flex flex-col h-full">
+    <div id="chat_messages" class="grow overflow-y-auto">
+      <% @chat.messages.where.not(content: nil).where.not(content: "...").where.not(hidden: true).order(created_at: :asc).each do |message| %>
```
chadwilken, line 4: If you move the magic string to the `Message` model you could make this a scope to filter it out.

app/controllers/chats_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+  def create
+    @chat = Current.user.chats.create
+    @message = @chat.messages.create(
+      user: Current.user,
+      content: chat_params[:content],
+      role: "user"
+    )
+
+    # Stub reply from bot
+    reply = @chat.messages.create(
```
chadwilken, line 19: This could be moved to the model to clean things up a bit. Something like `Message.create_pending_assistant_message`

app/controllers/chats_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      content: params[:message][:content],
+      role: "user"
+    )
+
+    # Stub reply from bot
+    reply = @chat.messages.create(
+      content: "...",
+      user: nil,
+      role: "assistant"
+    )
```
chadwilken, line 52: If you move it to the model, that can be reused here.

app/jobs/chat_job.rb
```diff
@@ -0,0 +1,295 @@
+class ChatJob < ApplicationJob
+  queue_as :default
+
+  def perform(chat_id, message_id)
+    chat = Chat.find(chat_id)
+    nil if chat.nil?
```
chadwilken, line 6: This line won't ever be hit because `find` would raise and `ActiveRecord::RecordNotFound` error if there isn't a `Chat` for the `chat_id`.

app/jobs/chat_job.rb
```diff
@@ -0,0 +1,295 @@
...
+    if !viability["content_contains_answer"]
+      handle_non_viable_conversation(openai_client, chat, message, viability)
+    else
+      generate_response_content(openai_client, chat, message)
+    end
+  end
+
+  private
+    def determine_viability(openai_client, chat)
+      chat_history = chat.messages.where.not(content: [ nil, "" ]).where.not(content: "...").where.not(role: "log").order(:created_at)
```
chadwilken, line 22: If you extract the scope for `where.not(content: "...")` then you can reuse that here.

chadwilken (commented):
Really like what you have going on here. Not sure if you're actually looking for feedback but I was bored and looking over the repo and decided to give some suggestions. Feel free to tell me to piss off 😄.

asdasdad23332s:
Hi @Shpigford , curious about how the AI feature will work. At the moment, any estimate for the timing to have the first release?

asdasdad23332s:
This will work only with OpenAI? Or another LLM?

M123-dev:
I just found this interesting ruby gem

https://github.com/icebaker/ruby-nano-bots

Is a library that provides an interface to address multiple chatbots. 

Among them ChatGPT and Oolama. Also seems to make the whole configuration of assistants very easy.
I think it's a good idea to not only concentrate on ChatGPT.
- No dependency (good for maybe hosted if OpenAi raises prices)
- It's one thing to give you the data, but not everyone wants to give it to OpenAi on top of that.
- With more effort for the hardcore selfhosters it can also be made to run independently

Translated with DeepL.com (free version)

PR #1432, opened 2024-11-07
Exclude inactive accounts from net-worth calculation and from sidebar

### LINKED ISSUE #1419: Bug: "Disabled" accounts still show in sidebar, but shouldn't
Currently, users have the ability to "disable" an account in their account management view. 

This should:

1. Remove the account from all aggregate calculations (like net worth)
2. Remove the account from the sidebar
3. User can still click on the account to view its history

We _may_ want to consider an "archive" functionality as either an _alternative_ or _replacement_ to this existing functionality to be more clear how this should affect things.

https://github.com/user-attachments/assets/d65b6d82-0d0b-43da-8dd3-2ac5df7555a9

tonyvince:
Fix #1419


zachgoll (approved):
Awesome!

Issue #1418, opened 2024-11-05
Bug: When balance and transaction exist on same day, balances not correctly calculated


zachgoll:
**Describe the bug**

There are two things worth fixing here:

1. The balance calculation should result in `$100` 
2. The "balance update" should be based on the trend vs. _previous balance_ ($100), and should be downward.  Currently, it is an "upward" trend because we're comparing it against the previous known balance of $0.

![CleanShot 2024-11-05 at 10 35 26](https://github.com/user-attachments/assets/7fd5bfbb-00d7-4f97-b138-b431af347156)

mikr13:
Also why do we have to do a balance update manually? Shouldn't balance be calculated based on transactions/incomes in a day?

mikr13:
Yeah, that's fair. But what I noticed is that the transactions marked as expenses don't decrease the total balance in that account. Only when balance update is done then the account's balance changes.

PR #1435, opened 2024-11-08
Skip account valuation on entry balance_after_entry

bruno-costanzo:
### Why?
Partially Fixes [1418](https://github.com/maybe-finance/maybe/issues/1418)

When creating a transaction after an account valuation, the latter should not be considered in the calculation of the balance of the entry.

### What?
Change the entry model in order to skip the account valuation amount in the balance_after_entry
image

### What should we test?
Create an account valuation and a transaction after, and ensure the account valuation is not part of the calculation of the balance_after_entry within the context of the transaction.

### Pending
There is another error related with [Issue 1418](https://github.com/maybe-finance/maybe/issues/1418), related to the way in which the amount of an account valuation is calculated, which in turn can cause the incorrect icon to be displayed.

Paired with @nicogaldamez

### DIFF
--- a/app/helpers/account/entries_helper.rb
+++ b/app/helpers/account/entries_helper.rb
@@ -14,9 +14,9 @@ module Account::EntriesHelper
 
   def entries_by_date(entries, selectable: true, totals: false)
     entries.group_by(&:date).map do |date, grouped_entries|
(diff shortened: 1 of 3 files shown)


zachgoll (approved):
Thanks for the fix here!

PR #1438, opened 2024-11-08
Fixes: Sidebar only displays cash balance for investment accounts, not full value

### LINKED ISSUE #1369: Bug: Sidebar only displays cash balance for investment accounts, not full value
**Describe the bug**
The balance on the sidebar doesn't match the actual balance

**To Reproduce**
1. Go to the dashboard
2. See some accounts that don't show the proper balance on the sidebar

Unfortunately I don't know how it got like this, but happy to help debug if needed.

**Expected behavior**
I expect the sidebar balance to show the actual balance.

Hosted

MacOS 14, Chrome 130

**Screenshots / Recordings**
image

donjuma:
fixes #1369 

PR fixes current issue with sidebar only showing cash value and not full account value.

Screenshot 2024-11-08 at 3 20 44 PM

### DIFF
--- a/app/models/family.rb
+++ b/app/models/family.rb
@@ -109,7 +109,7 @@ class Family < ApplicationRecord
   end
 
   def assets
-    Money.new(accounts.active.assets.map { |account| account.balance_money.exchange_to(currency, fallback_rate: 0) }.sum, currency)
+    Money.new(accounts.active.assets.map { |account| account.value }.sum, currency)
   end
 
   def liabilities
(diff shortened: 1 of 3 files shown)


app/models/family.rb
```diff
@@ -109,7 +109,7 @@ def net_worth
   end
 
   def assets
-    Money.new(accounts.active.assets.map { |account| account.balance_money.exchange_to(currency, fallback_rate: 0) }.sum, currency)
```
zachgoll, line 112: In these two places, we will need to call `account.value.exchange_to(currency, fallback_rate: 0)` so that users with accounts in multiple currencies can be "rolled up" properly.

zachgoll:
Closing as this should be fixed via #1526

PR #1439, opened 2024-11-08
Replaced Native Scrollbars with Tailwind Scrollbars

jestinjoshi:
Implements #1247 

## Changes Made:
- Replaced native scrollbars with Tailwind scrollbars for consistency
- Added hover effect on 'Delete Account' button

cc: @justinfar

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,6 +4,7 @@ module ApplicationHelper
   def date_format_options
     [
       [ "DD-MM-YYYY", "%d-%m-%Y" ],
+      [ "DD.MM.YY", "%d.%m.%Y" ],
       [ "MM-DD-YYYY", "%m-%d-%Y" ],
       [ "YYYY-MM-DD", "%Y-%m-%d" ],
       [ "DD/MM/YYYY", "%d/%m/%Y" ],
--- a/app/views/import/configurations/_trade_import.html.erb
+++ b/app/views/import/configurations/_trade_import.html.erb
@@ -3,7 +3,7 @@
 <%= styled_form_with model: @import, url: import_configuration_path(@import), scope: :import, method: :patch, class: "space-y-2" do |form| %>
   <div class="flex items-center gap-2">
     <%= form.select :date_col_label, import.csv_headers, { include_blank: "Leave empty", label: "Date" }, required: true %>
-    <%= form.select :date_format, [["DD-MM-YYYY", "%d-%m-%Y"], ["MM-DD-YYYY", "%m-%d-%Y"], ["YYYY-MM-DD", "%Y-%m-%d"], ["DD/MM/YYYY", "%d/%m/%Y"], ["YYYY/MM/DD", "%Y/%m/%d"], ["MM/DD/YYYY", "%m/%d/%Y"]], label: true, required: true %>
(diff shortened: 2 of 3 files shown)


zachgoll:
@jestinjoshi @justinfar this will add persistent scrollbars to the app.  Are we okay with that?

CleanShot 2024-11-11 at 09 55 30

@jestinjoshi I think rather than applying these across many places, we should just apply this globally:

```css
*::-webkit-scrollbar {
  width: 4px;
}

*::-webkit-scrollbar-thumb {
  background: #d6d6d6;
  border-radius: 10px;
}

*::-webkit-scrollbar-thumb:hover {
  background: #a6a6a6;
}
```

Shpigford:
Not a fan of the persistent scrollbars and not sure cross-browser/platform styling of the scrollbars is a big priority.

justinfar:
yeah I'm not a fan of the persistent scrollbar either

I'd lean more towards making this change about refining the scrollbar to be consistent with Tailwind rather than keep it showing up whenever we have a scrollable area

jestinjoshi:
Oops, I closed this PR by mistake, but honestly, it was not the best solution to the problem anyway.

@zachgoll, I like the global styling approach. It’s a great way to future-proof.

Regarding the persistent scrollbar, I think the difference here is that all of you are on macOS, where scrollbars are handled pretty smoothly by default. I’m on Windows, where big, chunky scrollbars can be a bit of an eyesore, especially on an otherwise beautiful UI. If Maybe is aiming to serve a global user base, it might be worth considering the Windows experience too.

For priority, I agree with @Shpigford that it’s probably not a blocker/high-priority task. It’s up to you whether to tackle it now or later.

cc: @justinfar @zachgoll @Shpigford

zachgoll:
@jestinjoshi what browser are you using?

jestinjoshi:
@zachgoll I'm using Chrome on Windows.

I've added the screenshots here: https://github.com/maybe-finance/maybe/discussions/1247#discussion-7282634

Just to clarify, by persistent scrollbars, do you mean the scrollbar stays visible even if the UI isn’t overflowing?

zachgoll:
@jestinjoshi no, when I say "persistent", I'm referring to the case where there is overflowing content and the scrollbar persists even when the user is not actively scrolling the container (i.e. "auto" behavior):

https://github.com/user-attachments/assets/dab887f3-cca4-4558-aaae-bbdfd69333c9

If we can find an simple/easy way to preserve this behavior with the custom scrollbars I think it makes sense to do.

jestinjoshi:
Thanks for clarifying. I just wanted to confirm that I understood your point on persistent scrollbars correctly.

This behaviour is not available on Windows. The scrollbar is always visible, even when the user isn’t actively scrolling within a container. Also, each browser on Windows handles scrollbar styling differently, leading to inconsistency. I’ve added a video below showing how the scrollbar looks in Chrome on Windows.

I think one way to tackle this issue would be to apply custom scrollbar styling only for Windows 🤔 

https://github.com/user-attachments/assets/e78c451a-7a10-4120-a689-411bf52e6304

zachgoll:
@jestinjoshi I understand this is not available on Windows, but by adding these changes, we are _removing_ it from Mac too, which is not ideal.  My main concerns are preserving the "auto" behavior on Mac and keeping the solution here very simple. 

The best solution is something that allows for custom scrollbars while preserving "auto" behavior on Mac.  As a fallback, I agree that we could do a "Windows only" custom scrollbar if the implementation is simple and straightforward.

jestinjoshi:
@zachgoll I agree with preserving the Mac behaviour. Implementing 'Windows-only' styling should be straightforward.

I'll start a new PR for this since the current one is merging commits from my `main` branch.

PR #1445, opened 2024-11-10
Adds a common DE date format

sbehrends:
## Overview

This PR introduces a new date format which is very frequent in Germany Bank exports.
`DD.MM.YY`

Example of a popular Germany Bank
Screenshot 2024-11-10 at 21 00 45

## Changes

- Adds a new date format to import configuration

Screenshot 2024-11-10 at 20 59 29


zachgoll:
@sbehrends thanks!  Good idea.

PR #1448, opened 2024-11-11
Fix the value shown on a balance update transaction

### LINKED ISSUE #1447: Bug: Amount in the balance edit screen doesn't show proper decimal points
**Describe the bug**
When editing a balance transaction the decimals are incorrect.

**To Reproduce**
1. Create a balance update in USD for $100.00
2. Click on the transaction to edit
3. Notice that the amount shown is $100.0

**Expected behavior**
I would expect this field to respect the proper steps of the currency

Hosted

**Screenshots / Recordings**
CleanShot-000281 2024-11-10 at 19 33 14@2x

ahatzz11:
I noticed today that when editing a balance update transaction the amount shown didn't respect the currency steps. After some playing around and looking at the normal transaction (which worked correctly), I believe this fix is appropriate. I tested this locally and it works great (I tested USD and JPY for different precision values from currencies.yml)

This is where I saw the `value` field for the `money_field` of a normal transaction: https://github.com/maybe-finance/maybe/blob/56ab092f6b89fa4873d97c285811b137c3918717/app/views/account/transactions/show.html.erb#L57

Before/After:
<p>
image
image
</p>
Fixes #1447

### DIFF
--- a/app/views/account/valuations/show.html.erb
+++ b/app/views/account/valuations/show.html.erb
@@ -40,7 +40,8 @@
           <%= f.money_field :amount,
                 label: t(".amount"),
                 auto_submit: true,
-                disable_currency: true %>
+                disable_currency: true,
+                value: entry.amount.abs %>
         <% end %>
       </div>
     <% end %>


zachgoll:
@ahatzz11 we probably should avoid calling `abs` on this for the case where a user has entered a negative balance on the account (rare, but possible and valid).

Just pushed through a fix that addresses this at the form builder level using the `_money_field` partial:

https://github.com/maybe-finance/maybe/commit/31ecd3ccd4e1b7d1c4d2e63c9507579a101634a5

Issue #1466, opened 2024-11-16
Bug: Can not start maybe container on raspberrypi 4


niits:
**Describe the bug**

TypeError: no implicit conversion of nil into Hash (TypeError) raised when start Maybe stacks defined on `docker-compose.example.yml`

**To Reproduce**
1. Set up Maybe with Docker by following <https://github.com/maybe-finance/maybe/blob/main/docs/hosting/docker.md>

**Expected behavior**

Maybe container show be started

**What version of Maybe are you using?**
Self-hosted, v0.1.0

**What operating system and browser are you using?**

Raspberrypi 4, Ubuntu 24.04.1 LTS

**Screenshots / Recordings**

```bash
Attaching to app-1, postgres-1
postgres-1  | 
postgres-1  | PostgreSQL Database directory appears to contain a database; Skipping initialization
postgres-1  | 
postgres-1  | 2024-11-16 08:52:47.416 UTC [1] LOG:  starting PostgreSQL 16.5 (Debian 16.5-1.pgdg120+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
postgres-1  | 2024-11-16 08:52:47.418 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
postgres-1  | 2024-11-16 08:52:47.418 UTC [1] LOG:  listening on IPv6 address "::", port 5432
postgres-1  | 2024-11-16 08:52:47.432 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
postgres-1  | 2024-11-16 08:52:47.498 UTC [30] LOG:  database system was interrupted; last known up at 2024-11-16 08:52:11 UTC
postgres-1  | 2024-11-16 08:52:47.724 UTC [30] LOG:  database system was not properly shut down; automatic recovery in progress
postgres-1  | 2024-11-16 08:52:47.767 UTC [30] LOG:  invalid record length at 0/19123C0: expected at least 24, got 0
postgres-1  | 2024-11-16 08:52:47.767 UTC [30] LOG:  redo is not required
postgres-1  | 2024-11-16 08:52:47.837 UTC [28] LOG:  checkpoint starting: end-of-recovery immediate wait
[...]

PR #1469, opened 2024-11-17
Add Active Record Encryption keys to environment configuration and docs

acflint:
Starting a PR with environment config and documentation updates to address activerecord encryption setup.

Feel free to share feedback, or let me know if I missed anything. Or if you want to address activerecord encryption setup in some other way, I'm happy to follow!

Fix for this issue: https://github.com/maybe-finance/maybe/issues/1466

Thanks!


zachgoll:
@acflint appreciate the diagnosis and quick fix for this!  I think we should go with #1476 to keep the self hosting setup as low-overhead as possible.

Issue #1462, opened 2024-11-14
CSV not importable


RMStyle:
**Describe the bug**
I have csv from my bank. I modified it to fit to the requirements of import. When I try to import it fails,

**To Reproduce**
1. Go 'Import'
2. Click on 'Import Transactions'
3. Click on 'Upload csv'
4. Choose attached file
5. Choose 'Semicolon' at Dropdown
6. Click on 'Upload csv'
7. See error

**Expected behavior**
Data should correctly be importetd

**What version of Maybe are you using?**
Selfhosted Version: v0.2.0-alpha.2

**What operating system and browser are you using?**
Fedora 41;  Firefox;

[testdata_maybe.csv](https://github.com/user-attachments/files/17758761/testdata_maybe.csv)

acflint:
I've opened a PR to fix this: https://github.com/maybe-finance/maybe/pull/1468

PR #1470, opened 2024-11-17
Allow custom column separator for CSV parsing in uploads controller

acflint:
Fix for this issue: https://github.com/maybe-finance/maybe/issues/1462

`csv_valid?` wasn't using the submitted `col_sep` argument, so semicolon separated CSVs were failing to be validated.

### DIFF
--- a/app/controllers/import/uploads_controller.rb
+++ b/app/controllers/import/uploads_controller.rb
@@ -32,7 +32,7 @@ class Import::UploadsController < ApplicationController
       require "csv"
 
       begin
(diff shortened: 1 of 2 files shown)


zachgoll:
@acflint disregard my comment on the other PR that was closed.  This looks good!

acflint:
@zachgoll I can fix the failing specs, if you'd like. Let me know.

zachgoll:
@acflint yep, that would be great.  Looks like we just need to pass `col_sep` as params for each of those failing tests

PR #1490, opened 2024-11-22
Feature: Add support for customized synth URL from env variable

Evlos:
Usage:

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: maybe
spec:
  template:
    spec:
      containers:
      - name: maybe
        image: evlos/maybe:latest
        env:
        - name: SYNTH_URL
          value: http://192.168.233.107:30800
```

### DIFF
--- a/app/models/provider/synth.rb
+++ b/app/models/provider/synth.rb
@@ -174,7 +174,7 @@ class Provider::Synth
     SecurityInfoResponse = Struct.new :info, :success?, :error, :raw_response, keyword_init: true
 
     def base_url
-      "https://api.synthfinance.com"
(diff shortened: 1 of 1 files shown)


zachgoll:
Hey @Evlos, can you explain this change a bit further?  I'm guessing your use case here involves some sort of incoming traffic proxy?

Typically, we only accept contributions that would apply to a broad range of users but want to make sure I understand this better first.

zachgoll:
@Evlos thanks for the explanation!  That makes sense.  We definitely want to provide support for various regions like this.

PR #1492, opened 2024-11-22
Fix bug: Loan % doesn't allow exact rate

### LINKED ISSUE #1487: Bug: Loan % Doesn't allow exact Rate
**Describe the bug**
I have a mortgage of 6.875%, most rates seem to go by 1/8ths for my bank. I am unable to enter this value. I'm not sure how much .005% changes calculations but it would be great if we could be exact.

image

**To Reproduce**
1. Create a loan
2. Enter a rate of 6.875
3. See error

**Expected behavior**
I'd expect to be able to enter my exact loan rate.

Hosted

MacOS 14, Chrome 140

arsenstorm:
Fixes #1487

Change the step of the interest rate field to 0.005.

It's unlikely that we'll see interest rates in smaller increments.


zachgoll (approved):
Agreed, this seems like a reasonable limit to use.

ahatzz11:
Thanks!

PR #1493, opened 2024-11-22
Replaced Native Scrollbars with Tailwind Scrollbars on Windows

jestinjoshi:
Continuing from #1439 

## Changes Made:
- Added a new `detect_os` function in [application_controller.rb](https://github.com/maybe-finance/maybe/compare/main...jestinjoshi:maybe:windows-scrollbar?expand=1#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07)
- The detected OS will be added as a class to the `html` tag
- Based on the added class, the scrollbar styling will be applied (can be scaled to other OS if required)

### DIFF
--- a/app/assets/stylesheets/application.tailwind.css
+++ b/app/assets/stylesheets/application.tailwind.css
@@ -163,3 +163,19 @@
     background: #a6a6a6;
   }
 }
+
+/* Custom scrollbar implementation for Windows browsers */
+.windows {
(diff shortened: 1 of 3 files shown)


zachgoll (approved):
Nice and simple, I like it!

jestinjoshi:
> Nice and simple, I like it!

Thank you! 😊

PR #1494, opened 2024-11-24
Use account currency instead of hardcoded USD

nikhilbadyal:
Currently USD is hardcoded. Causing it be shown on UI even when User Default currency or Account currency is not USD.

### DIFF
--- a/app/views/account/trades/_form.html.erb
+++ b/app/views/account/trades/_form.html.erb
@@ -28,7 +28,7 @@
       </div>
 
       <div data-trade-form-target="priceInput">
-        <%= form.money_field :price, label: t(".price"), currency_value_override: "USD", disable_currency: true %>
+        <%= form.money_field :price, label: t(".price"), currency_value_override: @account.currency, disable_currency: true %>
       </div>
     </div>
 


nikhilbadyal:
It fixes this screen. This screen shows hard-coded USD. 

image

nikhilbadyal:
Hey @zachgoll , 

> our securities price provider returns all prices in `USD`

Are we talking about [this](https://docs.synthfinance.com/reference/openclose) here  ? If yes, the prices doesn't look USD and are definitely INR.

zachgoll:
@nikhilbadyal yep, Synth is what we're using.  What specific request is returning that?  Can you provide a symbol I can take a look at?

nikhilbadyal:
Here is the sample curl. 

```
curl --request GET \
     --url https://api.synthfinance.com/tickers/SHALBY/open-close \
     --header 'Authorization: Bearer <MyToken>' \
     --header 'accept: application/json'
```

Response
```
{
  "ticker": "SHALBY",
  "exchange": {
    "name": null,
    "mic_code": "XNSE",
    "acronym": "NSE",
    "country": "India",
    "country_code": "IN",
    "timezone": "Asia/Kolkata"
  },
  "prices": [
    {
      "date": "2024-11-25",
      "open": 222.1,
      "close": 224,
      "high": 231.15,
      "low": 219.4,
      "volume": 11091
    }
  ],
  "paging": {
    "prev": "/tickers/SHALBY/open-close?page=",
    "next": "/tickers/SHALBY/open-close?page=",
    "total_records": 1,
    "current_page": 1,
    "per_page": 100,
    "total_pages": 1
  },
  "meta": {
    "credits_used": 1,
    "credits_remaining": 927
  }
}
```

OHCL are INR only - Can be verified from exchange website [itself](https://www.nseindia.com/get-quotes/equity?symbol=SHALBY).

zachgoll:
@nikhilbadyal thank you for posting, very helpful.

cc @Shpigford it appears that Synth is returning non-USD prices in some cases.  I think we either need to normalize everything to `USD` or we should be throwing an ISO currency code in the hash that is returned for each price so that the consumers of the API have a way to identify.

PR #1502, opened 2024-11-25
Synth error handling

Shpigford:


Shpigford:
@zachgoll Not sure if I'm using the error stuff correctly. Mind taking a look?

zachgoll:
@Shpigford to get this working, there are a few additional steps:

First, we'll need to implement the `stale?` method on the issue:

```ruby
class Issue::SynthLimitExceeded < Issue
  include Providable # Give this class access to our configured Synth Provider

  def title
    "Synth API Credit Limit Exceeded"
  end
 
  # Required so our app knows if the error is still happening after the user has taken some action to fix it
  def stale?
    self.class.synth_provider.usage.utilization < 1
  end 
end
```

Then, we need to add a method in `issuable.rb` to observe this error:

```ruby 
# app/models/concerns/issuable.rb

def observe_synth_credits_overage
  observe_issue(Issue::SynthLimitExceeded.new)
end
```

And finally, we need to call `account.observe_synth_credits_overage` during the account sync process:

```ruby 
class Account
   include Providable 

   def sync_data(start_date: nil)
    update!(last_synced_at: Time.current)

    resolve_stale_issues

    # Observe issue if credits are over limit
    if self.class.synth_provider.usage.utilization >= 1 || true
      observe_synth_credits_overage
    end

    Balance::Syncer.new(self, start_date: start_date).run
    Holding::Syncer.new(self, start_date: start_date).run
  end
end
```

**Simpler solution**

I think given this use case, there are probably some simplifications that could be made to the "issues" flow and the `Provider::Synth` to make all of that more seamless.

In the meantime, a simpler solution would be to add a convenience method on `Family` and then conditionally show some alert message somewhere in the UI when they are over the limit:

```ruby 
class Family 
  include Providable 

  def synth_overage?
    self.class.synth_provider && self.class.synth_provider.usage.utilization >= 1
  end
end
```

```erb
<% if Current.family.synth_overage? %> 
  <div>Some alert to user!!!</div>
<% end %> 
```

Shpigford:
@zachgoll I definitely prefer the simple approach. 🙂 Will get that swapped out and have you review in a bit.

zachgoll:
@Shpigford yeah I agree.  Longer term, I think we will basically have 1 single `Issue::SynthIssue < Issue` that will cover _all_ the possible Synth errors and provide UIs for dealing with them, but for now it definitely needs some fine-tuning to avoid adding a bunch of unnecessary complexity

zachgoll (approved):
Yep, looks good

PR #1506, opened 2024-11-27
Improve account transaction, trade, and valuation editing and sync experience

zachgoll:
This PR aims to drastically improve the account entry editing experience.

The current state of the app works as-is, but is very cumbersome in some areas.  Current issues include:

- Drawer editors closing while editing
- Synced data not updating immediately in the UI
- Minimal feedback to the user when account data is syncing in the background
- Static entry info not updating as the user edits the form below it

Furthermore, there was a ton of code duplication across the entry controllers (valuation, transaction, trade) that has been consolidated to the `EntryableResource` concern, which follows a similar pattern to the `AccountableResource` concern introduced in #1406.  This provides default implementations for all CRUD actions for each "Entryable" while allowing customization as needed.


app/javascript/controllers/application.js
```diff
@@ -52,6 +52,6 @@ Turbo.setConfirmMethod((message) => {
       { once: true },
     );
   });
-});
+};
```
zachgoll, line 55: `setConfirmMethod` will soon be deprecated

app/javascript/application.js
```diff
@@ -1,3 +1,7 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails";
 import "controllers";
+
+Turbo.StreamActions.redirect = function () {
+  Turbo.visit(this.target);
+};
```
zachgoll, line 7: Custom stream action that can be used to trigger a "refresh" to the page after a successful form submission.  This is helpful because many of our forms are rendered in modals, which means for validations/errors, we need to leverage Turbo frames, but for successful submissions, we generally want a full refresh.  Since Turbo doesn't provide an easy native way to handle this behavior, this provides a low-overhead, simple workaround:

```ruby
redirect_target_url = request.referer || account_path(@entry.account)
        format.turbo_stream { render turbo_stream: turbo_stream.action(:redirect, redirect_target_url) }
```

PR #1515, opened 2024-12-03
Handle invalid API key

### LINKED ISSUE #1507: Bug: undefined method `>=' for nil
**Describe the bug**
`>=` check on a nil object.

**To Reproduce**
1. Go to [Self Hosting](http://localhost:3000/settings/hosting)
2. Enter Invalid Key
3. Go to Dashboard

**Expected behavior**
Application should not crash.

"Self-hosted". Latest Commit

**Screenshots / Recordings**
image

**Additional context**
Currently the API Synth Settings screen doesn't force user to enter valid key or prevent itself from saving invalid key into DB.

image

If someone miss types a invalid API key or If somehow key becomes invalid like the key was revoked.

The [usage](https://github.com/maybe-finance/maybe/blob/c3248cd796b3052e72502e7b18ee75d9467770c1/app/models/provider/synth.rb#L13) method will return a [UsageResponse](https://github.com/maybe-finance/maybe/blob/c3248cd796b3052e72502e7b18ee75d9467770c1/app/models/provider/synth.rb#L17) object on 401. 

The [synth_overage](https://github.com/maybe-finance/maybe/blob/c3248cd796b3052e72502e7b18ee75d9467770c1/app/models/family.rb#L158) tries to do `.utilization >= 100` on usage. Since usage can of of type `UsageResponse` which doesn't have utilization. It throws an Exception. 

Possible Fix -
```
def synth_overage?
    self.class.synth_provider&.usage&.utilization.to_i >= 100
end
```

and displaying a flash that key is invalid or similar.

nikhilbadyal:
Fixes #1507

### DIFF
--- a/app/models/family.rb
+++ b/app/models/family.rb
@@ -162,6 +162,10 @@ class Family < ApplicationRecord
     self.class.synth_provider&.usage&.utilization.to_i >= 100
   end
 
+  def synth_valid?
+    self.class.synth_provider&.healthy?
+  end
+
   def subscribed?
     stripe_subscription_status == "active"
   end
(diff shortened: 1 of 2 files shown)


zachgoll (approved):

PR #1518, opened 2024-12-09
Add locking features

rylin7:
**Summary**
This pull request introduces optimistic locking to critical models (Account::Entry, Account::Transaction, and Account::Balance) to prevent race conditions and ensure data integrity during concurrent updates.

**Key Changes**

- Added lock_version for optimistic locking.

- Created locking-safe methods (update_entry!, process_transaction!, update_balance!).

- Preserved all existing functionality, including scopes and utilities (e.g., daily_totals, bulk_update!).

### DIFF
--- a/app/models/account/balance.rb
+++ b/app/models/account/balance.rb
@@ -1,9 +1,13 @@
 class Account::Balance < ApplicationRecord
-  include Monetizable
+  self.locking_column = :lock_version
 
-  belongs_to :account
-  validates :account, :date, :balance, presence: true
-  monetize :balance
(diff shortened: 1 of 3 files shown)


zachgoll:
Hey @rylin7, thanks for the contribution.  While we may explore optimizations like this in the future, I think we're a bit early to be introducing these.  Adding locking mechanisms comes with added complexity and reduced readability, so I think we need to have a strong reason to introduce these.

Since this application is very read-heavy with a low risk of concurrent updates (at the most, 1-2 users + a background job will be writing at once), I don't think there's a strong enough reason to introduce these at the moment.

PR #1555, opened 2024-12-18
Initial pass at Plaid EU

Shpigford:


Shpigford:
@zachgoll First high-level pass at adding EU support. You've got a better handle of how the Plaid stuff works, so let me know if I'm off base here.

app/models/concerns/plaidable.rb
```diff
@@ -5,10 +5,18 @@ module Plaidable
...
   end
 
   private
     def plaid_provider
       self.class.plaid_provider
     end
+
+    def plaid_eu_provider
+      self.class.plaid_eu_provider
+    end
```
zachgoll, line 21: I'm fine having 2 clients like this, but the main update we'll need to make is being able to grab the correct client based on the `PlaidItem` we're dealing with so that we know how to sync the data.  For example, here:

https://github.com/maybe-finance/maybe/blob/68617514b004e71b8750f9506de2dc2a9629992a/app/models/plaid_item.rb#L59

I think in `Plaidable`, we will want something like:

```ruby
def plaid_provider_for(plaid_item)
  # `region` is not yet in the schema, so we'll have to add a migration for it and populate it when we create the `PlaidItem` for the first time
  if plaid_item.region == "EU"
    plaid_eu_provider
  else
    plaid_provider
  end
end
```

Then, in `plaid_item.rb` and other places we're referencing `plaid_provider`, we'll need to switch to `plaid_provider_for(plaid_item)`

app/views/accounts/new/_method_selector.html.erb
```diff
@@ -6,16 +6,35 @@
...
+        </button>
+      <% else %>
+        <%# Default US-only Link %>
+        <button data-controller="plaid" data-action="plaid#open modal#close" data-plaid-link-token-value="<%= @link_token %>" class="flex items-center gap-4 w-full text-center focus:outline-none focus:bg-gray-50 border border-transparent focus:border focus:border-gray-200 px-2 hover:bg-gray-50 rounded-lg p-2">
+          <span class="flex w-8 h-8 shrink-0 grow-0 items-center justify-center rounded-lg bg-alpha-black-50 shadow-[inset_0_0_0_1px_rgba(0,0,0,0.02)]">
+            <%= lucide_icon("link-2", class: "text-gray-500 w-5 h-5") %>
+          </span>
+          <%= t(".connected_entry") %>
+        </button>
+      <% end %>
```
zachgoll, line 37: Slight simplification suggestion:

```erb
<%# Default US-only Link %>
<button data-controller="plaid" data-action="plaid#open modal#close" data-plaid-link-token-value="<%= @link_token %>" class="flex items-center gap-4 w-full text-center focus:outline-none focus:bg-gray-50 border border-transparent focus:border focus:border-gray-200 px-2 hover:bg-gray-50 rounded-lg p-2">
  <span class="flex w-8 h-8 shrink-0 grow-0 items-center justify-center rounded-lg bg-alpha-black-50 shadow-[inset_0_0_0_1px_rgba(0,0,0,0.02)]">
    <%= lucide_icon("link-2", class: "text-gray-500 w-5 h-5") %>
  </span>
  <%= t(".connected_entry") %>
</button>

<%# EU Link %>
<% unless Current.family.country == "US"  %>
  <button data-controller="plaid" data-action="plaid#open modal#close" data-plaid-link-token-value="<%= Current.family.get_link_token(webhooks_url: webhooks_plaid_url, redirect_url: accounts_url, accountable_type: accountable_type.name, region: :eu) %>" class="flex items-center gap-4 w-full text-center focus:outline-none focus:bg-gray-50 border border-transparent focus:border focus:border-gray-200 px-2 hover:bg-gray-50 rounded-lg p-2">
    <span class="flex w-8 h-8 shrink-0 grow-0 items-center justify-center rounded-lg bg-alpha-black-50 shadow-[inset_0_0_0_1px_rgba(0,0,0,0.02)]">
      <%= lucide_icon("link-2", class: "text-gray-500 w-5 h-5") %>
    </span>
    <%= t(".connected_entry_eu") %>
  </button>
<% end %> 
```

zachgoll (commented):
I reached out to Plaid and they said there is no way to figure out whether an "Item" is US or EU just based on the Plaid Item data.

So the biggest thing we're missing currently here is that identifying piece of data stored on `PlaidItem`.  I think since there are only 2 possible clients, we could consider a field called `plaid_region` that is a `string` and in our model we could probably validate that too:

```ruby
class PlaidItem < ApplicationRecord
  enum :plaid_region, { us: "us", eu: "eu" }
end
```

Which would then allow us to easily access it like, `some_item.eu?` 

Otherwise, I _think_ EU should work largely the same as US and I believe our existing implementation will properly handle these multi-currency accounts fine.

Shpigford:
@zachgoll Take a look now.

db/schema.rb
```diff
@@ -460,6 +469,33 @@
...
+    t.text "content"
+    t.text "log"
+    t.string "role"
+    t.string "status", default: "pending"
+    t.boolean "hidden", default: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["chat_id"], name: "index_messages_on_chat_id"
+    t.index ["user_id"], name: "index_messages_on_user_id"
+  end
```
zachgoll, line 484: Looks like we're still getting the extra tables here.  You may need a full `rails db:migrate:reset` since `db:reset` will just rebuild based on this existing `schema.rb` and not perform the migrations

app/controllers/plaid_items_controller.rb
```diff
@@ -5,6 +5,7 @@ def create
     Current.family.plaid_items.create_from_public_token(
       plaid_item_params[:public_token],
       item_name: item_name,
+      region: plaid_item_params[:region]
```
zachgoll, line 8: The `create_from_public_token` method on `PlaidItem` will need to be updated to accept this param

app/models/plaid_item.rb
```diff
@@ -1,6 +1,9 @@
 class PlaidItem < ApplicationRecord
   include Plaidable, Syncable
 
+  enum :plaid_region, { us: "us", eu: "eu" }
+  validates :plaid_region, inclusion: { in: plaid_regions.keys }
```
zachgoll, line 5: The `enum` should have this `inclusion` validation already built-in, so we can remove the second `validates` here

zachgoll (commented):
Yeah, I like this strategy of passing it from the Stimulus controller.  Should work nicely!

Shpigford:
@zachgoll Two things I'm not sure on...

1) Do we need to add Canada ("CA") to the region stuff? Or will "US" properly cover it thanks to `get_country_codes`?
2) There are failing system tests that I'm not familiar enough with to not potentially completely break them.

zachgoll (approved):
@Shpigford to my understanding, you have the `CA` configuration correct here.  I believe `CA` is fully covered under the "US" client:

CleanShot 2025-01-31 at 12 19 42

app/views/accounts/new/_method_selector.html.erb
```diff
@@ -6,16 +6,27 @@
       <span class="flex w-8 h-8 shrink-0 grow-0 items-center justify-center rounded-lg bg-alpha-black-50 shadow-[inset_0_0_0_1px_rgba(0,0,0,0.02)]">
         <%= lucide_icon("keyboard", class: "text-gray-500 w-5 h-5") %>
       </span>
-      <%= t("accounts.new.method_selector.manual_entry") %>
+      <%= t(".manual_entry") %>
```
zachgoll, line 9: Since this partial is shared across all accountables (delegated types), we need to keep the path absolute:

`accounts.new.method_selector.manual_entry`

That should fix the e2e tests

Shpigford:
@zachgoll ✅ Anything else before merging this?

zachgoll:
@Shpigford I think we're all set.  Only thing I can think of is making sure these are set in prod:

PLAID_EU_CLIENT_ID=
PLAID_EU_SECRET=

henrikbjorn:
Not to hijack, but would GoCardless be an idea also? That is easily obtainable for self-hosting.

Shpigford:
> Not to hijack, but would GoCardless be an idea also? That is easily obtainable for self-hosting.

https://github.com/maybe-finance/maybe/discussions/141

PR #1556, opened 2024-12-18
Preserve original transaction names when enriching

zachgoll:
A couple of data quality updates in this PR:

- Make `name` required on `Account::Entry` - since users are able to directly edit this value, enforcing non-nullability makes sense and prevents data quality errors that have come up recently
- When enriching transactions, store `enriched_name` rather than overwriting `name`.  This will allow us to preserve "rollback" abilities in the future if a user wants to revert (or just see) the original txn name before enrichment was applied (increase user confidence in enabling the enrichment feature)
- All transactions coming from Plaid accounts will be automatically enriched, regardless of user preference as part of the overall "transformation" of Plaid data to Maybe data.  Manual transactions will not be enriched unless explicitly enabled by the user.

### DIFF
--- a/db/migrate/20241218132503_add_enriched_name_field.rb
+++ b/db/migrate/20241218132503_add_enriched_name_field.rb
@@ -1,0 +1,37 @@
+class AddEnrichedNameField < ActiveRecord::Migration[7.2]
+  def change
+    add_column :account_entries, :enriched_name, :string
+
+    reversible do |dir|
+      dir.up do
+        execute <<-SQL
+          UPDATE account_entries ae
+          SET name = CASE ae.entryable_type
+            WHEN 'Account::Trade' THEN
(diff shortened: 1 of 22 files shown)


db/migrate/20241218132503_add_enriched_name_field.rb
```diff
@@ -0,0 +1,37 @@
...
+              END
+            WHEN 'Account::Transaction' THEN
+              CASE
+                WHEN ae.amount > 0 THEN 'Expense'
+                ELSE 'Income'
+              END
+            WHEN 'Account::Valuation' THEN 'Balance update'
+            ELSE 'Unknown entry'
+          END
+          WHERE name IS NULL
```
zachgoll, line 26: only updating "bad data" that currently exists

PR #1585, opened 2024-12-31
Transfer and Payment auto-matching, model and UI improvements

zachgoll:


nikhilbadyal:
@zachgoll  Auto Matching behaving weirdly. Auto Match should atleast try to match on same dates. Auto Matching with past dates doesn't sound strange.

Ex - 
image


From date - 7th 
To Date- 5th

zachgoll:
@nikhilbadyal I'm not sure I understand what you're referring to here.  Can you explain this a bit more?  Is the screenshot shown a valid transfer for you?

Currently, the matching window "tolerance" is 4 days to account for cases where the outgoing transfer was sent on a Friday, there was a 3 day holiday weekend, and the incoming settled the following Tuesday.

nikhilbadyal:
I got the tolerance but **Receiver transaction** date will always be greater (or euqal) than **Sender Transaction** date right ?

In above screenshot. 

Money was sent Transferred from ICICI account on 7th Jan 2025 to SBI account on 3Jan 2025 . This look like a transfer from Future to past.

zachgoll:
@nikhilbadyal yep, makes sense!  I agree that the _receiver_ should always be _equal or later date_ to the sender.

All that said, you should have the option to "Reject" the transfer in the UI, which was put in for incorrect detections just like this.  Are you seeing that and is it working?

nikhilbadyal:
Yes,  removed the transfer by that button. Found a few more such instances as well. I'll probably find all and revert manually or fire a query to revert instead of checking one by one.
Thanks for this cool feature. It's awesome 💯

zachgoll:
@nikhilbadyal nice, yeah you'll have to do a manual rejection of these, but I've opened #1602 so future auto-detects aren't wrong like this

PR #1611, opened 2025-01-09
fix: Only admins can generate invite codes

### LINKED ISSUE #1606: Bug: Self hosted invite codes should only be generated by admins
Hello 👋 

**Describe the bug**
I found a security issue. All users can create invite codes.

**To Reproduce**
1. Login or Signup
2. Activate `Require invite code for new sign ups`
3. Generate a code
4. Create a new account with the code generated in the previous step
5. With this new account got to `Settings > Self Hosted`

**Expected behavior**
I think only "admin" users could generate codes.

Self Hosted, v0.2.0

The problem is on all OS and browsers

---

An idea to fix this problem would be to set an "admin email" in env vars. Only the user logged in with this email could do some actions like generate invite code.
If this solution sounds good to you, I can implement it.

tonyvince:
close #1606

### DIFF
--- a/app/controllers/invite_codes_controller.rb
+++ b/app/controllers/invite_codes_controller.rb
@@ -1,12 +1,12 @@
 class InviteCodesController < ApplicationController
   before_action :ensure_self_hosted
-  before_action :ensure_admin, only: :create
 
   def index
     @invite_codes = InviteCode.all
   end
 
   def create
+    raise StandardError, "You are not allowed to generate invite codes" unless Current.user.admin?
     InviteCode.generate!
     redirect_back_or_to invite_codes_path, notice: "Code generated"
   end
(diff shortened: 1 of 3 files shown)


hecht-a:
Hey, 
For more privacy and security I think invite codes should be hidden from non admin users

app/controllers/invite_codes_controller.rb
```diff
@@ -15,4 +16,8 @@ def create
     def ensure_self_hosted
       redirect_to root_path unless self_hosted?
     end
+
+    def ensure_admin
+      redirect_to settings_hosting_path, alert: "You are not allowed to generate invite codes" unless Current.user.admin?
+    end
```
zachgoll, line 22: I think since we're disabling the input in the UI, we should probably just throw an exception here.  A user should not be able to trigger this POST request given the UI, so if they try to side-step that, they should receive an error I think.

zachgoll (commented):
Thanks @tonyvince!  Looks good, just a minor suggestion.

zachgoll (approved):
Looks good!

PR #1612, opened 2025-01-09
Add loan start date

tonyvince:
close #1605


zachgoll:
@tonyvince I think to solve this one, we'll need to take a slightly different approach.  Adding a `start_date` to an Account conflicts with our existing model that relies solely on `Account::Valuation` and `Account::Transaction` entries to calculate the historical balance of an account.

A possible approach here would be to include a `origination_date` and `origination_loan_balance` field in the initial form the user fills out and then use that information to create an `Account::Valuation` representing the "starting loan amount".

Currently, for all accounts we're calling `create_and_sync` which generates 2 valuations by default:

https://github.com/maybe-finance/maybe/blob/413ec6cbed72bcf22ea97dce6a434d4202c0facc/app/models/account.rb#L58-L85

For loan accounts, we likely want to "specialize" this behavior so that instead of blindly creating a valuation of 0 a day ago, we create a valuation on the `origination_date` with an amount equal to `origination_loan_balance`.

Skypay88:
README.
Loan solutions for life's unexpected twists! Reach out to me to discuss options 💸

PR #1622, opened 2025-01-15
FIX: correct display of percentages

### LINKED ISSUE #1619: Bug: Savings Rate calculation is wrong
**Describe the bug**
The calculation of the savings rate is incorrect by one decimal place

**To Reproduce**
Add 2 transactions:
1: Income with 1000€
2: Expense with 500€

**Expected behavior**
` (income - spending) / income`
`(1000 - 500) / 1000`
The result is 0.5, which is basically correct, but it still needs to be multiplied by 100 to be a percentage

Self-hosted  v0.2.0

**Screenshots / Recordings**
chrome_IafEN1MWrE

MrTob:
Fixes #1619

### DIFF
--- a/app/views/shared/_value_heading.html.erb
+++ b/app/views/shared/_value_heading.html.erb
@@ -5,7 +5,7 @@
     <% if value.is_a?(Money) %>
       <span class="<%= size == "lg" ? "text-xl" : "text-lg" %> font-medium"><%= format_money value %></span>
     <% else %>
-      <span class="<%= size == "lg" ? "text-xl" : "text-lg" %> font-medium"><%= is_percentage ? number_to_percentage(value, precision: 2) : value %></span>
+      <span class="<%= size == "lg" ? "text-xl" : "text-lg" %> font-medium"><%= is_percentage ? number_to_percentage(value*100, precision: 2) : value %></span>
     <% end %>
   </p>
   <% if trend.nil? %>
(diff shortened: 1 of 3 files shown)


app/views/shared/_value_heading.html.erb
```diff
@@ -5,7 +5,7 @@
     <% if value.is_a?(Money) %>
       <span class="<%= size == "lg" ? "text-xl" : "text-lg" %> font-medium"><%= format_money value %></span>
     <% else %>
-      <span class="<%= size == "lg" ? "text-xl" : "text-lg" %> font-medium"><%= is_percentage ? number_to_percentage(value, precision: 2) : value %></span>
+      <span class="<%= size == "lg" ? "text-xl" : "text-lg" %> font-medium"><%= is_percentage ? number_to_percentage(value*100, precision: 2) : value %></span>
```
zachgoll, line 8: @MrTob I don't think this is the right spot to change this.  This partial is used across _all_ metrics, so while it will fix the "savings rate" metric, it will break the other metrics.

Where I think we need to update this is directly in the query to make sure that we're multiplying the percent fraction *100:

https://github.com/maybe-finance/maybe/blob/195ec85d96bd578098c8dc9ea084de4db28be03b/app/models/family.rb#L158

zachgoll (changes requested):

MrTob:
@zachgoll 
I have changed the code, the calculation is always the same (0.xx). Only in the view the number is multiplied by 100.

Issue #1631, opened 2025-01-19
Bug: credit card activity page change in Rows per page leads to Content missing screen


developerdsk:
**Describe the bug**
when i try to change the number of rows for Credit card account , it shows Content missing instead of loading the number of rows selected

![Image](https://github.com/user-attachments/assets/d039bedd-b30e-472b-ad9e-94c276a90148)

**To Reproduce**
1. Go to Credit Card
2. Click on any credit card account
3. Click on Activity
4. Scroll down to change the Rows per page from 10 to 20
5. See error  Content missing
**Expected behavior**
should be able to load the number of rows selected 

**What version of Maybe are you using?**
This could be "Hosted" (i.e. app.maybefinance.com) or "Self-hosted". If "Self-hosted", please include the version you're currently on.
self hosted
docker latest 

**What operating system and browser are you using?**
Windows , firefox

one more observation . if i change the Rows per page in Depository account it shows   Loading entries... but never loads anything 

![Image](https://github.com/user-attachments/assets/f3d00ca9-1699-462b-8363-a44c38d3d5ee)

PR #1633, opened 2025-01-19
FIX: 1631

MrTob:
# Fix rows per page selection in account entries (#1631 )

The "rows per page" selector in account entries didn't work correctly because the form was submitting to the wrong endpoint. This was happening because it used `custom_pagy_url_for` which directed the request through the parent resource's show action instead of the entries endpoint.

## Changes
- Modified the form in `_pagination.html.erb` to submit directly to `account_entries_path` with the correct account_id
- Added turbo_frame targeting to ensure proper frame updates
- Added explicit parameter preservation for tab and search context
- Removed `custom_pagy_url_for` usage from the form (kept for pagination links)

The form now correctly submits to the entries endpoint while preserving all necessary context parameters, allowing proper rows per page updates within the Turbo Frame.


zachgoll (commented):
@MrTob this is definitely the cause of the bug, but I think we should be careful adding to much responsibility to this `_pagination.html.erb` partial.  This partial is meant to be agnostic of what view it is rendered in, so I don't think we should be adding checks such as `defined?(@account) ? account_entries_path(...) : ...`

For the moment, I think a clearer option would be to duplicate this partial into an entry-specific one, such as `views/account/entries/_pagination.html.erb`

That way, we can be highly specific with our intended routes and remove a lot of the conditional logic.

Yes, this will result in a bit of code duplication, but I think the clarity tradeoff is worth it at this stage.

zachgoll:
@MrTob be sure to request a review when this is ready again and I can take a final look!

Issue #1645, opened 2025-01-20
Bug: Loan Summary contains zero-values after first load


dmnk:
**Describe the bug**
Immediately after creating a loan, the fields "original principal" and "monthly payment" are filled with values. After navigating to another page, and returning, they stay at 0,00 €

**To Reproduce**
1. create new loan /loans/new?step=method_select
2. select "enter account balance"
3. fill the fields
4. click create accounts
5. see that rate and original principal 
6. open **activity** tab of the loan
7. switch back to the **Overview** of the loan
8. (Updating the loan doesn't re-trigger the calculation as well)

**Expected behavior**
Have the two fields updated and calculated every time the tab is opened

**What version of Maybe are you using?**
This could be "Hosted" (i.e. app.maybefinance.com) or "Self-hosted". If "Self-hosted", please include the version you're currently on.
0.3, but happened with 0.2 as well. Self-Hosted.

**What operating system and browser are you using?**
No difference between OS and Browsers (all relevant tried)

**Screenshots / Recordings**
If applicable, add screenshots or short video recordings to help show the bug in more detail.

![Image](https://github.com/user-attachments/assets/cc0cbf80-5e13-4ad8-b81c-3f4bc4f55dd2)

**Additional context**

iursevla:
Same here.

 @dmnk  did you have any place to set the original principal? 

<img width="2086" alt="Image" src="https://github.com/user-attachments/assets/efbfc0b6-2576-468a-b6f6-b889922d7be2" />

PR #1650, opened 2025-01-21
FIX: loan summary zero values after navigation

MrTob:
### Issue (#1645)
After creating a loan, the "original principal" and "monthly payment" fields display correct values initially. However, after navigating away and returning to the overview tab, these values incorrectly show as 0.00.

### Root Cause
The `original_balance` method in the `Account` model was prioritizing the first chronological balance record even when it had a zero value, instead of falling back to the current balance. This caused the loan calculations to use 0 as the base value after the initial page load.

### Solution
Modified the `original_balance` method in the Account model to handle zero-value balance records more appropriately.
The change ensures that:
- If the first balance record exists but is zero, use the current balance
- If the first balance record has a non-zero value, use that
- If no balance record exists, fall back to current balance


zachgoll (commented):
Thanks for tackling this!  I think your solution here would fix the summary card, but it will create inconsistencies with the historical graph. 

In this case, a user will see a "start balance" of say, "$50,000", but then on the graph, the first value will still show as $0 leading to potential confusion. 

I outlined a potential solution here that would make the initial values more appropriate for a loan-type account: https://github.com/maybe-finance/maybe/pull/1612#issuecomment-2587768657

r4f4siqueira (approved):

PR #1658, opened 2025-01-23
Add cabin / cottage as a property type

eithe:
Hope you would consider adding an option for cabins :) Open for suggestions on naming.

### DIFF
--- a/app/models/property.rb
+++ b/app/models/property.rb
@@ -6,7 +6,8 @@ class Property < ApplicationRecord
     [ "Multi-Family Home", "multi_family_home" ],
     [ "Condominium", "condominium" ],
     [ "Townhouse", "townhouse" ],
-    [ "Investment Property", "investment_property" ]
+    [ "Investment Property", "investment_property" ],
+    [ "Cabin / Cottage", "cabin" ]
   ]
 
   has_one :address, as: :addressable, dependent: :destroy
--- a/app/models/property.rb
+++ b/app/models/property.rb
@@ -7,7 +7,7 @@ class Property < ApplicationRecord
     [ "Condominium", "condominium" ],
     [ "Townhouse", "townhouse" ],
     [ "Investment Property", "investment_property" ],
-    [ "Cabin / Cottage", "cabin" ]
+    [ "Second Home", "second_home" ]
   ]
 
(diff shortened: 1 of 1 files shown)


Shpigford:
@eithe Thanks for the suggestion! How does this different from "Single Family Home"? The purpose of the types isn't so much to list every possible structure type, but rather the *purpose*.

eithe:
Thanks for the response! I understand where you’re coming from.

For me, the existing 'Single Family Home' doesn’t feel like the right fit for cabins, since I associate 'Home' more with a primary residence. Cabins, on the other hand, are often seen as vacation retreats, seasonal properties, or cottages, which don’t quite align with the concept of a 'home.' I think having 'Cabin' as an option could better reflect this distinction, and especially that they have a different _purpose_; it's not about the structure type.

Having said that this was just something I thought of when I looked at the list of properties in the UI and both my house and cabin was categorized as the same thing, which just looked a bit weird to me since I consider them different entities :)

I also want to add that I really appreciate how focused and streamlined Maybe is in general — it's a strength to avoid having too many unnecessary options. I completely understand the need to strike a balance between flexibility and simplicity!

zachgoll:
@eithe would `second_home` or `vacation_property` work here?

"Cabin" feels too specific to introduce, but I'd agree that we're missing a category for these "second home" type properties, which I'd assume is common enough for our users to consider.

eithe:
> @eithe would `second_home` or `vacation_property` work here?
Yes.

zachgoll:
Open for votes, but "Second home" (`second_home`) feels like it is the most general here and probably our best option.

app/models/property.rb
```diff
@@ -6,7 +6,8 @@ class Property < ApplicationRecord
     [ "Multi-Family Home", "multi_family_home" ],
     [ "Condominium", "condominium" ],
     [ "Townhouse", "townhouse" ],
-    [ "Investment Property", "investment_property" ]
+    [ "Investment Property", "investment_property" ],
+    [ "Cabin / Cottage", "cabin" ]
```
zachgoll, line 10:
```suggestion
    [ "Second Home", "second_home" ]
```

Issue #1672, opened 2025-01-23
Bug: Credit card account minimum payment does not accept decimal values


zachgoll:
![Image](https://github.com/user-attachments/assets/9869f61f-588d-4960-93e1-44ed00028d91)

PR #1681, opened 2025-01-24
Added Decimal Support in min transaction

Harry-kp:
fixes: #1672

### DIFF
--- a/app/views/credit_cards/_form.html.erb
+++ b/app/views/credit_cards/_form.html.erb
@@ -16,7 +16,8 @@
         <%= credit_card_form.number_field :minimum_payment,
                                         label: t("credit_cards.form.minimum_payment"),
                                         placeholder: t("credit_cards.form.minimum_payment_placeholder"),
-                                        min: 0 %>
+                                        min: 0,
+                                        step: 0.01 %>
         <%= credit_card_form.number_field :apr,
                                         label: t("credit_cards.form.apr"),
                                         placeholder: t("credit_cards.form.apr_placeholder"),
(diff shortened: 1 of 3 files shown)


zachgoll (commented):
Thanks for the fix!  I think we may want to use our built in `money_field` here so that we can take advantage of the built-in precision handling that it offers.  Here's an example of the usage:

https://github.com/maybe-finance/maybe/blob/7e0ec4bd8fad6642115b4f413cca0ed19754dfb9/app/views/accounts/_form.html.erb#L9

This is made available via our custom form builder:

https://github.com/maybe-finance/maybe/blob/7e0ec4bd8fad6642115b4f413cca0ed19754dfb9/app/helpers/styled_form_builder.rb#L42

Harry-kp:
> Thanks for the fix! I think we may want to use our built in `money_field` here so that we can take advantage of the built-in precision handling that it offers. Here's an example of the usage:
> 
> https://github.com/maybe-finance/maybe/blob/7e0ec4bd8fad6642115b4f413cca0ed19754dfb9/app/views/accounts/_form.html.erb#L9
> 
> This is made available via our custom form builder:
> 
> https://github.com/maybe-finance/maybe/blob/7e0ec4bd8fad6642115b4f413cca0ed19754dfb9/app/helpers/styled_form_builder.rb#L42

@zachgoll Updated the PR to incorporate the suggested changes. Also updated the test to include the decimal scenario

zachgoll (approved):
Awesome, looks great now!  Thanks for the fix.

Issue #1701, opened 2025-01-26
Dashboard: Add more periods or be able to customize it


javiermartingonzalez:
Hello, I want to be able to have more 'period' options when showing the chart. I think 7D, 1M, 1Y is not enough and other options like 3Y or even YTD for Investment will be a good point to check the performance.

I want to suggest add more periods, or be able to customize in Settings the desired options.

Thanks for your attention 

<img width="458" alt="Image" src="https://github.com/user-attachments/assets/6f276049-1d56-4b06-8442-e1bfd0ab3c4c" />

zachgoll:
Would be happy to accept a few more periods here.  The only period that we're not able to at the moment is `all` because we have not yet optimized the queries to handle long histories of balances and that period causes performance issues.  Long term, we'll accept an `all` period, but need to address a few performance bits before introducing.

I think a good set of periods would be:

- `7D` - previous 7 days
- `MTD` - month to date
- `1M` - previous 30 days
- `YTD` - year to date
- `1Y` - previous 365 days
- `5Y` - for most accounts, gives "all" of the history, but provides an important "cap" on the amount of history to avoid major performance issues.

PR #1714, opened 2025-01-27
Add More Timelines

nikhilbadyal:
The current setup has a significant gap in its timeline options, jumping from 1 week to 1 month and then directly to 1 year. Introducing a 90-day (quarterly) period would bridge this gap effectively. A 90-day timeline is crucial, as it aligns with how people typically evaluate goals and progress on a quarterly basis. This addition would provide a more comprehensive range of timeframes for users to review and plan.

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -8,6 +8,18 @@ module AccountsHelper
 
     days_apart = (end_date - start_date).to_i
 
+    # Handle specific cases
+    if start_date == Date.current.beginning_of_week && end_date == Date.current
+      return "Current Week to Date (CWD)"
+    elsif start_date == Date.current.beginning_of_month && end_date == Date.current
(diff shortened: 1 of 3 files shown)


zachgoll:
@nikhilbadyal agreed!  While we're here, would you mind also adding the periods I had mentioned in this comment?  I think that would get us to a pretty comprehensive set of options:

https://github.com/maybe-finance/maybe/issues/1701#issuecomment-2615954657

nikhilbadyal:
@zachgoll You can review now.

zachgoll (approved):
Looks good!  We may have to tweak based on user feedback, but I think this is a good base to go off for now.

nikhilbadyal:
@zachgoll , i was thinking maybe there should be a support to set default Timeline which can be set using `settings/preferences`. Like i personally like MTD instead of 1M, Some might like YTD or WTD.

PR #1722, opened 2025-01-28
Ensure Consistent Category Colors

### LINKED ISSUE #1668: Improvement: Add validation so that subcategories are always the same color as the parent
Currently, we allow users to have different colors for parent/subcategories.  This creates a visually confusing UI; especially in the budgets section.

Image

Image

**Solution**

- Add validation to ensure subcategory always gets assigned color of parent
- Update category form to reflect this validation

JLambertazzo:
Closes #1668

> [!NOTE]
> Was not able to accomplish reactivity in this PR. Disabling of other colors, in addition to displaying the error message, only shows up after a failed attempt to create the category.

## Approach
* Added validation in the `category.rb` model which displays the message below when trying to select a color different from the parent category
* Grayed-out and disabled other colors when validation fails to reflect validation rule
  Validation result. Invalid colors are greyed out and disabled. Error message is displayed in red
* Once correct color is selected, category can be created
  Successful result after selecting a valid color


zachgoll:
@JLambertazzo just an idea here—what if we used a simple Stimulus controller to hide the color picker entirely if the parent category field is populated, then on the backend, we simply use our internal logic to assign the parent color to the subcategory?

cc @justinfar in case you had other ideas here.

JLambertazzo:
> @JLambertazzo just an idea here—what if we used a simple Stimulus controller to hide the color picker entirely if the parent category field is populated, then on the backend, we simply use our internal logic to assign the parent color to the subcategory?

This worked really well! Here are the results:
Selection hides when parent is selected


Created subcategory is forced to have parent color

zachgoll (approved):
Looks good!  Once tests are passing it's good to merge.

JLambertazzo:
Completed merge, behavior seems to be the same!

Issue #1712, opened 2025-01-27
Bug: Incorrect date format option for CSV imports


zachgoll:
![Image](https://github.com/user-attachments/assets/7080ec1b-df05-41d2-b850-0b5c3381aae8)

We need to change `DD.MM.YY` -> `DD.MM.YYYY` to be consistent with the other formats.

quangsinhdev:
Ok

PR #1726, opened 2025-01-28
Fix: make date format year consistent overall #1712

scodes73:
Fixes simple year formatting, making it consistent everywhere
Fixes : https://github.com/maybe-finance/maybe/issues/1712
Before:
image
After:
Unsaved Image 1

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,20 +1,6 @@
 module ApplicationHelper
   include Pagy::Frontend
 
-  def date_format_options
-    [
-      [ "DD-MM-YYYY", "%d-%m-%Y" ],
-      [ "DD.MM.YYYY", "%d.%m.%Y" ],
-      [ "MM-DD-YYYY", "%m-%d-%Y" ],
-      [ "YYYY-MM-DD", "%Y-%m-%d" ],
-      [ "DD/MM/YYYY", "%d/%m/%Y" ],
-      [ "YYYY/MM/DD", "%Y/%m/%d" ],
-      [ "MM/DD/YYYY", "%m/%d/%Y" ],
-      [ "D/MM/YYYY", "%e/%m/%Y" ],
-      [ "YYYY.MM.DD", "%Y.%m.%d" ]
-    ]
-  end
-
   def icon(key, size: "md", color: "current")
     render partial: "shared/icon", locals: { key:, size:, color: }
   end
--- a/app/models/family.rb
+++ b/app/models/family.rb
@@ -1,7 +1,17 @@
 class Family < ApplicationRecord
   include Plaidable, Syncable
 
-  DATE_FORMATS = [ "%m-%d-%Y", "%d.%m.%Y", "%d-%m-%Y", "%Y-%m-%d", "%d/%m/%Y", "%Y/%m/%d", "%m/%d/%Y", "%e/%m/%Y", "%Y.%m.%d" ]
+  DATE_FORMATS = [
+    [ "MM-DD-YYYY", "%m-%d-%Y" ],
+    [ "DD.MM.YYYY", "%d.%m.%Y" ],
+    [ "DD-MM-YYYY", "%d-%m-%Y" ],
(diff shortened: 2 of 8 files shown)


zachgoll (changes requested):
Thanks for tackling this!

There are a few more spots in the codebase where these date formats are used.  I think this may be a good time to consolidate some of these and centralize so that changes like these can be quicker (i.e. referencing a helper method for all of the import forms that use these date formats)

CleanShot 2025-01-28 at 11 11 01

scodes73:
> There are a few more spots in the codebase where these date formats are used. I think this may be a good time to consolidate some of these and centralize so that changes like these can be quicker (i.e. referencing a helper method for all of the import forms that use these date formats)
> 
Using `date_format_options` from app/helps/application_helper.rb for the date formats.
Made changes in:

1. `app/views/import/configurations/_mint_import.html.erb`
2. `app/views/import/configurations/_trade_import.html.erb`
3. `app/views/import/configurations/_transaction_import.html.erb`

It's adding more formats [ "D/MM/YYYY", "%e/%m/%Y" ], [ "YYYY.MM.DD", "%Y.%m.%d" ], Is it fine having these 2 as well in these?

scodes73:
Also, Should we have required both in HTML attributes and also in the attributes section as well.
`<%= form.select :date_format, date_format_options, { label: t(".date_format"), required: true}, label: true, required: true, disabled: import.complete? %>` Here we're having required in both sections, is it required? I'm not sure if the required parameter is checked else where manually through code

scodes73:
> @scodes73 looking better! The `required` attribute only needs to be written once per input. It should be in the attributes section, not the select options.
> 
> I think the last step here is to consolidate that list of formats that we have in `family.rb` with the list in `application_helper`. Ideally, I think we actually remove the helper from `application_helper` entirely and store the canonical list of date formats directly in the Family model:
> 
> ```ruby
> class Family < ApplicationRecord
>   DATE_FORMATS = [
>     [ "MM-DD-YYYY", "%m-%d-%Y" ],
>     [ "DD.MM.YYYY", "%d.%m.%Y" ],
>     [ "DD-MM-YYYY", "%d-%m-%Y" ],
>     [ "YYYY-MM-DD", "%Y-%m-%d" ],
>     [ "DD/MM/YYYY", "%d/%m/%Y" ],
>     [ "YYYY/MM/DD", "%Y/%m/%d" ],
>     [ "MM/DD/YYYY", "%m/%d/%Y" ],
>     [ "D/MM/YYYY", "%e/%m/%Y" ],
>     [ "YYYY.MM.DD", "%Y.%m.%d" ]
>   ].freeze
> end
> ```
> 
> Then, in our forms, we can use `Family::DATE_FORMATS` as options and have confidence that these are only defined in one spot across the entire app

Done, let me know of any more changes. Thank you for guiding me thoroughly :)

zachgoll (approved):
Looks good!  Thanks for tackling this!

zachgoll:
@scodes73 looks like we have a few tests failing for `i18n` translations.  You can run the following commands to see which ones are unused/missing:

```
bundle exec i18n-tasks unused -l en
bundle exec i18n-tasks missing -l en
```

scodes73:
> @scodes73 looks like we have a few tests failing for `i18n` translations. You can run the following commands to see which ones are unused/missing:
> 
> ```
> bundle exec i18n-tasks unused -l en
> bundle exec i18n-tasks missing -l en
> ```

I've made new changes:
image

But it marks the variable as unused, I could always change it to individual labels if you want me to

image

scodes73:
> > @scodes73 looks like we have a few tests failing for `i18n` translations. You can run the following commands to see which ones are unused/missing:
> > ```
> > bundle exec i18n-tasks unused -l en
> > bundle exec i18n-tasks missing -l en
> > ```
> 
> I've made new changes: image
> 
> But it marks the variable as unused, I could always change it to individual labels if you want me to
> 
> image

Now reverting to previous state of individual translations

zachgoll (approved):
Nice work!

Issue #1677, opened 2025-01-24
Bug: subcategories of subcategory is not shown


mamarguerat:
**Describe the bug**
In a category has a subcategory, which have sub-subcategories, the sub-subcategories are not shown in the category edition page and in the budget page.

**To Reproduce**
1. Create 3 categories, A, B and C
2. Choose category B as parent of category C
3. Choose category A as parent of category B
4. The category A is shown, the category B is shown as a subcategory of A, and category C is not shown anywhere.

**Expected behavior**
I expect either that the category is shown as a sub-subcategory, or that I cannot set a parent for a category that already has subcategories.

**What version of Maybe are you using?**
self-hosted v0.3.0 - 0476f25952f568492c856f620d3352794295233a

**What operating system and browser are you using?**
Windows 11, Firefox

zachgoll:
@mamarguerat we only allow 1 level of nesting for categories in the app.  The app should have prevented and thrown an error if you tried to create a "sub-subcategory".  Could you share a screenshot of what you're seeing?

mamarguerat:
https://github.com/user-attachments/assets/25888d37-7525-4c3a-b91f-3402192d6f73

In the screenshot, I set *Clothing and accessories* as subcategory of *Charitable donations* and then *Charitable donations* as subcategory of *Bank fees*. The *Clothing and accessories* is not shown in any view.

zachgoll:
@mamarguerat thanks for the details!

> But the app doesn't prevent setting a category, which already has subcategory, as a subcategory.

Agree, this is definitely a bug.  We should remove all categories that already have a subcategory from that dropdown list.

> A side note, there is this 1password pop-up for the category name. I think that the field has an auto-field label, and I don't know if it is wanted.

Yeah I noticed that too recently.  Probably worth removing the auto field label.

PR #1730, opened 2025-01-28
Fix/1677

elvisserrao:
Fixes issue #1677

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -2,6 +2,7 @@ class CategoriesController < ApplicationController
   layout :with_sidebar
 
   before_action :set_category, only: %i[edit update destroy]
+  before_action :set_categories, only: %i[update edit]
   before_action :set_transaction, only: :create
 
   def index
@@ -10,7 +11,7 @@ class CategoriesController < ApplicationController
 
   def new
     @category = Current.family.categories.new color: Category::COLORS.sample
-    @categories = Current.family.categories.alphabetically.where(parent_id: nil).where.not(id: @category.id)
+    set_categories
   end
 
   def create
@@ -27,19 +28,21 @@ class CategoriesController < ApplicationController
         format.turbo_stream { render turbo_stream: turbo_stream.action(:redirect, redirect_target_url) }
       end
     else
-      @categories = Current.family.categories.alphabetically.where(parent_id: nil).where.not(id: @category.id)
+      set_categories
       render :new, status: :unprocessable_entity
     end
   end
 
   def edit
-    @categories = Current.family.categories.alphabetically.where(parent_id: nil).where.not(id: @category.id)
   end
 
   def update
-    @category.update! category_params
+    if @category.update(category_params)
 
-    redirect_back_or_to categories_path, notice: t(".success")
+      redirect_back_or_to categories_path, notice: t(".success")
+    else
+      render :edit, status: :unprocessable_entity
+    end
   end
 
   def destroy
@@ -57,6 +60,14 @@ class CategoriesController < ApplicationController
   private
     def set_category
       @category = Current.family.categories.find(params[:id])
+    end
+
+    def set_categories
+      @categories = unless @category.parent?
+        Current.family.categories.alphabetically.roots.where.not(id: @category.id)
+      else
+        []
+      end
     end
 
     def set_transaction
--- a/app/views/categories/_form.html.erb
+++ b/app/views/categories/_form.html.erb
@@ -34,7 +34,7 @@
       <div class="space-y-2">
         <%= f.select :classification, [["Income", "income"], ["Expense", "expense"]], { label: "Classification" }, required: true %>
         <%= f.text_field :name, placeholder: t(".placeholder"), required: true, autofocus: true, label: "Name", data: { color_avatar_target: "name" } %>
-        <%= f.select :parent_id, categories.pluck(:name, :id), { include_blank: "(unassigned)", label: "Parent category (optional)" } %>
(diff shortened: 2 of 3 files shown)


app/views/categories/_form.html.erb
```diff
@@ -34,7 +34,10 @@
       <div class="space-y-2">
         <%= f.select :classification, [["Income", "income"], ["Expense", "expense"]], { label: "Classification" }, required: true %>
         <%= f.text_field :name, placeholder: t(".placeholder"), required: true, autofocus: true, label: "Name", data: { color_avatar_target: "name" } %>
-        <%= f.select :parent_id, categories.pluck(:name, :id), { include_blank: "(unassigned)", label: "Parent category (optional)" } %>
+        <%= f.select :parent_id, categories.pluck(:name, :id), { include_blank: "(unassigned)", label: "Parent category (optional)" }, disabled: category.parent? %>
+        <% if category.parent? %>
+          <span class="text-xs italic pl-2 text-gray-500">This category couldn’t be assigned as a subcategory because it already has subcategories.</span>
+        <% end %>
```
zachgoll, line 40: I think here, we can just have a conditional:

```erb
<% unless category.parent? %>
  <%= f.select :parent_id ...
<% end %>
```

That way, if the user is not allowed to assign a parent, we don't ever give them the option.

app/controllers/categories_controller.rb
```diff
@@ -27,19 +28,21 @@ def create
...
   end
 
   def update
-    @category.update! category_params
+    if @category.update(category_params)
 
-    redirect_back_or_to categories_path, notice: t(".success")
+      redirect_back_or_to categories_path, notice: t(".success")
+    else
+      render :edit, status: :unprocessable_entity
```
zachgoll, line 44: IIRC, we're submitting the category form to the `:_top` turbo frame, so I believe this `:unprocessable_entity` will cause the background of the modal to go blank on an error.

I think we either need to keep the `update!` or remove that `:_top` target that was set here so that both success/failure cases render correctly:

https://github.com/maybe-finance/maybe/commit/8256d116ddd3b892d2a4256abdad86d1da25fd95

zachgoll (commented):
Overall, agree with the strategy here!

app/controllers/categories_controller.rb
elvisserrao: The problem with keeping the ```update!``` is that the user never receives any feedback, which is why I changed it.
In the create action, we’re facing the same issue, causing the background to go blank when there is a validation error. That’s why I chose to ignore this issue for now.

app/views/categories/_form.html.erb
elvisserrao: Nice, i agree.

app/controllers/categories_controller.rb
zachgoll: Ahh you're right, looks like we had a regression in https://github.com/maybe-finance/maybe/commit/8256d116ddd3b892d2a4256abdad86d1da25fd95 

You think we could throw in the correct handling for this endpoint (and _remove_ the `:_top` target introduced in that commit)?  Here's an example of how we're doing it for other endpoints:

```rb
    if @entry.save
      flash[:notice] = t(".success")

      respond_to do |format|
        format.html { redirect_back_or_to account_path(@entry.account) }

        redirect_target_url = request.referer || account_path(@entry.account)
        format.turbo_stream { render turbo_stream: turbo_stream.action(:redirect, redirect_target_url) }
      end
    else
      render :new, status: :unprocessable_entity
    end
```

We're essentially using a custom Turbo action to allow us to handle error states directly with streams, but do a full page refresh on success.  It helps avoid that "blank screen" issue:

https://github.com/maybe-finance/maybe/blob/247d91b99de256465001590b9b77c91df1942f9c/app/javascript/application.js#L5-L7

elvisserrao: I made some changes on [4087275](https://github.com/maybe-finance/maybe/pull/1730/commits/408727575142cb4f3620f6f665e2a6a19374e585) to handle the turbo_stream form requests. I removed the :_top turbo target and added a turbo_frame_tag to be replaced by the form partial and render the errors.

app/controllers/categories_controller.rb
```diff
@@ -23,23 +24,31 @@ def create
...
 
-    redirect_back_or_to categories_path, notice: t(".success")
+      respond_to do |format|
+        format.turbo_stream { render turbo_stream: turbo_stream.action(:redirect, categories_path) }
+      end
+    else
+      respond_to do |format|
+        format.turbo_stream { render turbo_stream: turbo_stream.replace("form", partial: "form", locals: { category: @category,  categories: @categories }) }
+      end
+    end
```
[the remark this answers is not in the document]
elvisserrao: You're right, I made these changes in [2926b84](https://github.com/maybe-finance/maybe/pull/1730/commits/2926b848b895253fd698a89c79db35713f4a2fe4). 
I believe it's correct now.

app/views/categories/_form.html.erb
```diff
@@ -1,46 +1,49 @@
 <%# locals: (category:, categories:) %>
+<%= turbo_frame_tag 'form' do %>
```
zachgoll, line 2: Do we need this?  What was the purpose of adding?

zachgoll (commented):
Looks good now.  Left one more comment to address, then I think we're good to merge!

app/views/categories/_form.html.erb
elvisserrao: ohhh, i forgot to remove this... it's done in [2c07cd6](https://github.com/maybe-finance/maybe/pull/1730/commits/2c07cd6d19ee5d43e9e4fb1a36e4ed49b3702d89). ;-)

zachgoll (approved):
Looking good!  Nice work, thanks for the fix.

PR #1744, opened 2025-01-30
Add/remove members and invitations

Shpigford:
This PR allows users to remove members from their account/family as well as remove pending invitations.

This functionality is limited to account/family admins.


zachgoll (approved):
Looks good

PR #1745, opened 2025-01-30
Allow users to update their email address

Shpigford:
Users can now update their email address. Upon updating, they'll be emailed a confirmation link.

Self-hosters can disable the need to confirm an email change.


Shpigford:
@zachgoll A couple of the failing tests are transaction related...could that have anything to do with the transaction work over the past 24-48 hours?

zachgoll:
@Shpigford possibly, although the last few commits passed just fine.  It looks like your `schema.rb` has several stale migrations flowing into it.  Can you remove those and re-run the tests?  My guess is that will get things fixed.

Shpigford:
@zachgoll Cleaned up the schema and still hitting those two errors (both here in github and locally). Not sure what else to try there.

zachgoll:
@Shpigford the transaction error is because with the addition of the `:new_email` in the fixtures, it's causing the evaluation of this user to be that user, who has no transactions:

https://github.com/maybe-finance/maybe/blob/ded42a8c33b57f5370fb43b577e71cd31310ef21/test/controllers/transactions_controller_test.rb#L35

We can just flip that to: `sign_in users(:empty)` and that error will be resolved.

app/controllers/users_controller.rb
```diff
@@ -4,10 +4,22 @@ class UsersController < ApplicationController
...
-    @user.profile_image.purge if should_purge_profile_image?
+    if email_changed?
+      if @user.initiate_email_change(user_params[:email])
+        if Rails.application.config.app_mode.self_hosted? && !Setting.require_email_confirmation
+          handle_redirect(t(".success"))
+        else
+          redirect_to settings_profile_path, notice: t(".email_change_initiated")
+        end
+      else
+        redirect_to settings_profile_path, alert: @user.errors.full_messages.to_sentence
```
zachgoll, line 15: There won't be any errors populated on the user model here for most cases (blank alert), so here and below we should probably just pass a string alert like "Something went wrong"

app/views/email_confirmations/confirm.html.erb
```diff
@@ -0,0 +1,4 @@
+<div>
+  <h1 class="font-bold text-4xl">EmailConfirmations#confirm</h1>
+  <p>Find me in app/views/email_confirmations/confirm.html.erb</p>
+</div>
```
zachgoll, line 4: Think we can delete this

app/views/settings/profiles/show.html.erb
```diff
@@ -5,10 +5,11 @@
   <h1 class="text-gray-900 text-xl font-medium mb-4"><%= t(".page_title") %></h1>
   <div class="space-y-4">
     <%= settings_section title: t(".profile_title"), subtitle: t(".profile_subtitle") do %>
-      <%= styled_form_with model: @user, class: "space-y-4" do |form| %>
+      <%= styled_form_with model: @user, url: user_path(@user), class: "space-y-4" do |form| %>
         <%= render "settings/user_avatar_field", form: form, user: @user %>
 
         <div>
+          <%= form.email_field :email, placeholder: t(".email"), label: t(".email") %>
```
zachgoll, line 12: From a UX standpoint, I think we need some sort of message underneath this when `unconfirmed_email` is present:

```erb
<p>You have requested to change your email to <%= @user.unconfirmed_email %>.  Please go to your email and confirm for the change to take effect.</p>
```

config/routes.rb
```diff
@@ -9,6 +9,11 @@
   resources :sessions, only: %i[new create destroy]
   resource :password_reset, only: %i[new create edit update]
   resource :password, only: %i[edit update]
+  resources :email_confirmations, only: [] do
+    collection do
+      get :confirm, as: :confirm_email
+    end
+  end
```
zachgoll, line 16: Totally semantic and optional, but for consistency with the other routes I think this is a great use case for something like:

```rb
resource :email_confirmation, only: :new
```

Would still use `email_confirmations_controller` and the new route would be `new_email_confirmation_url`

app/models/user.rb
```diff
@@ -25,6 +26,47 @@ class User < ApplicationRecord
...
+    if update(
+      email: unconfirmed_email,
+      unconfirmed_email: nil,
+      email_confirmation_sent_at: nil
+    )
+      true
+    else
+      false
+    end
+  end
```
zachgoll, line 68: I think we can remove this method and handle directly in the controller.  Then the controller would just be:

```rb
def confirm
  # Returns nil if the token is invalid OR expired
  @user = User.find_by_token_for(:email_confirmation, params[:token])

  if @user&.update(email: @user.unconfirmed_email, unconfirmed_email: nil)
    redirect_to root_path, notice: t(".success_login")
  else
    redirect_to root_path, alert: t(".invalid_token")
  end
end
```

db/migrate/20250130191533_add_email_confirmation_to_users.rb
```diff
@@ -0,0 +1,9 @@
+class AddEmailConfirmationToUsers < ActiveRecord::Migration[7.2]
+  def change
+    add_column :users, :unconfirmed_email, :string
+    add_column :users, :email_confirmation_token, :string
+    add_column :users, :email_confirmation_sent_at, :datetime
```
zachgoll, line 5: We should be able to delete `:email_confirmation_sent_at` and just let the `generates_token_for` expiry handle the expiration validation (see my comment above about removing the `confirm_email_change` method)

zachgoll (approved):
Looks good!

db/migrate/20250130191533_add_email_confirmation_to_users.rb
```diff
@@ -0,0 +1,9 @@
+class AddEmailConfirmationToUsers < ActiveRecord::Migration[7.2]
+  def change
+    add_column :users, :unconfirmed_email, :string
+    add_column :users, :email_confirmation_token, :string
```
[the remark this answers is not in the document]
Shpigford: Ah, this is leftover from before switching to the rails `generate_token_for` stuff. Will remove.

PR #1803, opened 2025-02-04
Refactor transaction enrichment to support batch processing

Shpigford:


app/models/account/data_enricher.rb
```diff
@@ -8,49 +8,61 @@ def initialize(account)
   end
 
   def run
-    enrich_transactions
-  end
+    total_unenriched = account.entries.account_transactions
+      .where("enriched_at IS NULL OR merchant_id IS NULL OR category_id IS NULL")
```
zachgoll, line 12: Both `merchant_id` and `category_id` are stored on the `Account::Transaction` record, while `enriched_at` is stored on `Account::Entry` (delegated type).

Given we're primarily enriching `Account::Transaction` records (and not other entry types like `Account::Valuation` and `Account::Trade`), I think it probably makes sense to _move_ the `enriched_at` field down to the `Account::Transaction` and only deal with those types of records in this process.

app/models/account/data_enricher.rb
```diff
@@ -8,49 +8,61 @@ def initialize(account)
...
-            next unless entry.name.present?
+    Rails.logger.info("Enriching batch of #{candidates.count} transactions for account #{account.id} (offset: #{offset})")
 
-            info = self.class.synth_provider.enrich_transaction(entry.name).info
+    merchants = {}
 
-            next unless info.present?
+    candidates.each do |entry|
+      begin
+        next unless entry.name.present?
```
zachgoll, line 38: Since this was originally written, we've added null validations to ensure `name` is present, so this can safely be removed now.

app/models/account/data_enricher.rb
zachgoll: Actually, second guessing the idea of moving `enriched_at` to `Account::Transaction`.  This enrichment is also modifying `enriched_name` on the `Account::Entry` model, so it probably makes sense to keep this as-is.

Will still need to update this query to read the filters off the correct DB records though.

app/models/account/data_enricher.rb
```diff
@@ -8,49 +8,61 @@ def initialize(account)
...
   def run
-    enrich_transactions
-  end
-
-  private
-    def enrich_transactions
-      candidates = account.entries.account_transactions.includes(entryable: [ :merchant, :category ])
+    total_unenriched = account.entries.account_transactions
+      .joins(:entryable)
+      .where("account_entries.enriched_at IS NULL OR account_transactions.merchant_id IS NULL OR account_transactions.category_id IS NULL")
```
Shpigford, line 13: Is the entryable join ncessary? AI seems to think so.

zachgoll: The join is necessary, but `joins(:entryable)` is not possible since this is a polymorphic association and will throw an error.  Here's what I would use:

```rb
account.entries.joins("JOIN account_transactions at ON at.id = account_entries.entryable_id AND account_entries.entryable_type = 'Account::Transaction'").where("account_entries.enriched_at IS NULL OR at.category_id IS NULL
 OR at.merchant_id IS NULL")
```

Shpigford:
@zachgoll let me know how this looks now.

zachgoll (commented):
Yep, functionality looks good now! 

The last thing I'm thinking about here is whether we even need `EnrichDataJob` anymore.  All that job is doing now is:

1. Calculating the number of batches
2. Enqueueing the batch jobs

I'm thinking we could get rid of that entirely (the job itself) and do this work directly in `account.enrich_data`. 

So in total, we'd:

- DELETE `account.enrich_data_later` method
- DELETE `EnrichDataJob`

Shpigford:
@zachgoll Good call. Updated. We good to push to production?

zachgoll (approved):
Yep, looks good!

PR #1810, opened 2025-02-05
Add scope to filter transactions from active accounts

Shpigford:


app/models/account/transaction.rb
```diff
@@ -16,6 +16,12 @@ class Account::Transaction < ApplicationRecord
   accepts_nested_attributes_for :taggings, allow_destroy: true
 
   scope :active, -> { where(excluded: false) }
+  # Transactions are associated with accounts through entries
+  scope :from_active_accounts, -> {
+    joins("INNER JOIN account_entries active_entries ON active_entries.entryable_id = account_transactions.id AND active_entries.entryable_type = 'Account::Transaction'")
+    .joins("INNER JOIN accounts active_accounts ON active_accounts.id = active_entries.account_id")
+    .where(active_accounts: { is_active: true, scheduled_for_deletion: false })
+  }
```
Shpigford, line 24: @zachgoll This feels janky as hell, but is it the appropriate way to do the joins on this?

Shpigford: To be clear, it _works_, but just didn't know if there was a simpler way to handle it.

zachgoll: I think the easiest way to do this would be to add a scope directly to `entry.rb`:

```rb
scope :active, -> {
  joins(:account).where(accounts: { is_active: true })
}
```

And then in `transactions_controller.rb`, just tack it on at the end:

```rb
def index
  @q = search_params
  search_query = Current.family.transactions.search(@q).active
  # ...
end
```

PR #1811, opened 2025-02-05
fix: Save completely allocated budget

M123-dev:
I noticed that when I completely budgeted I couldn't save the budget.

Previously, the logic only allowed positive values, which caused issues when handling fully allocated budgets or edge cases where no funds were left to distribute.


zachgoll:
@M123-dev nice catch!

PR #1817, opened 2025-02-06
Multi-factor authentication

Shpigford:
Implements #1372 

https://github.com/user-attachments/assets/b4047e5d-bb24-4c1a-8a2c-e376c8d5eec1

### DIFF
--- a/app/controllers/mfa_controller.rb
+++ b/app/controllers/mfa_controller.rb
@@ -1,0 +1,63 @@
+class MfaController < ApplicationController
+  layout :determine_layout
+  skip_authentication only: [ :verify, :verify_code ]
+
+  def new
+    redirect_to root_path if Current.user.otp_required?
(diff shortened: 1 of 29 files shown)


app/controllers/mfa_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      @backup_codes = Current.user.otp_backup_codes
+      render :backup_codes
+    else
+      Current.user.disable_mfa!
+      redirect_to new_mfa_path, alert: t(".invalid_code")
+    end
+  end
+
+  def verify
+    if Current.session
```
zachgoll, line 22: We can remove the `Current.session` check since we're skipping this route for authentication.

zachgoll (commented):
Everything works great!

From a security standpoint, we should probably protect the `password_resets_controller` with 2FA if the user has it enabled.  Other than that, I think the flow looks good.

PR #1818, opened 2025-02-06
Fix Account Holding validation and synchronization

### LINKED ISSUE #1781: Bug: Holding model should have ActiveRecord and DB null constraints
As seen in https://github.com/maybe-finance/maybe/pull/1767, we're getting some invalid data in the DB because the `Holding` model does not validate `amount`.  

We should ideally be adding both DB-level constraints and ActiveRecord validations on: `date`, `qty`, `price`, `amount`, and `currency`.

Since we have some invalid data states in the DB right now, this will require a migration to clear out invalid data _prior_ to adding the null constraints at the DB level.

Shpigford:
Fixes #1781

- Add comprehensive validations for Account::Holding
- Implement validation to ensure amount matches qty * price
- Update Account::Syncer to include qty and price during synchronization
- Add database constraints for holding attributes and calculations

### DIFF
--- a/db/migrate/20250206204404_add_constraints_to_account_holdings.rb
+++ b/db/migrate/20250206204404_add_constraints_to_account_holdings.rb
@@ -28,25 +28,9 @@ class AddConstraintsToAccountHoldings < ActiveRecord::Migration[7.2]
     change_column_null :account_holdings, :price, false
     change_column_null :account_holdings, :amount, false
     change_column_null :account_holdings, :currency, false
-
-    # Add check constraints
-    execute <<-SQL
-      ALTER TABLE account_holdings
-      ADD CONSTRAINT check_positive_values#{' '}
-      CHECK (qty >= 0 AND price >= 0 AND amount >= 0);
-    SQL
-
-    execute <<-SQL
-      ALTER TABLE account_holdings
-      ADD CONSTRAINT check_amount_matches
-      CHECK (ROUND(qty * price, 4) = ROUND(amount, 4));
-    SQL
   end
 
   def down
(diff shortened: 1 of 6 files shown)


Shpigford:
@zachgoll Is this what you had in mind for addressing #1781 and #1767?

db/migrate/20250206204404_add_constraints_to_account_holdings.rb
```diff
@@ -0,0 +1,56 @@
...
+      ALTER TABLE account_holdings
+      ADD CONSTRAINT check_positive_values#{' '}
+      CHECK (qty >= 0 AND price >= 0 AND amount >= 0);
+    SQL
+
+    execute <<-SQL
+      ALTER TABLE account_holdings
+      ADD CONSTRAINT check_amount_matches
+      CHECK (ROUND(qty * price, 4) = ROUND(amount, 4));
+    SQL
```
zachgoll, line 43: Migration makes sense overall, but I think for these more complex validations (i.e. `check_positive_values` and `check_amount_matches`), we should keep those in `ActiveRecord`.

While the current codebase has some gaps still, I'm trying to move our validations towards the following convention.  Here's a `project-conventions.mdc` format that can be pasted in for the AI (forgot this first time around!):

```
### Convention 7: Use ActiveRecord for complex validations, DB for simple ones, keep business logic out of DB

- Enforce `null` checks, unique indexes, and other simple validations in the DB
- ActiveRecord validations _may_ mirror the DB level ones, but not 100% necessary.  These are for convenience when error handling in forms.
- Complex validations and business logic should remain in ActiveRecord
```

zachgoll (approved):
Yep, that's pretty much what I was thinking!

PR #1819, opened 2025-02-07
feat(import): add currency and number format support for CSV imports

### LINKED ISSUE #1713: Bug: CSV number formats are not sanitized/parsed correctly for non-USD currencies
Currently, we're using a naive sanitization function to deal with CSV import `amount` string values that have non-numeric characters in them.

https://github.com/maybe-finance/maybe/blob/d2a7aef6efc6ab54e44f673fc6506fcb6b821f7e/app/models/import.rb#L146-L149

Instead, we should be asking the user for:

- Currency of transactions in the import
- Separator/Delimiter for each amount value

Based on these options, the CSV import should _default_ to the selected currency and sanitize inputs based on the separator/delimiter option, where the possible currencies are found in `config/currencies.yml` and the possible separator/delimiter options are:

```rb
NUMBER_FORMATS = {
  "1,234.56" => { separator: ".", delimiter: "," },  # US/UK/Asia
  "1.234,56" => { separator: ",", delimiter: "." },  # Most of Europe
  "1 234,56" => { separator: ",", delimiter: " " },  # French/Scandinavian
  "1,234"    => { separator: "",  delimiter: "," }   # Zero-decimal currencies like JPY
}
```

danestves:
Fixes #1713 
/claim #1713

Before this change, importing a CSV with European-style numbers (like `1.234,56`) would result in incorrect values because the parser would strip the decimal comma. Now, users can specify their number format and get accurate imports regardless of their region.

```csv
Date,Name,Amount
2023-05-01,Grocery Store,1.254,54
```

**Before**: Would parse as `125454`
**After**: Correctly parses as `1254.54` when European format is selected

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -2,7 +2,16 @@ class Import < ApplicationRecord
   TYPES = %w[TransactionImport TradeImport AccountImport MintImport].freeze
   SIGNAGE_CONVENTIONS = %w[inflows_positive inflows_negative]
 
+  NUMBER_FORMATS = {
+    "1,234.56" => { separator: ".", delimiter: "," },  # US/UK/Asia
+    "1.234,56" => { separator: ",", delimiter: "." },  # Most of Europe
+    "1 234,56" => { separator: ",", delimiter: " " },  # French/Scandinavian
+    "1,234"    => { separator: "",  delimiter: "," }   # Zero-decimal currencies like JPY
+  }.freeze
+
   belongs_to :family
+
+  before_validation :set_default_currency_and_format
 
   scope :ordered, -> { order(created_at: :desc) }
 
@@ -11,6 +20,8 @@ class Import < ApplicationRecord
   validates :type, inclusion: { in: TYPES }
   validates :col_sep, inclusion: { in: [ ",", ";" ] }
   validates :signage_convention, inclusion: { in: SIGNAGE_CONVENTIONS }
+  validates :currency, presence: true
+  validates :number_format, presence: true, inclusion: { in: NUMBER_FORMATS.keys }
 
   has_many :rows, dependent: :destroy
   has_many :mappings, dependent: :destroy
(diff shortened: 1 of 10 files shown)


db/migrate/20250207014022_add_currency_and_format_to_imports.rb
```diff
@@ -0,0 +1,6 @@
+class AddCurrencyAndFormatToImports < ActiveRecord::Migration[7.2]
+  def change
+    add_column :imports, :currency, :string
```
zachgoll, line 3: I'm thinking we might want to rename this to `default_currency` (and remove the method that's already in the model) because a user is allowed to add a `currency` column in their CSV and assign a mapping to that.

If they explicitly provide us a currency column, we'll want to use that value first.  If not, we use this default value.

zachgoll (changes requested):
Hey @danestves thanks for the contribution!

Overall, I think the code looks good.

Could we add a few tests cases to `import_interface_test.rb` that verifies:

- We're parsing these formats correctly
- The currency logic works (user-provided column takes precedence over the default)

db/migrate/20250207014022_add_currency_and_format_to_imports.rb
zachgoll: Coming back to this after looking at #1814 and I'm realizing that we might need to entirely _remove_ this `currency` configuration column and just stick with the `number_format`.

Since the user can directly map one of their CSV columns to "Currency" and we're defaulting to the family currency anyways, I'm thinking this "Default currency" configuration option might cause more confusion than its worth. 

So in summary, I think we need to:

- Remove this `currency` column and the associated configuration field in the forms
- Put the `number_format` config field side by side with the existing currency mapping columns added in #1814

danestves: For sure! Let me work on that!

app/models/import.rb
```diff
@@ -177,6 +185,39 @@ def parsed_csv
...
+
+      # Return empty string if not a valid number
+      unless sanitized =~ /\A-?\d+\.?\d*\z/
+        return ""
+      end
+
+      sanitized
+    end
+
+    def set_default_currency_and_format
```
zachgoll, line 220: Can we set this to `set_default_number_format` now that we've removed currency?

zachgoll (approved):
Looks great!  Just left one small comment about method naming, but otherwise once tests are passing I think we'll be good to merge this one.

app/models/import.rb
danestves: This is done, changed the method name, now we just need to wait for the tests to pass

danestves: Add another commit since I forgot to change it in `before_validation`

zachgoll:
@danestves make sure to create an Algora account to receive the bounty payout if you haven't already.  Let me know if you run into any issues.

verymilan:
This is a very promising update! However it does not work for me. For some reason, it just jumps back to the default of `1,234.56` and ignores my selection.

zachgoll:
@verymilan fix going in here:

https://github.com/maybe-finance/maybe/pull/1876

PR #1820, opened 2025-02-07
fix: improve CSV number format handling for non-USD currencies

onyedikachi-david:
https://github.com/user-attachments/assets/23433056-89fa-4327-a81b-5468b8ce0efa

Fixes: #1713 
/claim #1713 

## Solution
Implemented a more robust number sanitization system that:

1. Added support for multiple number formats:
   - US/UK/Asia (1,234.56)
   - European (1.234,56)
   - French/Scandinavian (1 234,56)
   - Zero-decimal currencies (1,234)

2. Enhanced the `sanitize_number` method to:
   - Handle different decimal separators and thousand delimiters
   - Preserve the correct numerical value regardless of format
   - Return empty string for non-numeric input
   - Handle edge cases like leading/trailing decimals

3. Added comprehensive tests covering:
   - All supported number formats
   - Currency symbols
   - Negative numbers
   - Edge cases (nil, empty string, non-numeric input)
   - Multiple decimal points
   - Leading/trailing zeros

## Testing
Added a new test file `test/models/import/number_sanitization_test.rb` with 36 assertions covering:

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -1,6 +1,12 @@
 class Import < ApplicationRecord
   TYPES = %w[TransactionImport TradeImport AccountImport MintImport].freeze
   SIGNAGE_CONVENTIONS = %w[inflows_positive inflows_negative]
+  NUMBER_FORMATS = {
+    "1,234.56" => { separator: ".", delimiter: "," },  # US/UK/Asia
+    "1.234,56" => { separator: ",", delimiter: "." },  # Most of Europe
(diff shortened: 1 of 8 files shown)


ianmartorell:
I'm testing out the csv import feature and the amount format doesn't work, would this fix it?

I'm selecting the format 1.234,56 but the data is parsed as if I had selected 1,234.56. Going back to the previous step shows the wrong amount format selected.

https://github.com/user-attachments/assets/79d7330c-a571-4151-a4b7-263c8712d855

PR #1826, opened 2025-02-07
Enhance security information retrieval and handling

Shpigford:
- Add support for operating MIC codes in security info fetching
- Update security uniqueness validation to handle unknown securities
- Improve security creation and update logic in Plaid investment sync
- Update combobox and view components to handle operating MIC codes
- Add unknown flag for securities with incomplete information

Todo:
- [x] Skip price syncing if unknown = true
- [x] Figure out proper migration/cleanup of duplicated ticker data in production


Shpigford:
@zachgoll There's a LOT happening here...definitely got more complex the deeper I dove into addressing the issue. So if something glaring jumps out at you on simplifying this, let me know.

Shpigford:
@zachgoll This is ready for another look.

Two big things that definitely need a closer eye:
1. The deduplication rake tasks
2. Failing tests

db/migrate/20250207194638_adjust_securities_indexes.rb
```diff
@@ -0,0 +1,6 @@
+class AdjustSecuritiesIndexes < ActiveRecord::Migration[7.2]
+  def change
+    remove_index :securities, name: "index_securities_on_ticker_and_exchange_mic"
+    add_index :securities, [ :ticker, :exchange_operating_mic ], unique: true
+  end
```
zachgoll, line 5: We should also throw in a NOT NULL constraint on `ticker` here since that's only enforced on the model right now

zachgoll: That said, we may want to defer the _new_ constraints until after we've run all the data migrations as I believe this migration will fail on our production data.

We can then introduce this after the data migrations (and follow-ups) are complete

app/models/plaid_investment_sync.rb
```diff
@@ -82,12 +82,15 @@ def get_security(plaid_security_id, securities)
       end
 
       return [ nil, nil ] if plaid_security.nil? || plaid_security.ticker_symbol.blank?
+      return [ nil, nil ] if plaid_security.ticker_symbol == "CUR:USD" # internally, we do not consider cash a security and track it separately
```
zachgoll, line 85: Okay so this seems to be the problematic line with the tests.  In the case of a Cash security, we need to return `[ nil, plaid_cash_security ]` rather than `[nil, nil]`. 

So to get tests passing:

1. Revert this early return
2. In `securities.yml` (fixture), change both of the `exchange_mic` keys to `exchange_operating_mic`

PR #1844, opened 2025-02-10
fix: subcategories are not properly handled for budget allocations

### LINKED ISSUE #1665: Bug: Subcategories are not properly handled for budget allocations
**Describe the bug**

There are 2 parts to this bug:

1. Users should not be able to allocate more to subcategories than the parent total
2. Progress bar should only sum up the *parent* totals and should exclude subcategories.  This is consistent with how the totals are summed in actual results in the budget view.

Image

pauleke65:
This PR ensures that subcategory budget allocations cannot exceed the total budget allocated to their parent category.

Changes:
- Updated the allocated_spending method to only sum parent category budgets.
- Added logic in the budget category form to enforce a max allocation based on the parent’s available budget.

Issue Fixed:
Fixes #1665

Testing:
- Tried allocating more to subcategories than the parent allows ✅
- Verified that subcategory allocations are correctly restricted ✅

Demo:

https://github.com/user-attachments/assets/15ea047d-a621-4098-a1d3-8a1804f2fc2a

### DIFF
--- a/app/views/budget_categories/_confirm_button.html.erb
+++ b/app/views/budget_categories/_confirm_button.html.erb
@@ -1,0 +1,11 @@
+<div id="<%= dom_id(budget, :confirm_button) %>">
+  <% if budget.allocations_valid? %>
+    <%= link_to "Confirm",
+        budget_path(budget),
+        class: "block btn btn--primary w-full text-center" %>
+  <% else %>
+    <span class="block btn btn--secondary w-full text-center text-gray-400 cursor-not-allowed">
+      Confirm
+    </span>
+  <% end %>
+</div>
--- a/app/models/budget.rb
+++ b/app/models/budget.rb
@@ -133,10 +133,10 @@ class Budget < ApplicationRecord
   end
 
   # =============================================================================
-  # Budget allocations: How much user has budgeted for all categories combined
+  # Budget allocations: How much user has budgeted for all parent categories combined
   # =============================================================================
   def allocated_spending
-    budget_categories.joins(:category).where(categories: { parent_id: nil }).sum(:budgeted_spending)
+    budget_categories.reject { |bc| bc.subcategory? }.sum(&:budgeted_spending)  
   end
 
   def allocated_percent
--- a/app/models/budget.rb
+++ b/app/models/budget.rb
@@ -136,7 +136,7 @@ class Budget < ApplicationRecord
   # Budget allocations: How much user has budgeted for all parent categories combined
   # =============================================================================
   def allocated_spending
-    budget_categories.reject { |bc| bc.subcategory? }.sum(&:budgeted_spending)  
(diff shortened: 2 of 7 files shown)


zachgoll (commented):
I like this solution!  I think we should rearrange a bit of the code though to fit the budget domain a bit closer.  I've left some ideas on how we could do that.

Finally, I think in order for these client-side validations to properly work, we'll need to update the Turbo stream responses for the update action so that subcategory forms and their directly siblings are replaced each time a subcategory budget is updated.  This will ensure that the `max: ` validation value is directly updated to reflect the new "max" based on the other subcategory sibling budget amounts.

pauleke65:
> Finally, I think in order for these client-side validations to properly work, we'll need to update the Turbo stream responses for the update action so that subcategory forms and their directly siblings are replaced each time a subcategory budget is updated. This will ensure that the `max: ` validation value is directly updated to reflect the new "max" based on the other subcategory sibling budget amounts.


I tried getting the forms to re-update on the new changes, but couldn't quite get it.
I'm still very new to rails, and not familiar with how turbo streams work but I'll give a second look

zachgoll:
@pauleke65 yeah the streams take a bit of getting used to!  I _think_ it can all be achieved here in this file:

https://github.com/maybe-finance/maybe/blob/main/app/views/budget_categories/update.turbo_stream.erb

We'll likely need to add something along the lines of (untested code below):

```erb
<% if @budget_category.subcategory? %>
  <% @budget_category.siblings.each do |sibling| %>
    <%= turbo_stream.update dom_id(sibling, :form), partial: "budget_categories/budget_category_form", locals: { budget_category: sibling } %>
  <% end %>
<% end %>
```

And then over in `_budget_category_form.html.erb`, add the id so it can be referenced:

```erb
<div id=<%= dom_id(budget_category, :form) %> class="w-full flex gap-3">
```

And finally, add that `siblings` method to `budget_category.rb`:

```rb
def siblings
  budget.budget_categories.select { |bc| bc.category.parent_id == category.parent_id && bc.id != id }
end
```

app/views/budget_categories/_confirm_button.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <% if budget.allocations_valid? %>
+    <%= link_to "Confirm",
+        budget_path(budget),
+        class: "block btn btn--primary w-full text-center" %>
+  <% else %>
+    <span class="block btn btn--secondary w-full text-center text-gray-400 cursor-not-allowed">
+      Confirm
+    </span>
+  <% end %>
+</div>
```
zachgoll, line 11: Nice addition!

zachgoll (approved):
Looks great!  Just waiting for tests to pass and will get it merged.

pauleke65:
resolved lint issues

PR #1854, opened 2025-02-12
Enhance Plaid connection management with re-authentication and error handling

### LINKED ISSUE #1583: Setup Plaid re-authentication flow
**Describe the bug**
Some accounts will fail to sync via Plaid.

The `syncs` table shows some items with the following error:
```json
{
  "display_message": null,
  "error_code": "ITEM_LOGIN_REQUIRED",
  "error_message": "the login details of this item have changed (credentials, MFA, or required user action) and a user login is required to update this information. use Link's update mode to restore the item to a good state",
  "error_type": "ITEM_ERROR",
  "request_id": "0x1234abcd",
  "suggested_action": null                                                                                                                                                                                                                                                  
 }                               
```

**Expected behavior**
Maybe should provide the user a way to re-initiate the Plaid login for whichever institution is needing to reestablish a valid session.

Self-hosted, latest container image as of this posting.

Shpigford:
Fixes #1583 

- Add support for Plaid item status tracking (good/requires_update)
- Implement re-authentication flow for Plaid connections
- Handle connection errors and provide user-friendly reconnection options
- Update Plaid link token generation to support item updates
- Add localization for new connection management states

### DIFF
--- a/app/controllers/plaid_items_controller.rb
+++ b/app/controllers/plaid_items_controller.rb
@@ -21,7 +21,10 @@ class PlaidItemsController < ApplicationController
       @plaid_item.sync_later
     end
 
-    redirect_to accounts_path
+    respond_to do |format|
+      format.html { redirect_to accounts_path }
+      format.json { head :ok }
+    end
   end
 
   private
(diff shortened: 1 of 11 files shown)


zachgoll (approved):
Looks good and works well locally.

My guess is that we'll need to tweak that `status` field as we handle more and more Plaid error states, but don't see any reason to try and predict those needs just yet.

PR #1857, opened 2025-02-13
Add new category flow

### LINKED ISSUE #1734: Improvement: New Add Category Flow
The category modal has been slightly redesigned with a focus on improving the selection process for colors and icons.

- Colors & icons are now in a submenu/secondary modal that appears when clicking the edit icon
- This brings up preset colors + icons that users can select.
- On selection, the icon preview updates in both the primary and secondary modals.
- Users can also use a color picker to manually select a HEX value.
- If the user enters an invalid HEX code, it defaults to #000 (open to better validation suggestions).
- If a new category has a parent category, its color will be inherited for consistency.

**Handling low contrast icon colors**

Since we’re introducing custom colors, there’s a high chance that users will pick a low-contrast color (e.g., white on a light background), making icons hard to see

- It'd be great if we detect when an icon’s color is too close to the background color.
- If detected, automatically darken the icon (e.g., switch to dark gray or black) to maintain contrast.

Please let me know if you have any questions as you're building this out
  
[Link to the Figma page](https://www.figma.com/design/lonJmVk3HYkwZoIO7xYP2w/Maybe-App-(Community)?node-id=5548-254)

Image

Image

syedbarimanjan:
A video showing the changes in action:

https://github.com/user-attachments/assets/ee8b7dae-40f9-493a-9ca2-fb7dc9fb6e0a

Fixes #1734 
/claim #1734


app/assets/stylesheets/application.css
```diff
@@ -1 +1,26 @@
...
+.pcr-palette:before{
+    border-radius: 10px !important;
+}
+.pcr-color-chooser{
+    height: 1.5em !important;
+}
+.pcr-picker{
+    height: 20px !important;
+    width: 20px !important;
+}
```
zachgoll, line 26: We've made some recent changes to the Tailwind files, so you'll want to pull those in first.  After that, we should move this to the `@layer components` directive within the `tailwind/application.css` file.

app/javascript/controllers/color_avatar_controller.js
```diff
@@ -19,13 +41,23 @@ export default class extends Controller {
...
     this.avatarTarget.style.borderColor = `color-mix(in srgb, ${color} 10%, white)`;
     this.avatarTarget.style.color = color;
   }
 
+  handleColorChange(e) {
+    const color = e.currentTarget.value;
+    this.updateAvatarColors(color)
+  }
+
   handleParentChange(e) {
```
zachgoll, line 61: Looks like something happened here with the parent change handling:


https://github.com/user-attachments/assets/263d2342-b67d-439e-a5c0-33330e862c4b

app/javascript/controllers/color_picker_controller.js
```diff
@@ -0,0 +1,151 @@
...
+
+      this.colorInputTarget.value = hexColor;
+      this.colorInputTarget.dataset.colorPickerColorValue = hexColor;
+      this.colorPreviewTarget.style.backgroundColor = hexColor;
+      
+      if(contrastRatio < 3.1){
+        this.colorInputTarget.value = "Contrast Ratio is too low, Please select a darker color."
+      }
+    });
+  }
```
zachgoll, line 49: I'm a little worried that we're doing too much here.  I think if we choose to integrate a 3rd party picker library, we shouldn't need this amount of code to simply choose a color.

I think we need to either:

1. Simplify this controller so that it initializes the color picker JS library and acts as a simple bridge between that library and our form (my guess is that the color library can tell us if the contrast is valid?  Or we could at least write a simpler validation)
2. Use the browser's built in `<input type="color" />` (pictured below)

CleanShot 2025-02-13 at 12 34 55

cc @justinfar would you be okay with a basic, native color picker like I've shown above if it drastically reduces the complexity of this feature? (we can style the button to open it better than i did here—this was just for an example)

justinfar: Would prefer to keep the color picker custom if possible, but if this is causing too much complexity I'm fine with a browser native picker

goldenstein64:
usability concern: triggering `mouseup`/`pointerup` from outside the modal always closes it, even when `mousedown` was triggered from the inside. Dragging motions to the outside of the modal feels common with this UI, so it would be annoying to close unintentionally, especially if there was more information written on the previous screen.

https://github.com/user-attachments/assets/bfdae9a5-7f93-4faf-8e58-f5fd9c5cd71e

app/javascript/controllers/color_picker_controller.js
zachgoll: Let's go ahead with the following solution:

- Keep the custom `pickr` library + Stimulus controller
- Simplify the contrast checking to use [WCAG 2.1 guideline for contrast (1.4.3)](https://www.w3.org/TR/WCAG21/#contrast-minimum), which simply checks relative luminance values.
  - The background should be the given color mixed with white at 10% opacity
  - Let's work exclusively with rgba values
  - In total, we should be able to do this with just 3 functions: `backgroundColor(rgba)`, `luminance(rgb)`, and `contrast(luminanceForeground, luminanceBackground)`
- When a color does not meet the contrast requirement of `4.5:1`, the following should happen:
  - Keep the value in the input untouched
  - Put a red border around the input with a red, `text-xs` message above saying "Choose a darker color"
  - Use the [setCustomValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity) method to also add this same validation at the form level (i.e. "Color is too light") to ensure that if the user fails to update their selection, the form will not submit.

zachgoll:
@goldenstein64 would you mind opening a new issue for this?  I agree with your suggestion, but since it applies to all of our modals / drawers in the app, it's probably worth a standalone issue.

app/javascript/controllers/color_picker_controller.js
zachgoll: The `picker.on('change')` event should pass an [HSVaColor Object](https://github.com/simonwep/pickr?tab=readme-ov-file#the-hsvacolor-object), so I was thinking we can use `toRGBA()` for contrast checking, and call `toHEXA()` for the actual `input` value.

app/views/categories/_form.html.erb
```diff
@@ -3,35 +3,55 @@
...
+                  <%= f.radio_button :color, color, class: "sr-only peer", data: { action: "change->color-picker#handleColorChange change->color-avatar#handleColorChange" } %>
+                  <div class="w-6 h-6 rounded-full cursor-pointer peer-checked:ring-2 peer-checked:ring-offset-2 peer-checked:ring-blue-500" style="background-color: <%= color %>"></div>
+                </label>
+              <% end %>
+              <div class="w-6 h-6 rounded-full cursor-pointer" data-color-picker-target="pickerBtn" style="background: conic-gradient(red,orange,yellow,lime,green,teal,cyan,blue,indigo,purple,magenta,pink,red)" ></div>
+            </div>
+            <div class="flex gap-2 items-center hidden flex-col" data-color-picker-target="paletteSection">
+              <div class="flex gap-2 items-center w-full">
+                <div class="w-6 h-6 p-4 rounded-full cursor-pointer" style="background-color: <%= category.color %>" data-color-picker-target="colorPreview"></div>
+                <%= f.text_field :color , data: { color_picker_target: "colorInput", color_avatar_target: "colorInput", color_picker_color_value: "color" }, class: "form-field__input w-auto" %>
```
zachgoll, line 29: There are a few style mismatches here.  In order to clear out the default `styled_form` input styles, you can pass `inline: true` which tells our form builder to omit default styles:

```erb
<%= f.text_field :color , data: { color_picker_target: "colorInput", color_avatar_target: "colorInput", color_picker_color_value: "color" }, inline: true %>
```

app/views/categories/_form.html.erb
```diff
@@ -3,35 +3,55 @@
...
+            <h4 class="text-gray-500 text-sm">Icon</h4>
+            <div class="flex flex-wrap gap-2 mb-4">
+              <% Category.icon_codes.each do |icon| %>
+                <label class="relative">
+                  <%= f.radio_button :lucide_icon, icon, class: "sr-only peer", data: { action: "change->color-avatar#handleIconChange" } %>
+                  <div class="w-6 h-6 rounded cursor-pointer hover:bg-gray-100 peer-checked:bg-gray-100 border-1 border-transparent ">
+                    <%= lucide_icon icon, class: "w-6 h-6 p-1" %>
+                  </div>
+                </label>
+              <% end %>
```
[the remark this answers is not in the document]
zachgoll: Also, the selected icon needs to match the chosen color:
CleanShot 2025-02-18 at 12 08 40

app/javascript/controllers/color_picker_controller.js
```diff
@@ -0,0 +1,126 @@
...
+    };
+    return 0.2126 * toLinear(r) + 0.7152 * toLinear(g) + 0.0722 * toLinear(b);
+  }
+
+  contrast(foregroundColor, backgroundColor) {
+    const fgLum = this.luminance(foregroundColor);
+    const bgLum = this.luminance(backgroundColor);
+    const [l1, l2] = [Math.max(fgLum, bgLum), Math.min(fgLum, bgLum)];
+    return (l1 + 0.05) / (l2 + 0.05);
+  }
```
zachgoll, line 94: Contrast checking looks good!

app/javascript/controllers/color_avatar_controller.js
```diff
@@ -3,13 +3,35 @@ import { Controller } from "@hotwired/stimulus";
...
+            this.updateAvatarColors(color);
+          }
+        });
+      });
+  
+      this.observer.observe(this.colorInputTarget, {
+        attributes: true,
+        attributeFilter: ["data-color-picker-color-value"]
+      });
+    }
```
zachgoll, line 27: I think we should avoid observers like this and update this with one of the following approaches:

1. Combine everything into a single controller (probably the easiest)
2. Use Stimulus outlets to update:

```js
static outlets = ["color-avatar"];

// ...

initPicker() {
 // ...
 this.picker.on("change", (color, source, instance) => {
   const hexColor = color.toHEXA().toString();
   this.colorAvatarOutlet.updateAvatarColors(hexColor);
 }
}
```

app/javascript/controllers/color_picker_controller.js
```diff
@@ -0,0 +1,126 @@
...
+        const span = document.createElement("span");
+        span.style.color = "var(--color-destructive)"
+        span.classList.add("text-sm")
+        span.textContent = "Choose a darker color";
+        this.paletteSectionTarget.prepend(span);
+      } else{
+        this.colorInputTarget.setCustomValidity("");
+        this.colorInputTarget.parentElement.style.border = "1px solid var(--color-alpha-black-100)";
+        const span = this.paletteSectionTarget.querySelector("span");
+        if (span) span.remove();
```
zachgoll, line 65: Validation states look good! 

@justinfar can you do a quick check here on the design for us?  Any changes you'd like?

CleanShot 2025-02-18 at 13 01 18

app/views/categories/_form.html.erb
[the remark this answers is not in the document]
justinfar: @zachgoll @syedbarimanjan re clearing an icon, for now let's just go with requiring an icon for all categories

The generic fallback icon should be the "shapes" icon (i.e the one below)
image

app/javascript/controllers/color_picker_controller.js
zachgoll: @syedbarimanjan I think we can just add a simple function that brings the hue down until it meets the contrast requirement.  Something along these lines; we should be able to reuse quite a bit of the existing contrast checking code:

```js
darkenColor([r, g, b, a]) {
  let darkened = [r, g, b, a];
  let backgroundColor = this.backgroundColor(darkened, 5);
  let contrastRatio = this.contrast(darkened, backgroundColor);
  
  // Keep darkening until we meet contrast requirements or hit black
  while (contrastRatio < 4.5 && (darkened[0] > 0 || darkened[1] > 0 || darkened[2] > 0)) {
    darkened = [
      Math.max(0, darkened[0] - 5),
      Math.max(0, darkened[1] - 5),
      Math.max(0, darkened[2] - 5),
      darkened[3]
    ];
    contrastRatio = this.contrast(darkened, backgroundColor);
  }
  
  return darkened;
}
```

app/javascript/controllers/color_picker_controller.js
```diff
@@ -0,0 +1,218 @@
...
+      if (span) span.remove();
+    }
+  }
+
+  backgroundColor([r,g,b,a], percentage) {
+    const mixedR = Math.round((r * (percentage / 100)) + (255 * (1 - percentage / 100)));
+    const mixedG = Math.round((g * (percentage / 100)) + (255 * (1 - percentage / 100)));
+    const mixedB = Math.round((b * (percentage / 100)) + (255 * (1 - percentage / 100)));
+    return [mixedR, mixedG, mixedB];
+  }
```
zachgoll, line 152: We have several hardcoded values of `5%`, `10%`, and `15%` throughout the controller.  We should add a private method:

```js
#backgroundColor(color) {
  return `color-mix(in oklab, ${color} 10%, transparent)`;
}
```

And then throughout the controller:

- Remove all borders (those are outdated from prev designs)
- Refer to this private method for avatar + icon backgrounds

zachgoll (commented):
Overall looking a lot better!  Core flow works great.

Before we merge, I think there are a few general "cleanup" items to address (several mentioned in comments):

**Model validation / migration**

Currently, the `Category` class allows `nil` for a category.  Per @justinfar comments about the `shapes` icon being our default, we'll need to:

- Add a default value for `lucide_icon` of `shapes` on the `Category` table
- Add a NOT NULL constraint on `lucide_icon`

**Edge Cases in UI**

- When you edit a subcategory, users can change the color to be different than the parent.  This is because we have `before_create :inherit_color_from_parent` in the model, but really, we need this to be `before_commit` to cover both create + update flows + hide the color option in the UI if it is editing a subcategory.
- When a user selects a preset color, then choose a custom color, the preset color remains selected in the UI (see video below):

https://github.com/user-attachments/assets/2da509c8-17e2-4a98-b327-dbce680d61f8

app/javascript/controllers/new_category_controller.js
```diff
@@ -0,0 +1,217 @@
...
+
+    adjustedRGBA = this.darkenColor(currentRGBA).toString();
+
+    this.colorInputTarget.value = adjustedRGBA;
+    this.colorInputTarget.dataset.colorPickerColorValue = adjustedRGBA;
+    this.colorPreviewTarget.style.backgroundColor = adjustedRGBA;
+    this.colorInputTarget.setCustomValidity("");
+
+    this.picker.setColor(adjustedRGBA);
+  }
```
zachgoll, line 134: This method can be simplified to just this since setting the color will trigger all these reassignments:

```js
autoAdjust(e) {
  const currentColor = this.picker.getColor();
  const adjustedRGBA = this.darkenColor(currentColor).toString();
  this.picker.setColor(adjustedRGBA);
}
```

Then simply work with the `HSVaColor` object directly in `darkenColor` method:

```js
darkenColor(color) {
  let darkened = color.toRGBA();
  const backgroundColor = this.backgroundColor(darkened, 10);
  let contrastRatio = this.contrast(darkened, backgroundColor);

  while (
    contrastRatio < 4.6 &&
    (darkened[0] > 0 || darkened[1] > 0 || darkened[2] > 0)
  ) {
    darkened = [
      Math.max(0, darkened[0] - 10),
      Math.max(0, darkened[1] - 10),
      Math.max(0, darkened[2] - 10),
      darkened[3],
    ];
    contrastRatio = this.contrast(darkened, backgroundColor);
  }

  return `rgba(${darkened.join(", ")})`;
}
```

Not a huge deal, but I think the more we can pass around the `HSVaColor` object throughout the controller, the more clear and concise the code becomes and the less conversions we have to deal with.

app/views/categories/_color_avatar.html.erb
```diff
@@ -0,0 +1,10 @@
+<%# locals: (color: "#000") %>
+
+<% background_color = "color-mix(in srgb, #{color} 5%, white)" %>
+<% border_color = "color-mix(in srgb, #{color} 10%, white)" %>
+<span 
+      data-color-picker-target="avatar"
+      class="w-14 h-14 flex items-center justify-center rounded-full"
+      style="background-color: <%= background_color %>; border-color: <%= border_color %>; color: <%= color %>">
+  <%= lucide_icon "shapes" %>
+</span>
```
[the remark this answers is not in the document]
zachgoll: Make sure the icon size and alpha channel are matching what I've suggested here so we're in parity with the Figma designs (transparent, w-8)

app/javascript/controllers/new_category_controller.js
```diff
@@ -0,0 +1,217 @@
+import { Controller } from "@hotwired/stimulus"
+import Pickr from '@simonwep/pickr'
+
+export default class extends Controller {
+  static targets = ["pickerBtn", "colorInput", "colorsSection", "paletteSection", "pickerSection", "colorPreview", "avatar", "details", "icon","validationMessage","selection","colorPickerRadioBtn"];
+
+  _predefined_colors = ["#e99537", "#4da568", "#6471eb", "#db5a54", "#df4e92", "#c44fe9", "#eb5429", "#61c9ea", "#805dee", "#6ad28a"];
```
zachgoll, line 7: We should pass these into the controller as values so that if we ever change these presets in the Category class, we don't have to come here and update this again to preserve functionality:

```js
static values = {
  presetColors: Array,
};
```

```erb
<div data-controller="color-picker" data-color-picker-preset-colors-value="<%= Category::COLORS %>">
```

app/views/categories/_color_avatar.html.erb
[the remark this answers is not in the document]
zachgoll: @syedbarimanjan can you update this to match the code I posted?

app/javascript/controllers/category_controller.js
```diff
@@ -0,0 +1,207 @@
+import { Controller } from "@hotwired/stimulus"
+import Pickr from '@simonwep/pickr'
+
+export default class extends Controller {
+  static targets = ["pickerBtn", "colorInput", "colorsSection", "paletteSection", "pickerSection", "colorPreview", "avatar", "details", "icon","validationMessage","selection","colorPickerRadioBtn"];
+  static values = {
+    presetColors: Array,
+  };
+
+  _predefined_colors = ["#e99537", "#4da568", "#6471eb", "#db5a54", "#df4e92", "#c44fe9", "#eb5429", "#61c9ea", "#805dee", "#6ad28a"];
```
zachgoll, line 10: We can remove this now

zachgoll (commented):
After all comments are addressed and merge conflicts resolved, we'll run the tests and get this merged.  

Functionality looks good!

zachgoll (approved):
Code looks good!  Looks like there is a failing check, so once resolved we're good to merge.

syedbarimanjan:
@zachgoll

zachgoll:
@syedbarimanjan great, nice work!  Let me know if you have any issues with the bounty payment.

zachgoll (approved):

PR #1876, opened 2025-02-21
Fix import configuration failures

### LINKED ISSUE #1860: Data import problem: Date not parsed properly
Importing a list of transactions or trades use to work (on 21 Jan 2025) bit different. The date was recoginized from the imported CSV and changed to a format I wished (in import settings). Unfortunately something must have been changed in the past weeks and now the date column (in format M/D/YYYY HH:MM) is showing error and require me to edit every row to match chosen date format DD-MM-YYYY. I am using same files from 21 Jan 2025 that is proven to work before.

zachgoll:
Fixes #1860

This PR fixes import failures when the user has invalid data after selecting a certain configuration.

In our latest import update, validation was added on the initial loading of rows:

https://github.com/maybe-finance/maybe/blob/c620d1fc1e53aea2f0db6388e67590485b2fd6d7/app/models/import.rb#L109

This is incorrect and we should not be validating rows when loading; they should be validated directly in our "Clean" step in the UI.

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -105,8 +105,8 @@ class Import < ApplicationRecord
   def generate_rows_from_csv
     rows.destroy_all
 
-    csv_rows.each do |row|
-      rows.create!(
+    mapped_rows = csv_rows.map do |row|
+      {
         account: row[account_col_label].to_s,
         date: row[date_col_label].to_s,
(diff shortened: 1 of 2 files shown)


danestves:
Thanks! I'm learning from this too

PR #1823, opened 2025-02-07
New Design System + Codebase Refresh

zachgoll:
Fixes #1848, Fixes #1836, Fixes #1651, Fixes #1330, Fixes #1670, Fixes #1530

CleanShot 2025-02-12 at 11 35 00

Since the very first `0.1.0-alpha.1` release, we've been moving quickly to add new features to the Maybe app.  In doing so, some parts of the codebase have become outdated, unnecessary,  or overly-complex as a natural result of this feature prioritization.

Now that "core" Maybe is complete, we're moving into a second phase of development where we'll be working hard to improve the accuracy of existing features and build additional features on top of "core".  This PR is a quick overhaul of the existing codebase aimed to:

- Establish the brand new and simplified dashboard view (pictured above)
- Establish and move towards the conventions introduced in #1788 
- Consolidate layouts and improve the performance of layout queries
- Organize the core models of the Maybe domain (i.e. `Account::Entry`, `Account::Transaction`, etc.) and break out specific traits of each model into dedicated concerns for better readability
- Remove stale / dead code from codebase
- Remove overly complex code paths in favor of simpler ones

PR #1894, opened 2025-02-24
Fix parent category sums in budget

zachgoll:
The changes in #1823 introduced a regression where if a user had a parent category with _zero_ transactions assigned to it while there were transactions assigned to one or more of its subcategories, the parent category summed to 0.

This is because the totals query filters out any categories with no transactions.  This PR fixes this by ensuring that _all_ family categories are included in the `PeriodTotals` response of the `IncomeStatement`.

### DIFF
--- a/app/models/budget.rb
+++ b/app/models/budget.rb
@@ -100,11 +100,11 @@ class Budget < ApplicationRecord
   end
 
   def income_category_totals
-    income_totals.category_totals.reject { |ct| ct.category.subcategory? }.sort_by(&:weight).reverse
+    income_totals.category_totals.reject { |ct| ct.category.subcategory? || ct.total.zero? }.sort_by(&:weight).reverse
   end
 
   def expense_category_totals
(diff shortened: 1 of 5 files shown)


Foiler25:
Still having this issue in v0.4.3, nothing attached to the parent category, everything in sub-categories.

image

Running self-hosted on Umbrel.

zachgoll:
@Foiler25 is there a chance Umbrel has not yet updated your app to the latest version?  Can you confirm that the latest deploy is on `v0.4.3`?  

This is working fine on both the hosted and self-hosted version for me so we'll need some more info to diagnose.

Foiler25:
> @Foiler25 is there a chance Umbrel has not yet updated your app to the latest version? Can you confirm that the latest deploy is on `v0.4.3`?
> 
> This is working fine on both the hosted and self-hosted version for me so we'll need some more info to diagnose.

Hi @zachgoll after digging a bit deeper, the version that was published on Umbrel says it is 0.4.3 but within Maybe it is showing it currently running 0.4.1, apologies for the mistake, I will reach out through the appropriate channels for Umbrel instead.

PR #1908, opened 2025-02-26
Fix conversion for European format

dmfilipenko:
Follow for this
https://discord.com/channels/1191501936281784361/1342779772970729533/1342779772970729533

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -6,7 +6,8 @@ class Import < ApplicationRecord
     "1,234.56" => { separator: ".", delimiter: "," },  # US/UK/Asia
     "1.234,56" => { separator: ",", delimiter: "." },  # Most of Europe
     "1 234,56" => { separator: ",", delimiter: " " },  # French/Scandinavian
-    "1,234"    => { separator: "",  delimiter: "," }   # Zero-decimal currencies like JPY
+    "1,234"    => { separator: "",  delimiter: "," },  # Zero-decimal currencies like JPY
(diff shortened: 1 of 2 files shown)


henrikbjorn:
Wouldn't it be easier and less error prone to take the string and make it in to characters, then split on the separator and ignore the delimiter.

Something like this

``` ruby
amount = "-233.002,23"

separator = ","
delimiter = "."

whole, cents = amount.chars.split(separator)
whole.reject! { it == delimiter }

p Float("#{whole.join}.#{cents.join}")
```

dmfilipenko:
@henrikbjorn might be. I just follow existing convention

henrikbjorn:
Maybe you can try

``` ruby
def sanitize_number(value)
  return "" if value.nil?

  format = NUMBER_FORMATS[number_format]
  return "" unless format

  # First, normalize spaces and remove any characters that aren't numbers, delimiters, separators, or minus signs
  sanitized = value.to_s.delete(' ').strip

  mains, cents = sanitized.chars.split(format[:separator])
  mains.reject! { it == format[:delimiter] }

  "#{mails.join}.#{cents.join}"
end
```

zachgoll (changes requested):
Thanks for the fix!

I think I agree with @henrikbjorn on this one that this is more of a parsing issue than anything.

To merge this, I we should probably:

- Remove the extra format 
- Update parsing to handle both `1.234,56` and `1234,56` properly
- Update the "Parses European number format correctly" test to have an additional CSV row (for total of 2 rows) with each of these formats

Issue #1863, opened 2025-02-16
Bug: Changing the currency on the "Onboarding preferences" page does not correctly update the UI


mithileshz:
**Where did this bug occur? (required)**

- [x] I am a self-hosted user reporting a bug from my self hosted app
- [ ] I am a user of Maybe's paid app

**Describe the bug**

When changing the currency from USD to GBP during the onboarding phase, the example box account does not update the currency correctly in one instance. The UI displays "$2,323.25" incorrectly, however, it displays "+£78.90 (+£6.39)" correctly.

To reproduce:
1. Create a new account
2. On the "Configure your preferences", change the currency from USD to GBP

**Expected behavior**
The "$2,323.25" value should be "£2,323.25"

**Screenshots / Recordings**

![Image](https://github.com/user-attachments/assets/8eedd059-3484-4d2e-abb4-404b69e9d462)

mithileshz:
Just thought I'd look into it a bit more - This seems to be an issue in this file - app/views/onboardings/preferences.html.erb line 15, due to the string being hardcoded. 

_I can try to make a PR for this if I get the development flow working but if you don't see a PR in a day or two, I'm probably not working on it :)_

PR #1910, opened 2025-02-26
Allow accout balance to dynamically use currency format on preference page

Harry-kp:
fixes https://github.com/maybe-finance/maybe/issues/1863

https://github.com/user-attachments/assets/eafe0963-d994-4e58-b1ce-b34dc0e14db1


zachgoll (approved):
Nice, looks good!

PR #1912, opened 2025-02-27
fix: Don't show Billings on settings navbar when self-hosted

### LINKED ISSUE #1911: Bug: Error when trying to subscribe with incorrect stripe secret key
**Where did this bug occur? (required)**

- [x] I am a self-hosted user reporting a bug from my self hosted app
- [ ] I am a user of Maybe's paid app

**Describe the bug**
Hey folks,

I’m not sure if this is a bug or just a UI enhancement, but when someone without a subscription attempts to subscribe, and the STRIPE_SECRET_KEY is either malformed or empty, it raises an exception: `Stripe::AuthenticationError`

I’ve made a few local changes to avoid raising an exception for the user and giving more information. I’d be happy to open a pull request if it makes sense. 

**To Reproduce**
1. Go to 'settings/billing'
2. Click on 'subscribe' button
3. See error

_Will only break if you have set your secret to either an invalid value like 1234 or is empty_

**Expected behavior**
Instead of just returning the error page, the user should be given more information about what happened so that they can try to fix it.

**Screenshots / Recordings**

https://github.com/user-attachments/assets/1199c1cc-78f1-4445-b080-4eef653f0e1b

tonyvince:
close #1911

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,4 +1,5 @@
 class SubscriptionsController < ApplicationController
+  before_action :redirect_to_root_if_self_hosted
   def new
     if Current.family.stripe_customer_id.blank?
       customer = stripe_client.v1.customers.create(
@@ -43,5 +44,9 @@ class SubscriptionsController < ApplicationController
   private
     def stripe_client
       @stripe_client ||= Stripe::StripeClient.new(ENV["STRIPE_SECRET_KEY"])
+    end
+
+    def redirect_to_root_if_self_hosted
+      redirect_to root_path, alert: I18n.t("subscriptions.self_hosted_alert") if self_hosted?
     end
 end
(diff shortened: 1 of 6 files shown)


tonyvince:
Can't reproduce test failures locally.

zachgoll (changes requested):
@tonyvince thanks for the fix here, but I think instead of trying to fix this error, we should never be experiencing it in the first place.

I think in self hosted mode, we should remove the "Billing" page entirely from the application and disable all Stripe / Billing related code when `SELF_HOSTED=true`

tonyvince:
@zachgoll changed implementation

zachgoll (commented):
I think we'll need to add `condition: :self_hosted?` in our settings helper too:

https://github.com/maybe-finance/maybe/blob/f7064fd4ddc3defcd4404d7b88bbb05420506091/app/helpers/settings_helper.rb#L2-L15

zachgoll (approved):
Looks good!

PR #1913, opened 2025-02-27
Feat: Data "reset" button

### LINKED ISSUE #1901: Feature: Data "reset" button
A common question we get across both the self-hosted and hosted apps is, "How can I reset my data and start over?"

We should implement a "Reset Account" button directly above the "Delete Account" that:

- Can only be performed by an admin (shouldn't show as a button for non-admins)
- Deletes family: Accounts, Categories, Merchants, Tags, PlaidItems, and other user data to get the family back to a "clean slate"

Image

**Self hosted**

Additionally, in self-hosted settings, we should implement a "Clear Data Cache".  This should delete all "cached" data that can be re-created easily:

- Exchange rates (fetched from provider)
- Security prices (fetched from provider)
- Account balances (these are generated by the app and recreated through `Account::Entry`)
- Account holdings (these are generated by the app and recreated through `Account::Trade`)

In the case that a user has bad data that came from a provider, this is a quick and easy way to pull that data fresh again.

Image

### Requirements

- Both actions are performed in background job
- Only admins can see these buttons and perform these actions
- Both should have "Are you sure" confirm modals

tonyvince:
feat: Allow admins to delete family data
feat: Allow self-hosting users to delete cached data

Close #1901

### DIFF
--- a/app/jobs/data_cache_clear_job.rb
+++ b/app/jobs/data_cache_clear_job.rb
@@ -1,0 +1,16 @@
+class DataCacheClearJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    ActiveRecord::Base.transaction do
+      ExchangeRate.delete_all
+      Security.find_each { |security| security.prices.delete_all }
+      Account::Balance.delete_all
+      Account::Holding.delete_all
+
+      # Reset last_synced_at and broadcast refresh for all families
+      Family.update_all(last_synced_at: nil)
+      Family.find_each(&:broadcast_refresh)
+    end
+  end
+end
--- a/app/jobs/data_cache_clear_job.rb
+++ b/app/jobs/data_cache_clear_job.rb
@@ -1,16 +1,16 @@
 class DataCacheClearJob < ApplicationJob
   queue_as :default
 
-  def perform
+  def perform(family)
     ActiveRecord::Base.transaction do
       ExchangeRate.delete_all
-      Security.find_each { |security| security.prices.delete_all }
-      Account::Balance.delete_all
-      Account::Holding.delete_all
+      Security::Price.delete_all
+      family.accounts.each do |account|
+        account.balances.delete_all
+        account.holdings.delete_all
+      end
 
-      # Reset last_synced_at and broadcast refresh for all families
-      Family.update_all(last_synced_at: nil)
-      Family.find_each(&:broadcast_refresh)
+      family.sync_later
     end
   end
 end
--- a/app/jobs/family_reset_job.rb
+++ b/app/jobs/family_reset_job.rb
@@ -6,22 +6,14 @@ class FamilyResetJob < ApplicationJob
     ActiveRecord::Base.transaction do
       # Delete accounts and related data
       family.accounts.destroy_all
-
       family.categories.destroy_all
-
       family.tags.destroy_all
-
       family.merchants.destroy_all
-
       family.plaid_items.destroy_all
-
       family.imports.destroy_all
-
       family.budgets.destroy_all
 
-      # Reset last_synced_at and broadcast refresh
-      family.update!(last_synced_at: nil)
-      family.broadcast_refresh
+      family.sync_later
     end
   end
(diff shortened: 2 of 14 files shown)


tonyvince:
Screenshot 2025-02-27 at 13 35 47
Screenshot 2025-02-27 at 13 36 05
Screenshot 2025-02-27 at 13 36 20
Screenshot 2025-02-27 at 13 36 30
Screenshot 2025-02-27 at 13 36 43

app/jobs/data_cache_clear_job.rb
```diff
@@ -0,0 +1,16 @@
...
+      ExchangeRate.delete_all
+      Security.find_each { |security| security.prices.delete_all }
+      Account::Balance.delete_all
+      Account::Holding.delete_all
+
+      # Reset last_synced_at and broadcast refresh for all families
+      Family.update_all(last_synced_at: nil)
+      Family.find_each(&:broadcast_refresh)
+    end
+  end
```
zachgoll, line 15: Even though we'll be deleting records _across_ families and a self hosted instance will typically just be one family, I still think we should pass in `perform(family)` here and call `family.sync_later` at the end of this:

```rb
ExchangeRate.delete_all
Security::Price.delete_all

family.accounts.each do |account|
  account.balances.delete_all
  account.holdings.delete_all
end

# We can replace the Family.update_all and Family.find_each(&:broadcast_refresh) with this:
family.sync_later
```

app/views/settings/hostings/_danger_zone_settings.html.erb
```diff
@@ -0,0 +1,18 @@
...
+            class: "bg-orange-500 text-white text-sm font-medium rounded-lg px-4 py-2",
+            data: { turbo_confirm: {
+              title: t("settings.hostings.show.confirm_clear_cache.title"),
+              body: t("settings.hostings.show.confirm_clear_cache.body"),
+              accept: t("settings.hostings.show.clear_cache"),
+              acceptClass: "w-full bg-orange-500 text-white rounded-xl text-center p-[10px] border mb-2"
+            }}
+    %>
+  </div>
+</div>
```
zachgoll, line 18: Can we add a `Current.user.admin?` guard for this too?

test/system/settings_test.rb
```diff
@@ -46,6 +76,28 @@ class SettingsTest < ApplicationSystemTestCase
...
+
+    within("#turbo-confirm") do
+      assert_text I18n.t("settings.hostings.show.confirm_clear_cache.title")
+      assert_text I18n.t("settings.hostings.show.confirm_clear_cache.body")
+      click_button I18n.t("settings.hostings.show.clear_cache")
+    end
+
+    assert_current_path settings_hosting_path
+    assert_text I18n.t("settings.hostings.clear_cache.cache_cleared")
+  end
```
zachgoll, line 99: Could we remove these system tests and keep all our testing for this feature in controller/job tests?

We're trying to keep the system test suite as minimal and stable as possible for CI speed and reliability, and I think we're adding in a lot of dependence on the UI structure here, which may change in the future.  The most important tests I think we need for this feature is to validate that:

- Jobs actually delete the intended data
- Admins can't trigger these jobs ever, regardless of what the UI looks like (the UI guards should be purely visual aids)

app/jobs/family_reset_job.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      family.plaid_items.destroy_all
+
+      family.imports.destroy_all
+
+      family.budgets.destroy_all
+
+      # Reset last_synced_at and broadcast refresh
+      family.update!(last_synced_at: nil)
+      family.broadcast_refresh
```
zachgoll, line 24: Similar to above, I'd just replace these with a `family.sync_later`

zachgoll (commented):
Nice work here! Just added a few suggestions, but overall this works great.

zachgoll (approved):
Looks great, nice work!

PR #1940, opened 2025-03-02
Handle period conflicts

### LINKED ISSUE #1939: Bug: Period Key Mapping Issue on March 2nd – current_month and last_day Collide
On March 2nd, the Period class incorrectly maps current_month (MTD) and last_day (1D) to the same date range. Since the key method uses .first when finding a matching period, last_day is selected, causing current_month to be missing from the selection list.

This issue leads to:
	•	Missing current_month (MTD) from the period selection.
	•	Duplicate last_day (1D) appearing twice.

Image

Steps to Reproduce:
1.	On March 2nd, run:
```
Period.all.map { |p| [p.label_short, p.key] }
```
2.	Observe that current_month is missing and last_day appears twice.

Expected Behavior:

Each period should have a unique mapping, and current_month should not be missing.

Root Cause:

The key method currently does:
https://github.com/maybe-finance/maybe/blob/c95bb082a98be370c90fc87a912763950eb59714/app/models/period.rb#L124-L126
On March 2nd, both last_day and current_month resolve to [2025-03-01, 2025-03-02].
•	Since .first is used, last_day is selected, and current_month is skipped.

Suggested Fix:

```
def period_select(form:, selected:, classes: "border border-secondary rounded-lg text-sm pr-7 cursor-pointer text-primary focus:outline-hidden focus:ring-0")
    periods_for_select = Period::PERIODS.map do |key, period|
    [ period[:label_short], key ]
  end
    form.select(:period, periods_for_select, { selected: selected.key }, class: classes, data: { "auto-submit-form-target": "auto" })
end
```

nikhilbadyal:
Fixes #1939

### DIFF
--- a/app/helpers/forms_helper.rb
+++ b/app/helpers/forms_helper.rb
@@ -18,8 +18,9 @@ module FormsHelper
   end
 
   def period_select(form:, selected:, classes: "border border-secondary rounded-lg text-sm pr-7 cursor-pointer text-primary focus:outline-hidden focus:ring-0")
-    periods_for_select = Period.all.map { |period| [ period.label_short, period.key ] }
-
+    periods_for_select = Period::PERIODS.map do |key, period|
+    [ period[:label_short], key ]
+  end
     form.select(:period, periods_for_select, { selected: selected.key }, class: classes, data: { "auto-submit-form-target": "auto" })
 end
 


zachgoll:
@nikhilbadyal I'm going to close this out as I've implemented a fix in #1944 that addresses the root cause of this at the `Period` class level.

#1944 will also make #1941 a bit simpler for you as we're now explicitly passing in period keys rather than using the ambiguous `fallback: true`

PR #1941, opened 2025-03-02
Preference to set default_period

nikhilbadyal:


nikhilbadyal:
@zachgoll  any opinion on this ?

zachgoll (commented):
Yeah just a couple quick thoughts off the top of my head:

- For the preference, we should set this at the `User` level as different users within the same family might want to see different default views
- We should avoid passing `family` into any of the `Period` methods to avoid mixing responsibilities.  Instead, I think we should rework the `fallback` option to accept a `Period` key rather than just a boolean.  Then we simply pass the user's preferred period into `Period.from_key` in the controller
- Sparklines should always use the default period and not be configurable as those are purely to see a trend of the account recently

db/migrate/20250303175124_add_default_period_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultPeriodToUsers < ActiveRecord::Migration[7.2]
+  def change
+    add_column :users, :default_period, :string
```
zachgoll, line 3: Since we're now setting this across several views, I'd be in favor of setting a default of `last_30_days` and then adding a controller concern to set it for each request (ensuring it _always_ has a valid value):

```rb
module Periodable
  extend ActiveSupport::Concern 

  included do
    before_action :set_period 
  end

  private
    def set_period
      @period = Period.from_key(params[:period] || Current.user&.default_period)
    rescue Period::InvalidKeyError
      @period = Period.last_30_days
    end
end
```

app/views/settings/preferences/show.html.erb
```diff
@@ -30,6 +30,11 @@
             { label: t(".country") },
             { data: { auto_submit_form_target: "auto" } } %>
 
+        <%= form.select :default_period,
+            Period::PERIODS.map { |key, period| [period[:label], key] },
+            { label: t(".default_period") },
+            { data: { auto_submit_form_target: "auto" } } %>
```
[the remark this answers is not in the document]
nikhilbadyal: Should i move this below the `Date format` setting ? @zachgoll

nikhilbadyal:
@zachgoll  can you review this.

app/views/accounts/chart.html.erb
```diff
@@ -1,4 +1,4 @@
-<% period = params[:period] ? Period.from_key(params[:period]) : Period.last_30_days %>
+<% period = @period %>
```
zachgoll, line 1: No need to assign this anymore, we can just reference `@period` directly now.

app/controllers/concerns/periodable.rb
```diff
@@ -0,0 +1,15 @@
+module Periodable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :set_period
+  end
+
+  private
+    # @return [void]
```
zachgoll, line 9: We typically try to avoid YARD comments and would not need them here since this is a private method

app/models/user.rb
```diff
@@ -9,6 +9,7 @@ class User < ApplicationRecord
 
   validates :email, presence: true, uniqueness: true, format: { with: URI::MailTo::EMAIL_REGEXP }
   validate :ensure_valid_profile_image
+  validates :default_period, inclusion: { in: Period::PERIODS.keys }, allow_nil: true
```
zachgoll, line 12: To match what's happening at the DB-level, we should remove the `allow_nil: true` from this validation

app/views/accounts/show/_chart.html.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: (account:, title: nil, tooltip: nil, **args) %>
 
-<% period = params[:period] ? Period.from_key(params[:period]) : Period.last_30_days %>
+<% period =  @period ||  Period.last_30_days %>
```
zachgoll, line 3: No need for a fallback here.  We can reference `@period` directly.

nikhilbadyal: `SyncTest` fails will fails if i remove this fallback. Any suggestion ? @zachgoll

zachgoll: Ahh, that's because in `accountable.rb`, the `post_sync` method broadcasts a replacement for this partial.  It's probably more correct for us to pass in `period:` as a local variable to this partial, then explicitly pass in `Period.last_30_days` where we do the broadcast.  It's not a perfect solution as it will clear out the prior period that was being used, but the current code isn't preserving that either, so no need to solve for that right now:

```rb
# accountable.rb
def post_sync
  broadcast_replace_to(
    account,
    target: "chart_account_#{account.id}",
    partial: "accounts/show/chart",
    locals: { account: account, period: Period.last_30_days } # Explicitly pass period to the partial
  )
end
```

nikhilbadyal: I'm not sure i got this correctly. But It won't make much of a difference. I still need to write a fallback. So that  during `request-response` cycle it uses period set in concern and for test case it will fallback to local variable which was passed. 

So it's basically a preference b/w  

**Current**
```
<%# locals: (account:, title: nil, tooltip: nil, **args) %>

<% period =  @period ||  Period.last_30_days %>
```

```
def post_sync
    broadcast_replace_to(
      account,
      target: "chart_account_#{account.id}",
      partial: "accounts/show/chart",
      locals: { account: account }
    )
  end
```

vs 

**New**
```
<%# locals: (account:,period:, title: nil, tooltip: nil, **args) %>

<% default_value_title = account.asset? ? t(".balance") : t(".owed") %>
<% final_period = @period || period %>
```

```
def post_sync
    broadcast_replace_to(
      account,
      target: "chart_account_#{account.id}",
      partial: "accounts/show/chart",
      locals: { account: account, period: Period.last_30_days }
    )
  end
```

zachgoll: These stream broadcasts will bypass the request-response cycle (they're happening over websockets), so I believe that's why the test is failing without the fallback.

That said, I do agree having a fallback here isn't a big deal so let's go with this for now.  Typically, I tend to stay away from directly referencing instance variables in partials as it creates an implicit dependency between the controller / view that we're using it in.

zachgoll (approved):

PR #1946, opened 2025-03-04
Account-level import configuration templates

zachgoll:
This PR helps users who are importing monthly bank statements and need Maybe to "remember" the configuration they used for the previous import on a specific account.  In the video below, notice how configuration properties _and_ mapping values are automatically set for the user on the _second_ import for the same account:

https://github.com/user-attachments/assets/0f9d0bfe-2d9c-4cca-bac4-369c70bd2789

### DIFF
--- a/app/controllers/import/rows_controller.rb
+++ b/app/controllers/import/rows_controller.rb
@@ -2,9 +2,7 @@ class Import::RowsController < ApplicationController
   before_action :set_import_row
 
   def update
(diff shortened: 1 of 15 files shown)


nikhilbadyal:
image
Can we have user selected `default format` as pre selected option. Since a user has setup the default date format. There are high chances that , the to be file being imported will have same date format. It just saves a extra click.

Issue #1846, opened 2025-02-10
Bug: App crashes on large CSV imports


zachgoll:
**Describe the bug**

On very large imports, the app crashes sporadically.

This is likely because on each import, we're simply looping through the rows and creating the records:

https://github.com/maybe-finance/maybe/blob/9e5f1574bcc71570d3e337082f5a78a0efd32d56/app/models/transaction_import.rb#L2-L23

**Improvements required**

While each import is run in a background job already, we should be:

- Chunking the imports
- Using bulk operations when possible

Joelute:
Hi there, I'm looking to work on this issue and improve the performance of the transaction imports. My plan is to import the [`activerecord-import`](https://github.com/zdennis/activerecord-import) gem for batching each import entries into a singular SQL query while keeping validations with AR. 

I'll be collecting all the `Account::Entry` instances into an array and import it at the end. Please let me know how this plan sounds and if I there are anything else I should account for as well.

PR #1962, opened 2025-03-08
perf(imports): Bulk import CSV transactions

Joelute:
Bulk import transactions from CSV instead of importing each row at a time resulting in a optimization in performance and faster imports

| Entries | Before: (s)  | After: (s) |
|:---:|:---:|:---:|
| 1000 | 5.7 | 2.2 |
| 10000 |    93.9   | 16.4 |
| 100000 | 2083.7 | 203.8|

Fixes: #1846.

### DIFF
--- a/app/models/transaction_import.rb
+++ b/app/models/transaction_import.rb
@@ -3,7 +3,7 @@ class TransactionImport < Import
     transaction do
       mappings.each(&:create_mappable!)
 
-      rows.each do |row|
+      transactions = rows.map do |row|
         mapped_account = if account
           account
         else
@@ -13,17 +13,22 @@ class TransactionImport < Import
         category = mappings.categories.mappable_for(row.category)
         tags = row.tags_list.map { |tag| mappings.tags.mappable_for(tag) }.compact
 
-        entry = mapped_account.entries.build \
-          date: row.date_iso,
-          amount: row.signed_amount,
-          name: row.name,
-          currency: row.currency,
-          notes: row.notes,
-          entryable: Account::Transaction.new(category: category, tags: tags),
-          import: self
-
-        entry.save!
+        Account::Transaction.new(
+          category: category,
+          tags: tags,
+          entry: Account::Entry.new(
+            account: mapped_account,
+            date: row.date_iso,
+            amount: row.signed_amount,
+            name: row.name,
+            currency: row.currency,
+            notes: row.notes,
+            import: self
+          )
+        )
       end
+
+      Account::Transaction.import!(transactions, recursive: true)
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/models/transaction_import.rb
```diff
@@ -13,17 +14,29 @@ def import!
...
+      Account::Transaction.import!(transactions)
+
+      # Match corresponding transaction id to the entry.
+      transactions.each_with_index do |transaction, index|
+        entries[index].entryable_id = transaction.id
+        entries[index].entryable_type = "Account::Transaction"
       end
+
+      Account::Entry.import!(entries)
     end
```
zachgoll, line 40: Since we're using Postgres, I think we could use the `recursive` option here to avoid doing all the matching logic and still leverage the benefits of activerecord-import:

```rb
entries = rows.map do |row|
  # Build the entry
  Account::Entry.new(
    account: mapped_account,
    # other attributes
    entryable: Account::Transaction.new(category: category, tags: tags)
  )
end

Account::Entry.import!(entries, recursive: true)
```

https://github.com/zdennis/activerecord-import?tab=readme-ov-file#recursive

Joelute: I've been playing around with it and it doesn't seem like it works well with delegated types. Even with the `recursive` option, I get the error: "Validation failed: Entryable must exist". Rails is expecting the `Entryable` to already be in the database with an ID.

 I've optimized the code a bit so that we are aren't matching between arrays.

zachgoll: @Joelute looks like `activerecord-import`'s `recursive` option doesn't handle `belongs_to` relationships; only `has_many` and `has_one`.  I got things working by flipping the association:

```rb
  def import!
    transaction do
      mappings.each(&:create_mappable!)

      transactions = rows.map do |row|
        mapped_account = if account
          account
        else
          mappings.accounts.mappable_for(row.account)
        end

        category = mappings.categories.mappable_for(row.category)
        tags = row.tags_list.map { |tag| mappings.tags.mappable_for(tag) }.compact

        Account::Transaction.new(
          category: category,
          tags: tags,
          entry: Account::Entry.new(
            account: mapped_account,
            date: row.date_iso,
            amount: row.signed_amount,
            name: row.name,
            currency: row.currency,
            notes: row.notes,
            import: self
          )
        )
      end

      Account::Transaction.import!(transactions, recursive: true)
    end
  end
```

The main reason I'd like to use this `recursive` option is because I think it is the _primary_ benefit we're getting from the additional gem.  If we're not using it, I'd prefer to remove the gem and use `insert_all` for the bulk updates.

Joelute: Interesting, I should take a look into how the recursive option works and the benefits of it. I've updated the PR to reflect those changes.

zachgoll (approved):
Looks good!  Thanks for tackling this.

Joelute:
Just rebased the branch with the new changes.

PR #1978, opened 2025-03-10
Add postStartCommand to devcontainer so dev server will be available

71walceli:
This pull request includes a small change to the `.devcontainer/devcontainer.json` file. The change adds a `postStartCommand` to run `bin/setup` and `bin/dev` after the container starts. This will make sure that Dev server will be able to start and be able to be reached via 127.0.0.1:3000. At current state, it's necessary to start a server manually. This PR should address that and create a more complete dev environment.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -11,6 +11,7 @@
     "PATH": "/workspace/bin:${containerEnv:PATH}"
   },
   "postCreateCommand": "bundle install && npm install",
+  "postStartCommand": "bin/setup && bin/dev",
   "customizations": {
     "vscode": {
       "extensions": [


.devcontainer/devcontainer.json
```diff
@@ -11,6 +11,7 @@
     "PATH": "/workspace/bin:${containerEnv:PATH}"
   },
   "postCreateCommand": "bundle install && npm install",
+  "postStartCommand": "bin/setup && bin/dev",
```
zachgoll, line 14: I don't think we should couple the devcontainer lifecycle directly with the app process management (i.e. `bin/dev`).  

We could probably throw `bin/setup` at the end of the `postCreateCommand` to make things a bit easier on initial setup, but I think users should run `bin/dev` manually.

PR #2008, opened 2025-03-19
Add tags selection and notes input to new transaction form

### LINKED ISSUE #1882: Add tag assignment to new transaction modal + bulk assignments
## Requirements

- User can assign 1+ tags in new transaction modal
- User can bulk assign tags

### Discussed in https://github.com/maybe-finance/maybe/discussions/1310

<sup>Originally posted by **thorgin** October 15, 2024</sup>
### Describe the feature

Add all option in the edit transaction window to the new transaction window.

Right now I have the ability to add notes and tags to each transaction but not while inputting it into the database. It would be better just to make the two window have the same options. This just adds more work in adding transactions and adding tags and notes to the new transactions window will avoid creating any unnecessary added steps.

image
image</div>

tonyvince:
close #1882

### DIFF
--- a/app/controllers/account/transactions_controller.rb
+++ b/app/controllers/account/transactions_controller.rb
@@ -27,7 +27,7 @@ class Account::TransactionsController < ApplicationController
     end
 
     def bulk_update_params
-      params.require(:bulk_update).permit(:date, :notes, :category_id, :merchant_id, entry_ids: [])
+      params.require(:bulk_update).permit(:date, :notes, :category_id, :merchant_id, entry_ids: [], tag_ids: [])
     end
 
     def search_params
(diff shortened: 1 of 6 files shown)


tonyvince:
Screenshot 2025-03-20 at 17 14 55
Screenshot 2025-03-20 at 17 15 43
Screenshot 2025-03-20 at 17 16 58

zachgoll:
@justinfar you cool with this for now?  I think longer term those new multi-step forms will replace this, but seems like a reasonable solution for the time being.

justinfar:
yep I'm good with this for now

and yea I think the multi-step forms are a good candidate for this, but this works for now.

PR #2004, opened 2025-03-19
Use Redis for ActiveJob and ActionCable

zachgoll:
## Self hosters: BREAKING CHANGES

**This PR introduces _breaking changes_ and _action is required_ to keep your self hosted app working and receiving updates**

### What are the changes?

1. **We will be adding Redis as a required service** and moving to Sidekiq for background job processing.  This means we will be removing GoodJob so that all apps are running the same job backend for consistency.
2. **We are removing Render as an "officially supported" self hosting method.**  This will not affect your existing services—it simply means that we as a team will not be "officially" recommending/supporting this hosting method.  Our recommendation moving forward is Docker Compose, which we have a [setup guide for](https://github.com/maybe-finance/maybe/blob/main/docs/hosting/docker.md).

### Why the changes?

As we introduce our new AI chat feature and continue to iterate on our background data syncs, we believe that moving away from a database-backed job backend is advantageous for the long-term quality of the app.

## How to migrate

Choose the migration guide below based on which hosting platform you are using.

### Docker Compose

Below are the steps to migrate your docker compose setup to use the new Maybe stack.

#### How this impacts you

- Background jobs (e.g. "account syncs") will stop working until you make required changes
- No data will be altered

#### Required migration steps

**Step 1: Backup your database**

While the following changes should not affect your database, it is always good to backup your database when making significant changes to a Docker compose setup.

You can make a quick backup locally by:

1. Make sure your Maybe app is running
2. Go to the directory your `compose.yml` file is in
4. Run the command: `docker compose exec postgres pg_dump -U maybe_user maybe_production > backup_$(date +%Y%m%d).sql` where `postgres` represents the name of the database service in your `compose.yml` file.

**Step 2: Stop your service**

```
docker compose down
```

**Step 3: Replace your `compose.yml` file with the new example file**

The new `compose.example.yml` file has been refactored to use the new services and should work with your existing `.env` file.  You can simply copy/paste the new file found here:

https://github.com/maybe-finance/maybe/blob/f56e85facdd07f3c0c8c9bd8018b0b6cd850e974/compose.example.yml

**Step 3: Start your service**

As noted in the [official guide](https://github.com/maybe-finance/maybe/blob/f56e85facdd07f3c0c8c9bd8018b0b6cd850e974/docs/hosting/docker.md), you can pull the latest updates and start the service per usual:

```
docker compose pull # This pulls the "latest" published image from GHCR
docker compose build # This rebuilds the app with updates
docker compose up --no-deps -d
```

### Render

[...]

PR #2017, opened 2025-03-22
devContainer: Use Redis for ActiveJob and ActionCable.

Joelute:
After #2004, we are now using Redis and Sidekiq for background job processing. This means that we will need to update the dev container to utilize those services.


zachgoll (approved):
Nice catch, thanks for the update!

PR #2040, opened 2025-03-31
perf(imports): Bulk import CSV trades

Joelute:
Following up to #1962, we should also bulk import investments from CSV to improve performance.


zachgoll (approved):
Good idea!

PR #2050, opened 2025-04-03
add XBUD as an exchange

vigsterkr:


Shpigford:
We're not currently making use of the exchanges.yml file. These things are largely handled through [Synth](https://synthfinance.com)

PR #2053, opened 2025-04-04
Fix: Notes not saved on transfer transactions

the-spectator:
This fixes a bug where users could not add notes to a transfer transaction.

Before:

https://github.com/user-attachments/assets/9111d5f9-b9ff-46a9-abd8-c9506203357f

After:

https://github.com/user-attachments/assets/24ec290a-b4f3-4667-9cdc-916948151f6b


zachgoll (approved):

PR #2058, opened 2025-04-06
Show Uncategorized segment currency as per budget currency

the-spectator:
Updated the Uncategorized budget segment UI to reflect the actual currency symbol from budget instead of hardcoding $.

Before:
image

After:
image


zachgoll (approved):
Looks good!

Issue #2077, opened 2025-04-10
Feature: Implement Mobile Responsiveness


Shpigford:
## Overview
This project is for implementing mobile responsiveness into the Maybe app. It currently has some very rudimentary responsiveness, but the scope of this is to match the views in Figma.

## Design Files
https://www.figma.com/design/jQmZps9aotfUk8P9aFZbSF/%F0%9F%9F%A1-Responsive?node-id=1-4&m=dev

## Note
We're rapidly making commits to `main` as we get ready for launch in the coming weeks. **We ask that you make regularly commits to your draft PR and make sure to resolve merge conflicts and any non-passing tests as you go** so that there aren't any major merge issues once this is complete.

## Bounty
There is a $1500 bounty on this. To participate in the bounty:

1. Comment `/attempt #2077` to indicate you plan on working on this.
2. Submit a draft PR once you start and include `/claim #2077` along with your general game plan for how you intend to implement this.
3. Change out of draft mode on the PR once you're ready to submit your attempt.
4. Include a short video showing functionality in your PR (drastically speeds up us approving).

In the event of multiple attempts, we generally pick the one that most closes matches the scope with the cleanest code. However, given the size/scope of this, once someone has made clear progress that meets proper code standards, we'll likely assign to them.

Shpigford:
Thank you everyone for the work many of you started on implementing this issue!

Based on @neo773's PR (#2092), we'll be moving forward with their work and have assigned this issue to them!

LEstradioto:
Hi @Shpigford, I know I was 2 hours late. But Ive just sent my complete PR. It is bigger and has more features, but all aligned with the Figma files.

I've noticed some differences between my implementation and others. For instance, I've implemented a mobile menu in the settings with correct page navigation and scroll preservation. Additionally, I've addressed some design aesthetics that were missing, such as fixing the donut loading SVG by adding the correct keyframes.

I'm okay with the first-come, first-served approach. However, I think it would be beneficial to review both PRs.

PR #2083, opened 2025-04-11
mobile responsiveness

Ani-4x:
/claim #2077 ,


Shpigford:
Thank you for the work you started on this! We've moved forward and merged #2092.

PR #2088, opened 2025-04-11
loan: Set the first valuation as the original principal.

Joelute:
|Before: |After: |
|---|---|
|image|image|
|image|image|

Fix: #1645.


zachgoll (approved):
Nice work here.  I'll go ahead and get this merged as it definitely improves the existing view.  We'll likely have to circle back to this when we add our "multi-step wizard" account forms, but this feels like a reasonable interim for right now.

PR #2094, opened 2025-04-12
Small count fix in hosting section

busybox11:
This is a really, really minor change, but somewhat needed as it is one of the first things you see when you open this repo.

Figured it could be useful despite not being very important. :')


zachgoll (approved):
Yep I agree, good to get this one right!

PR #2101, opened 2025-04-13
fix: prevent content overflow in transaction modal by enabling internal scrolling

the-spectator:
Before:

https://github.com/user-attachments/assets/5f255e37-d810-48ab-a4b4-8d7ea0651c64

After:

https://github.com/user-attachments/assets/66b3091c-4c1a-4ec9-b649-5841b798da0c


zachgoll (commented):
Thanks for the fix here.  This modal partial is used in several places across the app.  Would you mind doing a quick check to make sure this doesn't introduce any UI regressions elsewhere?

PR #2111, opened 2025-04-14
Fix: No comma when locality is empty (small fix)

LEstradioto:
Currently when Properties have empty address it shows a single comma on subtitle
image
 
 This PR pushes a fix for that, when no Locality is saved, there is no comma
 
And, perhaps, another fix could be added here.

It seems i18n format should supposed be shown like right side of image? If so, I could just update that too.
 
image

### DIFF
--- a/app/models/address.rb
+++ b/app/models/address.rb
@@ -6,7 +6,7 @@ class Address < ApplicationRecord
       line1: line1,
       line2: line2,
       county: county,
-      locality: locality,
+      locality: locality.present? ? "#{locality}, " : "",
       region: region,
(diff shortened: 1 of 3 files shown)


Shpigford:
@LEstradioto So "proper" for a one-liner would actually be a comma after the street name as well.

`123 Main St, San Francisco, CA 94105 US`

LEstradioto:
Great, thanks. Address string will now follow the I18n formatting and as a one-liner

Shpigford:
Merged. Thanks so much, @LEstradioto!

PR #2138, opened 2025-04-21
Add nice formatting for subtypes on account list

ahatzz11:
On the sidebar for accounts the subtype is currently shown with the `humanize` formatting. This doesn't work super well for some subtypes because it only capitalizes the first letter, and standard formatting for some subtypes are all caps (e.g. IRA). 

This PR introduces a new helper that:
- Uses `titleize` to have title formatting (for cases like "Mutual Fund" instead of "Mutual fund")
- Adds some special cases (e.g. "HSA" instead of "Hsa" or "hsa", and "401(k)" instead of "401k"

Before and after:
image

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -157,15 +157,11 @@ class Account < ApplicationRecord
 
   # Get short version of the subtype label
   def short_subtype_label
-    return nil unless subtype && accountable_type
-    accountable_class = accountable_type.constantize
     accountable_class.short_subtype_label_for(subtype) || accountable_class.display_name
   end
 
   # Get long version of the subtype label
   def long_subtype_label
-    return nil unless subtype && accountable_type
-    accountable_class = accountable_type.constantize
     accountable_class.long_subtype_label_for(subtype) || accountable_class.display_name
   end
 
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -155,6 +155,20 @@ class Account < ApplicationRecord
     first_valuation&.amount_money || balance_money
   end
 
+  # Get short version of the subtype label
+  def short_subtype_label
+    return nil unless subtype && accountable_type
+    accountable_class = accountable_type.constantize
+    accountable_class.short_subtype_label_for(subtype) || accountable_class.display_name
+  end
+
+  # Get long version of the subtype label
+  def long_subtype_label
+    return nil unless subtype && accountable_type
+    accountable_class = accountable_type.constantize
+    accountable_class.long_subtype_label_for(subtype) || accountable_class.display_name
+  end
+
   private
     def sync_balances
       strategy = linked? ? :reverse : :forward
--- a/app/models/balance_sheet.rb
+++ b/app/models/balance_sheet.rb
@@ -59,12 +59,12 @@ class BalanceSheet
           account.define_singleton_method(:weight) do
             classification_total.zero? ? 0 : account.converted_balance / classification_total.to_d * 100
           end
-          
+
           # Define our subtype labeling logic
           account.define_singleton_method(:subtype_label) do
             accountable.subtype_label_for(account.subtype) || accountable.display_name
           end
-          
+
           account
         end.sort_by(&:weight).reverse
       )
--- a/app/models/balance_sheet.rb
+++ b/app/models/balance_sheet.rb
@@ -60,16 +60,6 @@ class BalanceSheet
             classification_total.zero? ? 0 : account.converted_balance / classification_total.to_d * 100
           end
 
-          # Define our subtype labeling logic with short format (for sidebar)
-          account.define_singleton_method(:short_subtype_label) do
-            accountable.short_subtype_label_for(account.subtype) || accountable.display_name
-          end
-
-          # Define long label method (for other views that might need it)
-          account.define_singleton_method(:long_subtype_label) do
-            accountable.long_subtype_label_for(account.subtype) || accountable.display_name
-          end
-
           account
         end.sort_by(&:weight).reverse
       )
--- a/app/models/balance_sheet.rb
+++ b/app/models/balance_sheet.rb
@@ -60,9 +60,14 @@ class BalanceSheet
             classification_total.zero? ? 0 : account.converted_balance / classification_total.to_d * 100
           end
 
-          # Define our subtype labeling logic
-          account.define_singleton_method(:subtype_label) do
-            accountable.subtype_label_for(account.subtype) || accountable.display_name
+          # Define our subtype labeling logic with short format (for sidebar)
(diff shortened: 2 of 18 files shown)


zachgoll (commented):
I agree with the formatting changes here, but I think we should try and keep these labels in one spot in the codebase.  We're already defining them in each accountable class, such as for investment:

https://github.com/maybe-finance/maybe/blob/a7dfafc9072b2986834a0b9fff60e8241b1d82f1/app/models/investment.rb#L4-L17

Long term, I think there's likely a better solution than this, but to leverage these existing definitions + labels, we can define a little helper in the accountable concern and define a singleton method within our "Account Group" struct:

```rb
module Accountable
  # Define empty array to ensure all accountables have this defined
  SUBTYPES = [] 

  class_methods do
      # Given a subtype, look up the "label" for this accountable type
      def subtype_label_for(subtype)
        self::SUBTYPES.find { |subtype_label, subtype_value| subtype_value == subtype }&.first
      end
  end
end
```

Then over in `balance_sheet.rb`, we'd just pass it in:

```rb
AccountGroup.new(
  key: accountable.model_name.param_key,
  name: accountable.display_name,
  classification: accountable.classification,
  total: group_total,
  total_money: Money.new(group_total, currency),
  weight: classification_total.zero? ? 0 : group_total / classification_total.to_d * 100,
  missing_rates?: accounts.any? { |a| a.missing_rates? },
  color: accountable.color,
  accounts: accounts.map do |account|
    account.define_singleton_method(:weight) do
      classification_total.zero? ? 0 : account.converted_balance / classification_total.to_d * 100
    end

    # Define our subtype labeling logic
    account.define_singleton_method(:subtype_label) do
      accountable.subtype_label_for(account.subtype) || accountable.display_name
    end

    account
  end.sort_by(&:weight).reverse
)
```

Then the view simply becomes:

```erb
<%= tag.p account.subtype_label, class: "text-sm text-secondary truncate" %>
```

It's far from a perfect solution given we're mixing a little bit of model / view logic together, but I'd really like to keep the `value / label` pairs defined in a single spot for consistency and so when we make changes to the labels it only has to be changed once.

ahatzz11:
@zachgoll I love this idea. I kind of started going down that path but wasn't really sure the best place to put that mapping. I made those changes! Appreciate the help 👍 

The one note here is that some of them have changed a bit, like `HSA` is now `Health Savings Account`. I don't think this is a bad thing - it now matches 1:1 to the dropdown menu where people can select a subtype. Maybe this is a bit overly verbose, but at least for now it's an improvement. Some day if there's a 'short name' or something we want to add it would be possible.

zachgoll:
@ahatzz11 it would be a slight bit more work, but I think it would be reasonable to define a short/long version of each:

```rb
SUBTYPES = {
  hsa: { short: "HSA", long: "Health Savings Account" },
  401k: { short: "401 (k)", long: "401 (k)" }
}
```

There are likely a few views throughout the codebase we'd have to alter for select options, but I don't think too many:

```erb
<%# something like this %>
<%= form.select :subtype, Depository::SUBTYPES.map { |k, v| [v[:short], k] } ... %>
```

ahatzz11:
@zachgoll Since I started this I might as well keep it going!

I now have a short/long string for each type that has a subtype (for consistency across all subtypes), and to support the new label on the `/accounts` page.

For the dropdown menus when creating the subtype and the `/accounts` page I use `:long`. For the balance sheet I use the `:short` since that sidebar is a bit shorter.

Balance sheet sidebar looks the same as I previously implemented (although an improvement here is now properties/depository subtypes also get the same treatment!):
CleanShot-002497 2025-04-21 at 14 20 52@2x


Here are some screenshots validating there aren't any regressions in the dropdown selectors:
CleanShot-002491 2025-04-21 at 14 18 23@2x

And here's a view of the `/accounts` page with the subtypes (I matched the styling from the balance sheet sidebar):
CleanShot-002495 2025-04-21 at 14 19 31@2x

app/views/accounts/_account.html.erb
```diff
@@ -17,6 +17,9 @@
           </p>
         <% else %>
           <%= link_to account.name, account, class: [(account.is_active ? "text-primary" : "text-subdued"), "text-sm font-medium hover:underline"], data: { turbo_frame: "_top" } %>
+          <% if account.respond_to?(:long_subtype_label) && account.long_subtype_label %>
```
zachgoll, line 20:
```suggestion
          <% if account.long_subtype_label %>
```

app/models/balance_sheet.rb
```diff
@@ -59,6 +59,17 @@ def account_groups(classification = nil)
...
+
+          # Define our subtype labeling logic with short format (for sidebar)
+          account.define_singleton_method(:short_subtype_label) do
+            accountable.short_subtype_label_for(account.subtype) || accountable.display_name
+          end
+
+          # Define long label method (for other views that might need it)
+          account.define_singleton_method(:long_subtype_label) do
+            accountable.long_subtype_label_for(account.subtype) || accountable.display_name
+          end
```
zachgoll, line 71: If I'm not mistaken, now that you've got both of these defined on `account.rb` directly, we can remove these singleton overrides here and things will work fine in the sidebar.

app/models/account.rb
```diff
@@ -155,6 +155,20 @@ def first_valuation_amount
...
+    accountable_class = accountable_type.constantize
+    accountable_class.short_subtype_label_for(subtype) || accountable_class.display_name
+  end
+
+  # Get long version of the subtype label
+  def long_subtype_label
+    return nil unless subtype && accountable_type
+    accountable_class = accountable_type.constantize
+    accountable_class.long_subtype_label_for(subtype) || accountable_class.display_name
+  end
```
zachgoll, line 170: Small nit, but `accountable_class` should automatically be available on each Account instance, so these could be simplified to:

```rb
def short_subtype_label
  accountable_class.short_subtype_label_for(subtype) || accountable_class.display_name
end

def long_subtype_label
  accountable_class.long_subtype_label_for(subtype) || accountable_class.display_name
end
```

zachgoll (commented):
Nice!  Labels look good.  Just added a few potential simplifications.

ahatzz11:
@zachgoll thanks for the comments! I changed those things and cursor also decided to add a little test case. Let me know what you think!

zachgoll (approved):
Looks great!  Once tests pass I'll get it merged.

ahatzz11:
> Looks great! Once tests pass I'll get it merged.

Whoops - cursor wasn't very good about where that field on the model was, I did look for a `bin/test` but didn't realize that was just via `bin/rails`, good to know! Should be good to re-run now.

Issue #2115, opened 2025-04-17
Bug: Loan value reset


ahatzz11:
### General checklist

- [x] I have removed personal / sensitive data from screenshots and logs
- [x] I have searched [existing issues](https://github.com/maybe-finance/maybe/issues?q=is:issue) and [discussions](https://github.com/maybe-finance/maybe/discussions) to ensure this is not a duplicate issue
    
### How are you using Maybe?

- [x] I am a paying Maybe customer (hosted version)
  - Paying Maybe users can also open requests in Intercom (if there is sensitive info involved)
- [ ] I am a self-hosted user

## Bug description

A loan that I have had in maybe for quite a while lost all of its history up to today when I opened the app and it synced. This actually happened to two loans - both from the same connection. I will only have one here since it's a small value

### To Reproduce

I'm unsure if this is reproducible, but:

1. Have a loan connected via plaid
2. Open maybe and see it's value change over time
3. Open maybe one day and value before today is now $0, with current value applied today and an infinite increase

### Expected behavior

I'd expect that my loan value is acurately tracked and stored over time. This impacts the accuracy of my account and my total net worth. One of the loans is my mortgage, so it has a very dramatic impact on my net worth graph. 

As a note - all of my accounts are synced (see screenshot below).

### Screenshots and/or recordings

![Image](https://github.com/user-attachments/assets/b236f86c-3eb2-477c-ba3e-ce63ba566eaf)

![Image](https://github.com/user-attachments/assets/50113c17-059f-40fb-890b-ffe243ee0835)

zachgoll:
I _think_ what happened here is due to this PR we merged recently: https://github.com/maybe-finance/maybe/pull/2088

That PR handles _manual_ loan accounts fine, but we forgot to incorporate it with Plaid (where the "initial balance" wouldn't be defined).

Your data is _not_ lost—we'll just have to update the sync process to capture the original principal value from Plaid and incorporate it into the logic from #2088

ahatzz11:
> I _think_ what happened here is due to this PR we merged recently: [#2088](https://github.com/maybe-finance/maybe/pull/2088)
> 
> That PR handles _manual_ loan accounts fine, but we forgot to incorporate it with Plaid (where the "initial balance" wouldn't be defined).
> 
> Your data is _not_ lost—we'll just have to update the sync process to capture the original principal value from Plaid and incorporate it into the logic from [#2088](https://github.com/maybe-finance/maybe/pull/2088)

[...]

PR #2139, opened 2025-04-22
loans: Remove initial_balance field from Loan model.

Joelute:
This PR features changes to remove the initial_balance field allowing loan accounts to operate using its first entry as the source of truth.

**Note:** I've tested this in manually, however I couldn't fully verify it with Plaid 

Fixes part of #2115.

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -33,8 +33,8 @@ class Account < ApplicationRecord
   class << self
     def create_and_sync(attributes)
       attributes[:accountable_attributes] ||= {} # Ensure accountable is created, even if empty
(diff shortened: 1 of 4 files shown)


zachgoll:
Hey @Joelute, thanks for the fix here, but I think we're going to need a full resolution of #2115 in order to merge.  I think your original solution of adding the original balance to the Loan table made sense and we should keep it.  We'll just need to integrate that domain logic with our Plaid accounts.

zachgoll:
@Joelute I'm going to close this one out for now.  We've got some pretty big changes in #2267 around the Plaid sync and one of the changes I'm working on is dealing with the initial loan balance so I'll knock this one out!

PR #2081, opened 2025-04-11
Feature: Mobile responsiveness

LEstradioto:
/claim #2077

Hi, Im working on this, opening as draft now and have some questions below.
I'd appreciate it if you could provide some insights.

QUESTIONS

1.  Some texts are not the same from current, like buttons have Continue or Next. Should I follow exactly the Figma texts?
2.  Should I implement a Money formatter abbreviated on Assets/Liabilites Totals like 1.5B, 100M or 100K:
   image
If so, how would we handle intl ?
    
3.  Do the Help Button on mobile sidebar should go to Contact (intercom)?
image
4.  I cant see some features, like “Export transactions”, “Download a CSV example?”, “That AI thing on CSV”.

---

I will keep a TODO to stay organized:

- [ ]  LOGIN: Add a Sign in / Sign up switcher at mobile
- [ ]  SINGUP: Avatar border dashed (ust use background-image or another trick)
- [ ]  Step 4 and 5 from Import
- [ ]  Budgeting / Sttings / Assistant

PR #2149, opened 2025-04-24
Feature: Money Abbreviate Formatting

LEstradioto:
Hello!

This adds the abbreviate option to the Money Formatting. This was initially part of #2081, I've split it into this separate PR for clarity.

The method abbreviates values near the Assets/Debts titles. Could possibly be improved, please see the Tests to understand how it formats (it has a way to format with i18n, although internationalization is not yet fully implemented throughout the app I think)

If this feature isn't necessary, I'd be happy to remove it or revise the implementation based on feedback. Let me know if there's anything I can improve. Thanks.

<p align="center">
image
</p>

### DIFF
--- a/app/views/pages/dashboard/_balance_sheet.html.erb
+++ b/app/views/pages/dashboard/_balance_sheet.html.erb
@@ -3,7 +3,7 @@
 <div class="space-y-4">
   <% balance_sheet.classification_groups.each do |classification_group| %>
     <div class="bg-container shadow-border-xs rounded-xl space-y-4 p-4">
-      <h2 class="text-lg font-medium"><%= classification_group.display_name %></h2>
+      <h2 class="text-lg font-medium"><%= classification_group.display_name %> · <span class="text-secondary"><%= classification_group.key == "asset" ? format_money(balance_sheet.total_assets, abbreviate: true) : format_money(balance_sheet.total_liabilities, abbreviate: true) %></span></h2>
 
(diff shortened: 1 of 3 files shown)


zachgoll:
@LEstradioto appreciate the work here and agree that we'd benefit from some sort of abbreviation and summary value in that asset/debt section.

We're prepping pretty hard for our public launch right now, so I won't be able to look at this / merge until after.  These money formatting rules are super nuanced and while the code looks good, I'm a bit hesitant pushing any `Money` class related changes just prior to launch due to the edge cases that might exist here.  After launch we can dig in a bit deeper and make sure all cases have been covered here.

zachgoll:
@LEstradioto sorry for the delay here, we've been pretty busy supporting the launch.  I took a look at our Figma design spec and I'm not seeing any cases where we'll need to abbreviate numbers.  I think the addition of abbreviation formatting also opens us up to a pretty wide number of edge cases given the hundreds of currencies and locales that we're operating in, so from a strategic codebase perspective, we'd like to keep this surface area as small as possible for as long as possible to make sure we're focused on the right things as a team.

Going to close this out for now, but will be sure to drop a comment here if we ever decide this needs to be supported.

LEstradioto:
I understand completly and agree 100%, but since I did cause last Bounty program, I just thought to add here to your pondering. I was referring this Figma view, close to Assets title.

PR #2151, opened 2025-04-24
Improve category deletion styling and form-field__input styling

ahatzz11:
I was playing with some category stuff and noticed some odd styling things. I think there's still lots of these around, but figured I'd try to fix something I was actively seeing. 

- Use `text-primary` for all form-field__inputs to support dark mode
- Improve form field submit buttons to match normal buttons a bit more (sizing)

Here's a before and after, which shows the new input `text-primary` as well as the button improvement on the tag/category deletion:

CleanShot-002567 2025-04-24 at 12 33 43@2x

Another example of the `text-primary` on the form input:
CleanShot-002572 2025-04-24 at 12 37 31@2x

Here are some issues that no longer exist that I've noticed with some of the current styling:

Bright flash on hover in dark mode, and hard to read text on form input:
CleanShot-002570 2025-04-24 at 12 34 35@2x

In light mode, the success button is very hard to read:
CleanShot-002574 2025-04-24 at 12 39 35@2x

One question - A lot of the success buttons in the app are black. I decided to choose the `success` color for these forms, but happy to change it - it doesn't feel very consistent right now. It seems like the `btn` classes are heavily used, and the `.form-field__submit` has it's own thing. I don't know enough about the forms so I'm wondering - should the forms just use the `btn` classes for styling? Or should these be special? I also noticed in dark mode the 'buttons' don't really stand out from the background color (below), not sure if there's any improvement there to make it more of a button like in light mode

CleanShot-002576 2025-04-24 at 12 42 36@2x


app/assets/tailwind/maybe-design-system.css
```diff
@@ -655,7 +656,8 @@
   }
 
   .form-field__submit {
-    @apply cursor-pointer rounded-lg bg-surface p-3 text-center text-white hover:bg-surface-hover;
+    @apply inline-flex items-center gap-2 px-3 py-2 rounded-lg text-sm font-medium cursor-pointer disabled:cursor-not-allowed focus:outline-gray-500;
+    @apply transition-all duration-300;
```
ahatzz11, line 660: I copied this from the `.btn` to try and make this more consistent. Should we just use the `.btn` classes for form submit buttons?

zachgoll:
@ahatzz11 I have a bunch of button / form related UI changes going through in #2154 so I'll make sure to handle these there.  In short, we're a bit scattered with all of these styles right now.  I'm adding in ViewComponent for buttons to standardize all of this in #2154 which should help a ton with the design system overall.

ahatzz11:
> @ahatzz11 I have a bunch of button / form related UI changes going through in #2154 so I'll make sure to handle these there. In short, we're a bit scattered with all of these styles right now. I'm adding in ViewComponent for buttons to standardize all of this in #2154 which should help a ton with the design system overall.

Oh great! that sounds way better, I'll close this one out. Thanks 👍

app/assets/tailwind/maybe-design-system.css
hanhhavip: > I copied this from the `.btn` to try and make this more consistent. Should we just use the `.btn` classes for form submit buttons?

hanhhaviphanhha@gmail.com

Issue #2051, opened 2025-04-03
Bug: Incorrect currency conversion in holding total return


ttkjakub:
### General checklist

- [x] I have removed personal / sensitive data from screenshots and logs
- [ ] I have searched [existing issues](https://github.com/maybe-finance/maybe/issues?q=is:issue) and [discussions](https://github.com/maybe-finance/maybe/discussions) to ensure this is not a duplicate issue
    
### How are you using Maybe?

- [ ] I am a paying Maybe customer (hosted version)
  - Paying Maybe users can also open requests in Intercom (if there is sensitive info involved)
- [x] I am a self-hosted user

### Self hoster checklist

If you are a self-hosted user, please complete all of the information below.  Issues with incomplete information will be marked as `Needs Info` to help our small team prioritize bug fixes.

- Self hosted app commit SHA (find in user menu): [02bfa9f](https://github.com/maybe-finance/maybe/commit/02bfa9f25162452533629cd2749337cc2c505dac)
  - [x] I have confirmed that my app's commit is the latest version of Maybe
- Where are you hosting?
  - [ ] Render
  - [x] Docker Compose
  - [ ] Umbrel
  - [ ] Other (please specify)

---

## Bug description

"Total return" of a holding in an investment account is wrong if trades have different currency from the account itself.

### To Reproduce

1. Create an investment account in your primary currency (SEK in my case)
2. Add a new trade in a different currency (e.g. buy 1 AAPL for 204 USD - current market price at the time of writing this)
3. Observe that:
4. the total return is wrong (should be around 0 since we bought the stock at the current market price);
5. and that the average cost is shown with a wrong currency symbol

### Expected behavior

Currencies should be correct converted when calculating total return of a holding

### Screenshots and/or recordings

The following screenshot demonstrates both issues:
- in blue: average cost should show $204 USD (or be correctly converted to SEK)
- in red: total return and trend should be around 0

<img width="1419" alt="Image" src="https://github.com/user-attachments/assets/6bbbaf6c-fb35-432c-b333-0bfbcb034f0f" />

Ripwords:
![Image](https://github.com/user-attachments/assets/09135ad6-2250-41ff-a21c-6b7668dae021)

The issue applies to other areas of the application as well, when currencies are mixed, the default account's currency is used regardless of what each transaction's selected currency

zachgoll:
I'll have to dig in to find the root cause, but just wanted to outline the expected behavior here:

[...]

PR #2153, opened 2025-04-24
holding: Improve calculations for average costs and convert to account currency.

Joelute:
|Before: |After: |
|---|---|
|image|image |

image

Fixes: #2051.


zachgoll (commented):
Thanks for the changes here.  Just a few points of feedback:

- I would prefer to _extend_ these calculations so that we have both a `holding.avg_cost` and `holding.weighted_avg_cost`.  That way in the future we can provide a UI for the user to select which type of calculation they want.
- Let's make sure and keep our naming super clear.  The existing calculation is a "Simple Avg" while your new calculation is a "Weighted Avg".  In tests and methods I think we should reflect this terminology
- Since we're doing this calculation on _read_, it needs to be very quick (esp. for accounts with lots of holdings).  I think at the very least, we need to get this output number in a single SQL statement.  Ideally, I think this becomes very similar to our `BalanceSheet` and `IncomeStatement` classes, but instead, we'd call it `Portfolio` and do one big aggregate query that calculates all the cost basis across all holdings, then these model-specific methods simply "look up" the values needed from those aggregated results that are queried once, cached, and read many times by the various holdings.

Ultimately, I think moving to a weighted average introduces quite a bit of hidden complexity, so I think the highest impact thing we can do if we're looking for a quick improvement is integrating `exchange_rates` into the existing simple avg. query.

Joelute:
@zachgoll Thanks for the feedback on this PR! For now, I've moved forward with just integrating exchange_rates into the existing avg_cost and remove the weighted average cost as I'm not sure where would be a good place for that sort of calculation. However I would be have to introduce it to the code base later on :)

app/models/holding.rb
```diff
@@ -27,10 +27,19 @@ def weight
...
+        "LEFT JOIN exchange_rates ON (
+          exchange_rates.date = entry.date AND
+          exchange_rates.from_currency = trades.currency AND
+          exchange_rates.to_currency = ?
+        )", account.currency
+      ]))
+      .where(entry: { account_id: account.id })
+      .where(security_id: security.id)
+      .where("trades.qty > 0 AND entry.date <= ?", date)
+      .average("trades.price * COALESCE(exchange_rates.rate, 1)")
```
zachgoll, line 42: For clarity of the _ownership_ here, can we use some of the existing query helpers (see [entryable](https://github.com/maybe-finance/maybe/blob/main/app/models/entryable.rb))?

```rb
avg_cost = account.trades.with_entry.joins( ...rest of query)
```

Also, check `entry.date <= ?` I _think_ that should be `entries.date`

zachgoll (commented):
Overall changes look good now thanks!  Just left one improvement we can make to the query.

app/models/holding.rb
Joelute: Updated to use `with_entry` and `entries` in the latest commit.

I think entry was ok as it was just aliased to `entry` instead of `entries`. But I've updated it to use `entries` instead.

app/models/holding.rb
```diff
@@ -27,10 +27,19 @@ def weight
 
   # Basic approximation of cost-basis
   def avg_cost
-    avg_cost = account.entries.trades
-                  .joins("INNER JOIN trades ON trades.id = entries.entryable_id")
-                  .where("trades.security_id = ? AND trades.qty > 0 AND entries.date <= ?", security.id, date)
-                  .average(:price)
+    avg_cost = Trade
```
zachgoll, line 30:
```suggestion
    avg_cost = account.trades
```

app/models/holding.rb
```diff
@@ -27,10 +27,19 @@ def weight
...
+      .with_entry
+      .joins(ActiveRecord::Base.sanitize_sql_array([
+        "LEFT JOIN exchange_rates ON (
+          exchange_rates.date = entries.date AND
+          exchange_rates.from_currency = trades.currency AND
+          exchange_rates.to_currency = ?
+        )", account.currency
+      ]))
+      .where(entries: { account_id: account.id })
```
zachgoll, line 39:
```suggestion
```

app/models/holding.rb
zachgoll: I'd like to scope this to the account just to stay consistent with our query patterns elsewhere in the app.

Joelute:
Thanks for the suggestions @zachgoll! I've updated the PR to follow the suggestions.

zachgoll (approved):
Awesome, looks great!

PR #2168, opened 2025-04-28
add OPENAI_ACCESS_TOKEN to compose example

letavocado:


fabiand93:
With this change it works if you set the openai token in your .env

Status: Working

compose.example.yml
```diff
@@ -41,6 +41,7 @@ x-rails-env: &rails_env
   DB_HOST: db
   DB_PORT: 5432
   REDIS_URL: redis://redis:6379/1
+  OPENAI_ACCESS_TOKEN: ${OPENAI_ACCESS_TOKEN}
```
zachgoll, line 44:
```suggestion
# NOTE: enabling OpenAI will incur costs when you use AI-related features in the app (chat, rules).  Make sure you have set appropriate spend limits on your account before adding this.
  OPENAI_ACCESS_TOKEN: ${OPENAI_ACCESS_TOKEN}
```

zachgoll (approved):
Just added a little note so self-hosters remember to set spend limits on their account.

PR #2175, opened 2025-04-29
Add rule option to change transaction name

ahatzz11:
This PR allows someone to rename a transaction as part of a transaction rule. This was kind of inspired by #2174. I personally have been wanting a transaction rename, because the transaction list is often a mess of names. 

I think this is probably a good candidate for some AI work in the future (e.g. autodetect transaction name), which would probably be similar to autodetecting merchant, but this would allow for a manual rule as well. 

I'd love some feedback here on the changes I made to `actions_controller.js` - I had started making some of the showing/hiding more agnostic and cursor helped fill in some of the nuances of swapping select<>input types. 

Here's a video showing off:
- New rename transaction functionality
- No regressions in changing the action between select/text types (improvement on some spacing)
  image
- No regressions in deletion of an action

https://github.com/user-attachments/assets/a87492ca-5f7b-4c4b-a7d3-c79e8fdd4296

~@zachgoll I'm curious - what is the future functionality for the `function` type?~

---

**update:**
This PR has been refactored a bunch to include HTML templates that are cloned and injected by the stimulus controller.


app/javascript/controllers/rule/actions_controller.js
```diff
@@ -19,37 +19,76 @@ export default class extends Controller {
...
-      this.#showAndEnableValueSelect();
+      this.#convertToSelect(actionExecutor);
+      this.#showActionValue();
+    } else if (actionExecutor.type === "text") {
+      this.#convertToTextInput();
+      this.#showActionValue();
     } else {
-      this.#hideAndDisableValueSelect();
+      this.#hideActionValue();
     }
```
zachgoll, line 29: This is definitely the direction we'll need to go to introduce a text input but I'd like to try and keep it consistent with the pattern I've built over in `conditions_controller.js` with the `#build[FieldType]For(model)` convention:

https://github.com/maybe-finance/maybe/blob/0946a1497a0865185ad943d0d82a21f71e171d31/app/javascript/controllers/rule/conditions_controller.js#L37-L41

My ultimate goal with this pattern is to move all style + Rails form attribute building out of the Stimulus controller so the Stimulus controller can work with 100% generic HTML and have minimal coupling with the view template.  I'm not totally sold on my existing pattern yet and I haven't found great documentation around using Stimulus for this type of thing.

One thing I started doing that seems to work pretty well is using HTML templates where we build a template in our ERB view (which has access to our app's styles, form structure, etc.), and then cloning the template and show/hiding it in the Stimulus controller.  That _might_ be helpful in this case too.  Here's an example of doing that:

ERB defines HTML template (notice how we're heavily leveraging the Rails form builder, which is _not_ available in a Stimulus controller):

https://github.com/maybe-finance/maybe/blob/0946a1497a0865185ad943d0d82a21f71e171d31/app/views/rule/conditions/_condition_group.html.erb#L30-L35

Stimulus controller stays "dumb" and only has to know which target element to append the template to:

https://github.com/maybe-finance/maybe/blob/0946a1497a0865185ad943d0d82a21f71e171d31/app/javascript/controllers/rule/conditions_controller.js#L14-L21

zachgoll (commented):
I think this rule idea makes sense.  I'm thinking of cases like:

If `amount > 1000` AND `amount is positive`

Set transaction name to: `Paycheck: Employer Name`

app/views/rule/actions/_action.html.erb
```diff
@@ -14,8 +15,18 @@
 
     <%= tag.div class: class_names("min-w-1/2 flex items-center gap-2", "hidden" => !needs_value),
                 data: { rule__actions_target: "actionValue" } do %>
-      <span class="font-medium uppercase text-xs">to</span>
-      <%= form.select :value, action.options || [], {}, disabled: !needs_value %>
+      <span class="font-medium uppercase text-xs" data-rule--actions-target="toSpan">to</span>
+      <%# Initial rendering based on rule.action_executors.first from the form. %>
+      <%# This is currently always SetTransactionCategory from app/models/rule/action_executor/transaction_resource.rb. %>
+      <%# Only the first select option will render here, as subsequent renders are injected by the Stimulus controller, which uses the templates from below. %>
+      <% case action_type
```
ahatzz11, line 22: Now that we have the templates and the injection of those templates from the stimulus controller, this `case` becomes a bit weird since it is only ever used for the initial render. The action type for that initial render is determined by the form, which currently has a hardcoded action (and I don't see this changing since we want a first item in the list):

https://github.com/maybe-finance/maybe/blob/441f436187231ae4a0a696d353dcb38bbbf1eed0/app/views/rules/_form.html.erb#L24

I left a comment here explaining this since there is technically unused code here, and if more types come into play they _could_ be put here, but thought I'd get your opinion on what you'd like to see for the initial load vs subsequent injections.

ahatzz11:
A side effect of this PR is that it fixes some weird sizing issues that seem to exist in production:

CleanShot-002815 2025-05-02 at 16 02 06@2x

app/views/rule/actions/_action.html.erb
ahatzz11: For sake of cleaner code without dead paths, I opted to remove the `case` here since that's only required for the initial render. [This commit](https://github.com/maybe-finance/maybe/pull/2175/commits/690daceba20b853fc26186048fcaf81f86f5719f) is the diff between the `case` and no `case`. This `case` isn't actually needed any more because of the template injection. 

I'm happy to revert this if it makes more sense to have it, but I personally like the cleaner code with comment.

ahatzz11: Another piece of cleanup we can do now that we have template injection on the entire action target (which also means we don't need the `case`), is we no longer have a need for `action_type` or `needs_value`:
- `action_type` was used to determine which one to show for each action. Now that we are injecting rules, this wasn't used. Since the initial load is now hardcoded, we don't really need this
- `needs_value` was used to show/hide the entire action target for cases like `function`. This is also not needed anymore since the controller hides the entire action value for cases that don't need any input: https://github.com/maybe-finance/maybe/pull/2175/files#diff-cd0bc09d066aff1b98283823f34902bd4d57c03b6fd45a3bb2f08ea8f6fa0dadR36-R37

app/views/rule/actions/_action.html.erb
```diff
@@ -2,21 +2,28 @@
...
+    <%= tag.div class: class_names("min-w-1/2 flex items-center gap-2"),
                 data: { rule__actions_target: "actionValue" } do %>
       <span class="font-medium uppercase text-xs">to</span>
-      <%= form.select :value, action.options || [], {}, disabled: !needs_value %>
+      <% case action_type
+         when "select" %>
+        <%= form.select :value, action.options || [], {} %>
+      <% when "text" %>
+        <%= form.text_field :value, placeholder: "Enter a value", class: "form-field__input" %>
+      <% end %>
```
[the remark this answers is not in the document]
zachgoll: Yep, looks good!  You're correct I misspoke on the case statement.  We do not need the `function` type there.

app/javascript/controllers/rule/actions_controller.js
```diff
@@ -19,38 +25,69 @@ export default class extends Controller {
...
-  }
+  #clearFormFields() {
+    const toRemove = [];
 
-  #hideAndDisableValueSelect() {
-    this.actionValueTarget.classList.add("hidden");
-    this.valueSelectEl.disabled = true;
+    // Find all elements to remove, unless it's the "to" span
+    Array.from(this.actionValueTarget.children).forEach(child => {
+      if (child !== this.toSpanTarget) {
```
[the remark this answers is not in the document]
zachgoll: I don't think it's a big deal either way, but I'd probably throw the "to" span directly in each of the templates.  It's obviously slightly duplicative, but I think it increases clarity and simplifies things enough to warrant that.  This way, the replacements all happen as a single string of markup:

```erb
<template data-rule--actions-target="selectTemplate">
  <span class="font-medium uppercase text-xs">to</span>
  <%= form.select :value, [], {} %>
</template>

<template data-rule--actions-target="textTemplate">
  <span class="font-medium uppercase text-xs">to</span>
  <%= form.text_field :value, placeholder: "Enter a value" %>
</template>
```

zachgoll (approved):
Works great!  Left one small suggestion around the templating but otherwise I think this is good to go.

app/javascript/controllers/rule/actions_controller.js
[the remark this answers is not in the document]
ahatzz11: Great idea - I like this a lot more than having a secondary piece of this be updated by the controller instead of inside the template. Definitely cleaner code this way.

ahatzz11:
@zachgoll I agree about the `to` inside each template - that is a lot cleaner. When I was playing (since I had that PR that deleted all my tags) I found an edge case where no options for a specific thing (tags, merchants) was not a great UX, so I added a `(none)` disabled select box for that case, which matches the current drawer text when there is no tag to choose from. 

Let me know what you think!

image

zachgoll (approved):
Yeah I think these improvements and changes all make sense! Looks great!

PR #2186, opened 2025-05-01
Fix initials + profile pictures

ahatzz11:
This PR makes some fixes to the user_avatar. I think this might have broken in some of the recent UI updates. This does include changes from the user_avatar_form in #2185.

- Ensures initials are passed to the `_user_avatar_field` partial
- Uses the small variant for the household list
- Fixes some small spacing on the household list
- Changes for processing the image variant url - I would love some feedback on this change, I'll make a comment on it.

Profile picture set (no image left _and_ improper initials):
CleanShot-002739 2025-05-01 at 16 20 52@2x

No profile picture set, expecting proper initials:
CleanShot-002748 2025-05-01 at 16 28 00@2x

Household page:
CleanShot-002751 2025-05-01 at 16 28 32@2x


app/controllers/application_controller.rb
```diff
@@ -23,4 +24,12 @@ def set_default_chat
...
       @chat = @last_viewed_chat
     end
+
+    def set_active_storage_url_options
+      ActiveStorage::Current.url_options = {
+        protocol: request.protocol,
+        host: request.host,
+        port: request.optional_port
+      }
+    end
```
ahatzz11, line 34: This is something well outside of my current (and small) knowledge of ruby - but after some googling and cursor chatting it seems like the ActiveStorage URLs need to properly be set for the images to work, and I _think_ this has something to do with the variants? If I remove the variant references, the images in the menu/menu component work.

I also added the `preprocessed` to the `:small` variant because we know those will be used. 

If there's a better way to handle this I'm happy to change it - or if this is something that differs between local dev and the real dev I can work around that as well.

zachgoll (approved):
Thanks for this fix!  This is a very weird one and was not happening prior to some of the recent UI changes I made, which leads me to think it has to do with the _context_ in which we're fetching that attachment URL from.

These fixes are certainly better than the failures we're currently having so I'll get this merged and then try and dig in and see why this all of a sudden started happening.

PR #2191, opened 2025-05-02
Use icon helper for all-the-things

ahatzz11:
While pulling in changes from #2154 into #2177 I noticed there were some places that were using the lucide_icon helper instead of the new icon helper. This PR moves the rest of these over. Unfortunately searching for `lucide_icon` still returns a bunch of results because of the field name on a `Category`, but I am pretty confident I got the rest of these.


app/components/link_component.html.erb
```diff
@@ -1,13 +1,13 @@
 <%= link_to href, **merged_opts do %>
   <% if icon && (icon_position != "right") %>
-    <%= lucide_icon(icon, class: icon_classes) %>
+    <%= helpers.icon(icon, class: icon_classes) %>
```
ahatzz11, line 3: Is this `helpers.icon()` the right way of doing this? It seems like anything outside of the view doesn't have access to `icon` directly, but not super familiar with the ruby scoping here.

zachgoll: Yep, `helpers.icon` is the correct way!

app/components/button_component.html.erb
```diff
@@ -1,13 +1,13 @@
 <%= container do %>
   <% if icon && (icon_position != :right) %>
-    <%= lucide_icon(icon, class: icon_classes) %>
+    <%= helpers.icon(icon, class: icon_classes) %>
```
zachgoll, line 3: I think we may need to double-check both this `button_component` and `link_component` to make sure the classes we're passing through don't introduce any conflicts.  I think maybe the ideal solution is that our components pass through valid opts to the icon helper, such as:

```erb
<%= helpers.icon(icon, size: size, color: choose_based_on_variant_fn) %>
```

That would also allow us to _remove_ these icon dimensions from the `buttonish_component.rb` config and just use the sizes that the icon helper defines:

https://github.com/maybe-finance/maybe/blob/441f436187231ae4a0a696d353dcb38bbbf1eed0/app/components/buttonish_component.rb#L43

zachgoll (commented):
Other than the quick check on our "Buttonish" components, all looks good!

app/components/button_component.html.erb
ahatzz11: Good feedback, thanks! I think I've implemented what you were looking for now. In theory we could have some kind of separate `icon_size` on the buttonish components, but I kind of like that icon size being tied to the size of the button itself, which should help ensure nice dimensions too. Let me know if my update wasn't quite what you were looking for.

zachgoll (approved):
Yep, that's what I was thinking!

PR #2200, opened 2025-05-03
Add ability to delete all tags

ahatzz11:
I've done a few imports that I messed up and have a lot of tags that I don't want/care about, and I'd love to be able to start fresh. I think having the ability to mass select tags (or categories) for deletion would be great as well in the future, but as a starting point I thought I'd bring tags up to the same functionality that categories have.

https://github.com/user-attachments/assets/5a970982-cf13-445c-b463-cc2831a9a22e

I also noticed, when copying this from the categories, that the capitalization of the confirmation button was a bit strange. I updated the `CustomConfirm` to have `titlelize` for the title & button, and `downcase` for the text. I think this is a bit nicer looking personally, but happy to revert if others feel differently. 

CleanShot-002820 2025-05-02 at 19 57 47@2x

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -34,6 +34,11 @@ class TagsController < ApplicationController
     redirect_to tags_path, notice: t(".deleted")
   end
 
+  def destroy_all
+    Current.family.tags.destroy_all
+    redirect_back_or_to tags_path, notice: "All tags deleted"
+  end
+
   private
 
     def set_tag
--- a/app/helpers/custom_confirm.rb
+++ b/app/helpers/custom_confirm.rb
@@ -6,7 +6,7 @@ class CustomConfirm
       new(
         destructive: true,
         high_severity: high_severity,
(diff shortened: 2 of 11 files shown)


app/controllers/tags_controller.rb
```diff
@@ -34,6 +34,11 @@ def destroy
     redirect_to tags_path, notice: t(".deleted")
   end
 
+  def destroy_all
+    Current.family.tags.destroy_all
+    redirect_back_or_to tags_path, notice: "All tags deleted"
+  end
```
ahatzz11, line 40: I'm still new to rails, but some of the functionality you can build out in a handful of lines of code is kinda insane

zachgoll: The appeal of Rails for sure!

zachgoll (approved):
Agree, this is a nice feature for users who have a mess of tags they want to cleanup.  And definitely matches up with our existing "delete all" pattern.

PR #2205, opened 2025-05-04
Fix shrinking `FilledIconComponent`

haydenbleasel:
Before:

CleanShot 2025-05-04 at 16 34 23@2x

After:

CleanShot 2025-05-04 at 16 34 33@2x


zachgoll (approved):

Issue #2203, opened 2025-05-03
Bug: Transactions bulk update is broken


javiermartingonzalez:
### General checklist

- [X] I have removed personal / sensitive data from screenshots and logs
- [ ] I have searched [existing issues](https://github.com/maybe-finance/maybe/issues?q=is:issue) and [discussions](https://github.com/maybe-finance/maybe/discussions) to ensure this is not a duplicate issue
    
### How are you using Maybe?

- [ ] I am a paying Maybe customer (hosted version)
  - Paying Maybe users can also open requests in Intercom (if there is sensitive info involved)
- [X] I am a self-hosted user

### Self hoster checklist

_Paying, hosted users should delete this entire section._

If you are a self-hosted user, please complete all of the information below.  Issues with incomplete information will be marked as `Needs Info` to help our small team prioritize bug fixes.

- Self hosted app commit SHA (find in user menu): 441f436187231ae4a0a696d353dcb38bbbf1eed0
  - [X] I have confirmed that my app's commit is the latest version of Maybe
- Where are you hosting?
  - [ ] Render
  - [X] Docker Compose
  - [ ] Umbrel
  - [ ] Other (please specify)

---

## Bug description

Trying to bulk update transactions is broken with a "Content missing" message. It was working 1-2 months ago.

### To Reproduce

Be as specific as possible so Maybe maintainers can quickly reproduce the bug you're experiencing.

1. Go to Transactions
2. Select multiple transactions
3. In the tooltip showed in the bottom on the screen, press the Edit button to open bulk edition
4. A "Content missing" error appears on the tooltip and nothing is opened

### Expected behavior

To open the transactions bulk edit dialog

### Screenshots and/or recordings

We highly recommend providing additional context with screenshots and/or screen recordings.  This will _significantly_ improve the chances of the bug being addressed and fixed quickly.

<img width="485" alt="Image" src="https://github.com/user-attachments/assets/3500f447-4812-435c-800f-da8e08220b00" />

zachgoll:
This should be all fixed now.

Issue #2195, opened 2025-05-02
Bug: Cannot assign a merchant to my transactions


esclear:
### General checklist

- [ ] I have removed personal / sensitive data from screenshots and logs
- [x] I have searched [existing issues](https://github.com/maybe-finance/maybe/issues?q=is:issue) and [discussions](https://github.com/maybe-finance/maybe/discussions) to ensure this is not a duplicate issue
    
### How are you using Maybe?

- [ ] I am a paying Maybe customer (hosted version)
  - Paying Maybe users can also open requests in Intercom (if there is sensitive info involved)
- [x] I am a self-hosted user

### Self hoster checklist

If you are a self-hosted user, please complete all of the information below.  Issues with incomplete information will be marked as `Needs Info` to help our small team prioritize bug fixes.

- Self hosted app commit SHA (find in user menu): ` v0.5.0 (bc7e32d)`
  - [x] I have confirmed that my app's commit is the latest version of Maybe
- Where are you hosting?
  - [ ] Render
  - [x] Docker Compose
  - [ ] Umbrel
  - [ ] Other (please specify)

---

## Bug description

I can't assign any merchant to my transactions, because the merchant dropdown / list is empty.

### To Reproduce

I installed the application, set up a few accounts and imported transactions.
At some point, I decided that I'd like to use merchants, so I did the following:

1. Go to the families' merchants
2. Add a merchant
  
![Image](https://github.com/user-attachments/assets/9b324253-3ac9-4089-8403-d66dd864cba8)

3. Go to any transaction, edit it
4. The mechant dropdown is empty

![Image](https://github.com/user-attachments/assets/8bd30acf-721d-4d63-882f-1f36a02933ae)

Interestingly, the merchants _are_ listed when trying to edit _multiple_ transactions:

![Image](https://github.com/user-attachments/assets/0ef83535-4f74-4621-9c44-e3960940aeec)

However, when saving that dialogue, nothing happens, looking at my browser's console, I find the following error:

```
Error invoking action "click->bulk-select#submitBulkRequest"

 TypeError: form is null
    _resetFormInputs https://maybe.example.com/assets/controllers/bulk_select_controller-560f4e0d.js:98
    _addHiddenFormInputsForSelectedIds https://maybe.example.com/assets/controllers/bulk_select_controller-560f4e0d.js:86
    submitBulkRequest https://maybe.example.com/assets/controllers/bulk_select_controller-560f4e0d.js:38
[...]

PR #2213, opened 2025-05-06
fix: bulk updater

LEstradioto:
This is a ~~quick~~ fix for the bulk updater problem #2203 and #2195
~~(probably not ideal since it doesn't follow view component structure)~~

### DIFF
--- a/app/components/dialog_component.rb
+++ b/app/components/dialog_component.rb
@@ -33,7 +33,7 @@ class DialogComponent < ViewComponent::Base
     end
   end
 
-  attr_reader :variant, :auto_open, :reload_on_close, :width, :disable_frame, :opts
+  attr_reader :variant, :auto_open, :reload_on_close, :width, :disable_frame, :frame, :form, :opts
 
   VARIANTS = %w[modal drawer].freeze
   WIDTHS = {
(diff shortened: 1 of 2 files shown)


zachgoll:
@LEstradioto thanks for the contribution, but this isn't quite the fix we're looking for.  This just required a small update to the frame references of the dialog as shown in this commit: https://github.com/maybe-finance/maybe/commit/2e0794b8e164effaa89c77549e0c1a44de7ad952

Going to close out as this is now fixed.

LEstradioto:
Hey @zachgoll 

That update (as I had done too) don't solve the record updaes, since `bulk_select_controller` searches for the closest form.
But the form is not there. As seen in this current error:

image

Due to the limitations of ViewComponent, specifically the rendering order of slots, it's not feasible to nest `with_sections` within the form as currently structured.

And if you inspect html you will find a solo form :

image

zachgoll:
@LEstradioto ah, I see what you mean.  I had thought the error was just a frames issue. 

The dialog component is fairly flexible with its structure, so we can opt-out of the default layout by just throwing everything in the `body` slot and just using raw view component instances rather than the dialog slots:

https://github.com/maybe-finance/maybe/blob/0fdeebceb10525574047bb77d89c398b82837b4b/app/views/transactions/bulk_updates/new.html.erb

PR #2214, opened 2025-05-06
fix: Rule notification should not be triggered when category is unassigned

the-spectator:
## Motiviation

Currently, when a transaction's category is changed, the rule CTA (category_rule) is triggered. However, it incorrectly assumes that a category is always present, which causes issues when the category is unassigned (i.e., nil). This leads to an error when transaction.category is `nil`.

```
flash[:cta] = {
          type: "category_rule",
          category_id: transaction.category_id,
          category_name: transaction.category.name
        }
```


zachgoll (approved):
Good catch!

PR #2223, opened 2025-05-08
Config: put Redis service in Docker local network

JIBSIL:
The old config exposed the Redis server to the internet if the user had not configured a firewall to block port 6379

Issue ref: https://github.com/maybe-finance/maybe/issues/2192


simon-isler:
Nice, thanks @JIBSIL!

PR #2225, opened 2025-05-08
Fix dark mode welcome screen for self-hosting

ahatzz11:
This is a super tiny PR, but I decided to start up a self-hosted instance and noticed this screen was a bit weird so figured I'd fix it quickly.

CleanShot-002916 2025-05-07 at 11 25 17@2x


Shpigford:
Thanks @ahatzz11!

PR #2227, opened 2025-05-08
Use single list for desktop and mobile nav bars

ahatzz11:
When working on the rules, I noticed that the "Rules" item in the settings nav wasn't available on mobile. Digging into this, desktop and mobile were using two different hardcoded list. I've changed this so both views are dynamically generated from a single list, which simplifies the nav bar code, ensures rules shows up on mobile, and future proofs the nav bars from diverging.

This also moves the `"Rules"` text to the locales file.


zachgoll (approved):
Good idea here!

PR #2234, opened 2025-05-11
Fix AND prefix on rule form

### LINKED ISSUE #2230: Improve how conditions are shown/referenced in the rule list
From #2177:

> I had thrown a lot of this checking in a bit hastily for a quick fix a few weeks back and I'm totally fine keeping it as is for now.  But probably worth noting that we'll eventually need to capture this a bit more clearly.  Just an idea, but something that looks more like:
>
>   ```erb
> <span class="px-2 py-1 border border-secondary rounded-full">
>   <%= rule.primary_condition_title %>
> </span>
>
>  <% if rule.conditions.count > 1 %>
>    and <%= rule.conditions.count - 1 %> more <%= rule.conditions.count - 1 == 1 ? "condition" : "conditions" %>
>  <% end %>
>  ```

_Originally posted by @zachgoll in https://github.com/maybe-finance/maybe/pull/2177#discussion_r2081762981_

I also have personally noticed that there is a lot of weirdness in how groups look/work that is a bit confusing, like showing `AND` when its the first conditional group. I think there's some cleanup that can be done around this. 

This issue is slightly different than #2221 which is more about unintentional form submissions and extra inputs being created.

ahatzz11:
Currently, there are a handful of weird things when creating/editing rules around the prefix for a condition:
- When creating a new rule, there is no AND prefix on any condition until the form is submitted
- When creating a new rule, there is always the prefix on a condition group, even if its the first condition on the rule
- When editing a rule with multiple conditions, the saved conditions have the prefix, but new conditions do not
- When editing a rule with multiple conditions, if the first condition is deleted, the new first condition has the prefix

I'm sure there are some more odd cases for different circumstances, but ultimately it seems like the issues are:
- The prefix isn't updated on any change
- The index for an item that isn't yet saved is a very large number, and not `0`, which breaks the current logic.

To solve for all of this weird behavior, I've done a few things:
- Add a new method to our stimulus controller to dynamically add/remove the prefix, filtering out hidden items (those marked for deletion)
- Run the stimulus method when form is rendered, adding a condition or group, and removing a condition or group. Removing a condition is done by the condition controller, so instead of duplicating the prefix logic, I look for the closet rule controller to the condition controller and use the prefix update method from there
- I fell down a deep and dark rabbit hole trying to dynamically show `AND` vs `OR` on the 2+ subcondition in a condition group based on the `ANY` or `ALL` operators. This added a lot of complexity for a small payoff in my opinion, since the any/all operator already clearly show what the condition will be. Maybe this could be a future improvement, but it didn't seem worth the complexity at the moment. 

Here's a video showing proper prefixes for all the actions (adding condition, group, subconditions, removals, edits, etc):

https://github.com/user-attachments/assets/b121eeb3-115e-4a2f-9de3-ffc4f5ae097c

Fixes #2230

This also implements the `primary_condition_title` field on a rule that [Zach recommended here](https://github.com/maybe-finance/maybe/pull/2177#discussion_r2081762981).

### DIFF
--- a/app/javascript/controllers/rule/conditions_controller.js
+++ b/app/javascript/controllers/rule/conditions_controller.js
@@ -21,11 +21,24 @@ export default class extends Controller {
   }
 
   remove(e) {
+    // Find the parent rules controller before removing the condition
+    const rulesEl = this.element.closest('[data-controller~="rules"]');
+
+    // Remove the condition
     if (e.params.destroy) {
       this.destroyFieldTarget.value = true;
       this.element.classList.add("hidden");
     } else {
       this.element.remove();
+    }
+
+    // Update the prefixes of all siblings from the parent rules controller
+    if (rulesEl) {
+      const rulesController = this.application.getControllerForElementAndIdentifier(rulesEl, "rules");
+      if (rulesController && typeof rulesController.updateConditionPrefixes === "function") {
+        rulesController.updateConditionPrefixes();
+        console.log("updated prefixes")
+      }
     }
   }
 
--- a/app/javascript/controllers/rule/conditions_controller.js
+++ b/app/javascript/controllers/rule/conditions_controller.js
@@ -24,7 +24,6 @@ export default class extends Controller {
     // Find the parent rules controller before removing the condition
     const rulesEl = this.element.closest('[data-controller~="rules"]');
 
-    // Remove the condition
     if (e.params.destroy) {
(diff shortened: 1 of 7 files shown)


zachgoll (approved):
Nice, this looks good!  Good call on the complexity/value tradeoff, I 100% agree that it's not worth overloading this with complexity at this stage.

PR #2240, opened 2025-05-14
Create shared ruler and standardize across views

ahatzz11:
When building out some of the design for the rules, one of the things that came up was how the divider between each rule should look. There were quite a few ways this was done before - `<hr>`, a `<div>` with `h-px`, some places had a `ruler` partial, even an instance of `border-b last:border-b-0`.

This PR introduces a new pattern to standarize all of this a bit with the usage of a `shared/ruler` partial. This ruler uses the `border-divider` utility (an alias of `border-tertiary`, if this is a dumb alias happy to change it), with a defualt of `mx-4` to handle the majority use case of separating items in a list (tags, rules, categories, merchants, etc). 

This new partial is very small, but I think is required for use with the `spacer_template` for collection rendering (I also moved a few things over to this pattern to match some of the others). The normal `<hr ..>` line is easy enough to put everywhere but automatic spacers is nice, so I thought I'd extend it to everything. 

Right now I allow customization through a `classes` local, for places that require a full width spacer or y padding. In general, there are very few actual design differences in this PR. There were a few lists where the spacer did not appear under the icon where it does now, but things are consistent now. Here's an example from merchants:
CleanShot-003016 2025-05-14 at 07 48 17@2x

If there's a different direction for this that is better/more idiomatic, I'm happy to adjust!


zachgoll (approved):
Agree on consolidating all of these into a shared partial!

I'm not 100% sure on the design end whether the divider would always have the same type of padding, so will defer to @justinfar on that one.

ahatzz11:
I agree that they probably won't all be the same, which is why I decided to have the `classes` option. I do use this in a few places to maintain the current styling (e.g. a full-width line on the menu popup

CleanShot-003021 2025-05-14 at 13 40 31@2x

If we want we could avoid any kind of default and pass in the padding/other each time, but I opted to default to the most common at this point. Definitely happy to tweak this though.

justinfar:
@zachgoll @ahatzz11 

They’re not meant to be consistent across the app, we use different divider styles depending on the context, so it's fairly flexible in terms of usage. I'll need to set some time and define this in the design system. 

There's three main patterns for dividers within list items:
- Full-width divider – strong visual break between sections
- Padded divider (left/right) – separates items while maintaining their relationship within the same group
- Leading icon with content divider – highlight icon/graphic as the primary differentiator within list group

ahatzz11:
> @zachgoll @ahatzz11
> 
> They’re not meant to be the same across the app, we use different divider styles depending on the context, so it's fairly flexible in terms of usage. I'll need to set some time and define this in the design system.
> 
> There's three main patterns for dividers within list items:
> 
> * Full-width divider – strong visual break between sections
> * Padded divider (left/right) – separates items while maintaining their relationship within the same group
> * Leading icon with content divider – highlight icon/graphic as the primary differentiator within list group
> 
> In your screenshot @ahatzz11, the padded divider works well for the Merchants table. But if that were a regular list with multiple related items (as is usually the case with that component), a leading icon with content divider might make more sense to maintain grouping.

Okay that is good to know! Having something that shows what should use which type of divider would be awesome - and maybe in the future we implement some kind of `variant` for the ruler/divider so that's a bit more standardized. 

The merchant one (in my screenshot) is one of two things where I _changed_ the design - so good to know that one is fine! The other one I changed was the accounts on the dashboard balance sheet. Here is a before and after. I think the right padding on the before in this is a bit strange, but I can see why it would make sense to have the divider to be past the icon on the left in this instance:

CleanShot-003056 2025-05-16 at 13 13 47@2x

Here's what it would look like with the the leading icon with divider:

CleanShot-003058 2025-05-16 at 13 18 24@2x

zachgoll:
@ahatzz11 I'll be honest, my design sense isn't the greatest, so will need @justinfar to do final approvals on this. 

@justinfar what's the correct padding for those account groups?

justinfar:
Let's use the one the leading icon with content divider for the account groups, which you showed in your last image @ahatzz11 

Gap is 12px and each of these list items have a padding of `16px` for L,R and `12px` for T,B

The nested list items have a left padding of `48px`

ahatzz11:
@justinfar sounds great! I think this is all updated and complete now. Here are some additional screenshots showing off dividers throughout the app from this PR:

CleanShot-003076 2025-05-20 at 13 32 36@2x

CleanShot-003090 2025-05-20 at 13 52 02@2x

zachgoll (approved):
Looks great! Thanks for consolidating all of these.

PR #2243, opened 2025-05-15
transfer: Support transfers of different currencies between accounts. 

### LINKED ISSUE #1852: Feature: Auto match multi-currency transfers
### Discussed in https://github.com/maybe-finance/maybe/discussions/1850

<sup>Originally posted by **pbendus** February 12, 2025</sup>
### Describe the feature

Currently, when transferring money between accounts with different currencies (e.g., USD to local currency in Revolut), the app does not recognize these transactions as transfers. Instead, it prompts the user to create a separate transaction manually.

This feature request proposes an enhancement that would allow the app to automatically match multi-currency transfers across accounts. The system should be able to detect linked transactions based on similar timestamps and amounts (after currency conversion) and categorize them as a transfer.

* Helps users properly track money movements between their own accounts without manual intervention.
* Eliminates the need to manually categorize transfer transactions.
* Many users frequently deal with currency conversions in modern banking apps like Revolut, Wise, and others. Auto-matching transactions would make tracking finances much more seamless.

A potential implementation could match transactions based on exchange rates, timestamps, and amounts within a reasonable variance range.</div>

Joelute:
Allow transfer matcher to match transfers to a different currency.

image

Fixes: #1852.


pbendus:
How do we upvote this lol?

Shpigford:
I'll let @zachgoll ultimately weigh in on this, but adding at least a couple of tests for this will go a long way to getting in merged.

zachgoll (commented):
@Joelute thanks for the addition here!

Per @Shpigford's comments, we'll need some concise and simple tests that clearly demonstrate that this query works and does not break any of our existing functionality for auto matching and manual matching.

We should also combine this logic into the existing `transfer_match_candidates` query.  There's a ton of logic to reason about with this process, so keeping it centralized in a single query allows use to debug things in one spot if it were to break.

Joelute:
bugbot run

Shpigford:
bugbot run

test/models/family/auto_transfer_matchable_test.rb
```diff
@@ -53,4 +98,72 @@ class Family::AutoTransferMatchableTest < ActiveSupport::TestCase
...
+      date: 10.days.ago.to_date,
+      rate: 0
+    )
+    create_transaction(date: 10.days.ago.to_date, account: @depository, amount: 500)
+    create_transaction(date: 10.days.ago.to_date, account: @credit_card, amount: -700, currency: "CAD")
+
+    assert_no_difference -> { Transfer.count } do
+      @family.auto_match_transfers!
+    end
+  end
```
zachgoll, line 139: I think even when the rate is `0` we should go ahead and run the standard matching logic.

A value of `0` in the exchange rates table would be considered "bad data" and I don't think it is the responsibility of the transfer matcher to handle. 

So in other words, let's remove this test as I don't think it's testing the right thing.

app/models/family/auto_transfer_matchable.rb
```diff
@@ -24,12 +22,29 @@ def transfer_match_candidates
...
+      .where("
+        (
+          inflow_candidates.currency = outflow_candidates.currency AND
+          inflow_candidates.amount = -outflow_candidates.amount
+        ) OR (
+          inflow_candidates.currency <> outflow_candidates.currency AND
+          exchange_rates.rate IS NOT NULL AND
+          exchange_rates.rate != 0 AND
+          outflow_candidates.amount != 0 AND
+          ABS(inflow_candidates.amount / (outflow_candidates.amount * exchange_rates.rate)) BETWEEN 0.95 AND 1.05
```
zachgoll, line 45: I think we can avoid all of the null / 0 checking with a NULLIF:

```sql
OR (
  inflow_candidates.currency <> outflow_candidates.currency AND
  ABS(inflow_candidates.amount / NULLIF(outflow_candidates.amount * exchange_rates.rate, 0)) BETWEEN 0.95 AND 1.05
)
```

zachgoll (commented):
Looking a lot better! left a few suggestions, but otherwise looks good.

app/models/family/auto_transfer_matchable.rb
Joelute: Looks cleaner than the current conditions. Updated in the latest commit.

test/models/family/auto_transfer_matchable_test.rb
Joelute: Sounds good, I've removed the testcase in the latest commit.

zachgoll (approved):
Looks good now! Thanks for working through the feedback on this one!

AlexDaimler:
Hi Guys could you please let me know how to update to this? thanks.

PR #2244, opened 2025-05-16
Don't override changed name from plaid

ahatzz11:
I've noticed that when I update the name of an account, the next plaid sync will revert the custom name. 

I don't have a plaid sandbox (I probably should do this at some point) so I don't have a for sure way to validate this, but I think this should do the trick. If I'm off the right path here, happy to dig further!

### DIFF
--- a/app/models/plaid_account.rb
+++ b/app/models/plaid_account.rb
@@ -20,7 +20,10 @@ class PlaidAccount < ApplicationRecord
 
         internal_account = family.accounts.find_or_initialize_by(plaid_account_id: plaid_account.id)
 
-        internal_account.name = plaid_data.name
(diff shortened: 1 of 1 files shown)


Joelute (approved):
Did a quick simple test with Plaid sandbox just now and worked just fine. Changing the name and syncing doesn't override the changed name.

image

ahatzz11:
> Did a quick simple test with Plaid sandbox just now and worked just fine. Changing the name and syncing doesn't override the changed name.
> 
> image

Thanks for the validation @Joelute!

zachgoll (approved):
Yep, this should work!  Thanks for the fix.

Issue #2251, opened 2025-05-16
Add Redis check and "clear syncs" button on self hosted instances


zachgoll:
Self hosted apps _require_ Redis to work properly (see #2004)

A great contribution would be the addition of:

1. A "check" that runs before each request (only on self hosted instances) to verify if the app can connect to Redis properly.  If not, show a warning in the UI to alert users to fix this.
2. A button that says "Clear syncs cache" on the "Self Hosted" settings page that simply updates all current syncs to `failed` (to get rid of the loading state)

PR #2253, opened 2025-05-17
syncs: Allow self host clients to clear sync cache.

Joelute:
Add button to self host settings to allow admins remove sync cache.

Fixes part of #2251.
image
image


Joelute:
Yeah, seems like the SyncCleanerJob does this hourly. Will close the PR :)

PR #2256, opened 2025-05-17
Fix subconditions and condition group form

### LINKED ISSUE #2247: Bug: Rules : group of condition --> diseapear
Hello

I am using the hosted version in order to use the sync features.

I create some rules with conditions : 

Several conditions, any rule match, set category 

I save the rule, I open it again --> some conditions have disapear (for example, I have 6 rules, 2 disappeared)

I redefine the rule by adding the two missing rules, save and reopen the rule : in some cases, the rule is correctly saved or not

### LINKED ISSUE #2248: Bug: Rules with conditon group --> condition disappear
Hello

Hosted version

I create a rule

A group condition, any rule apply, 6 conditions, define a category

I save the rule

Reopen the rule

Some conditions (2 in average) have disappeared

I add them again, save and same issue

### LINKED ISSUE #2249: Bug: Rules, conditions disappear
Hello

Hosted version

I create a rule

A group condition, any rule apply, 6 conditions, define a category

I save the rule

Reopen the rule

Some conditions (2 in average) have disappeared

I add them again, save and same issue

ahatzz11:
These 3 tiny changes took me hours - this was really weird and as a rails/ruby novice I got tripped up on a lot, but it's working now!

There's ultimately a two main bugs fixed by this:
- When saving 3 or more conditions, only the first 2 conditions are saved
- When editing an existing condition, an extra subcondition is immediately rendered, meaning a validation error occurs if you were to change something else, e.g. a name or action

I noticed quite a few things debugging all of this:
- When adding subconditions they were _always_ given the same value. The html template also had the same value. This resulted in the form submitting with duplicate ids/names, so only the last one was saved. The first subcondition that automatically appeared had a value of `0`, so it was also saved. This was a result of the top level `rules_controller` doing a replace_all on the `IDX_PLACEHOLDER` of both the condition _and_ subcondition template. When the subcondition template was duplicated, it had the unique key from the parent no matter what - resulting in many subconditions having the same index.
- Once I changed `IDX_PLACEHOLDER` on the subcondition to `IDX_CHILD_PLACEHOLDER`, each subcondition was properly getting the uniquekey from the conditions_controller, but the function there returned an alphanumeric string. After a bunch of chatting with o3, I found out that there seems to be something in rails that makes an alphanumeric key unpermitted. I updated the `#uniqueKey()` function in the condition controller to match the same function in the rules controller.
- After this, all of the subconditions were created, but on an edit of the form you still got a new subcondition each time - and I think this was because the condition_group always rendered a new condition. I've updated this to only render persisted subconditions, and the controller still can dynamically add new conditions.

Here is a video showing some weird behavior before:

https://github.com/user-attachments/assets/fbdfa7fb-e65d-40b7-8526-bfc47d0726f4

and after:

https://github.com/user-attachments/assets/0706b89b-dce5-417d-bdb5-838316152c41

I still think there's some room for improvement on this page - like each rule showing that there is a condition _group_ instead of a single condition from that group, but that's going to be outside of this.

fixes #2247
fixes #2248
fixes #2249


zachgoll (approved):
Thanks for tracking this down!  Definitely a tricky one!

PR #2258, opened 2025-05-18
feat(providers): SimpleFIN integration

cameronroudebush:
# SimpleFIN integration

This PR introduces a new provider "SimpleFIN" which allows another way of account syncing

## What is SimpleFIN?

SimpleFIN is similar to Plaid but much less of a pain on end developers. You can read more [about it here.](https://beta-bridge.simplefin.org/)

## Why integrate SimpleFIN?

SimpleFIN allows self hosters, like myself, to have automated account updates without having to deal with Plaid.

Many similar projects also implement this provider which can be seen in their [blogpost here.](https://www.simplefin.org/ecosystem.html)

# Major Changes

> [!caution]
> I am in no means a ruby developer. This is my first time actually even trying to use ruby. If things don't seem "ruby like" please let me know and I can try to address! I did try to align this closely with how Plaid is implemented but I'm sure it's not perfect.

- Adds a new way to add certain accounts (Depositories, Investments, Loads, Credit Cards) in the add menu.
- Creates new components and providers specific to SimpleFIN.
- Adds 3 new environment variables related to SimpleFIN
  - `SIMPLE_FIN_ACCESS_URL`: Is how we actually connect to SimpleFIN
    - This also determines if SimpleFIN capabilities should be enabled
  - `SIMPLE_FIN_LOOK_BACK_DAYS`: How many days back we should look for transactional data, default is 7 days.
  - `SIMPLE_FIN_RATE_LIMIT`: How many queries to SimpleFIN we should allow in a day, default is 24.
    - We batch process most syncs but in the event you're adding a bunch of accounts, it's at least 2 api calls per group add.
- Adds error displays so when sync errors are detected, we can now easily be notified and given a link to update it.
  - This is both in the account header (includes text and an icon) and in the accountable_group sidebar (just an icon).
- Adds new display for SimpleFIN account groupings by institution within the accounts list.
- Adds new simple_fin related files
  - `controllers/simple_fin_controller`: Adds the actual creation and new account picking endpoints for component display.
  - `provider/simple_fin`: Adds the actual web requests so we can pull data from SimpleFIN's API and align it to maybe formatted data.
  - `models/simple_fin_account`: Provides syncing of transactional/investment data and tracks simple_fin_item references.
  - `models/simple_fin_item`: Provides syncing info and holds onto institutional information.
  - `models/simple_fin_item/syncer`: Follows the new separated syncing similar to plaids `syncer`.
  - `models/simple_fin_rate_limit`: Lets us keep database records of how many syncs are fired today so we can easily rate limit.
  - `views/accounts/_account_error`: Displays the account error flashing icon to alert the user they need to fix their accounts. Made this generic so we could re use it.
  - `views/simple_fin/_accounts_list_content`: Allows picking what accounts you want to add for the matching accountable type. Also can pick the subtype.
  - `views/simple_fin/accounts_list`: Renders the simple_fin accounts but async so we don't block the display.
  - `views/simple_fin/new`: The basic spot to wait to load accounts by showing a spinner.
  - `config/intializers/simple_fin`: Creates a configuration class as well configures SimpleFIN on startup to see if it should be used.
- Added 3 new tables
  - `simple_fin_items`: List of instituion related info for SimpleFIN
  - `simple_fin_accounts`: Links the simple_fin_item to an account with some additional metadata
  - `simple_fin_rate_limits`: Rate limit tracking so we don't overwhelm the API endpoints.

# Images of this in action

how-to-add
select-accounts
header_error
group_error
accounts

# Where can I find this?

I've published a docker image under my account for this which can be located on dockerhub at `croudebush/maybe-simple-fin:latest`

# Conclusion

I'm still testing this personally but wanted to make this available before I disappear for a week. Let me know if you have any questions.

References: #754 #742


zachgoll:
Hey @cameronroudebush, really appreciate the changes here, but we're unfortunately not at a spot with the codebase to integrate additional providers.  I believe there may be one or two SimpleFin "companion" repositories floating around that integrate SimpleFin to Maybe, so those could be worth checking out.

We may eventually reach a state where we could implement a generic "bank provider" concept, but it will require a pretty significant effort and due to our small team size and funding requirements, it's not feasible for us right now.

cameronroudebush:
That's a shame :(

Let me know if it changes.

cameronroudebush:
@jjmata Honestly not really. After they said they didn't want to maintain this and seeing how much fluctuation was in their code base, I kind of pivoted to something that better fit my needs: [cameronroudebush/sprout](https://github.com/cameronroudebush/sprout). I thought this project was cool but lacking some of what I cared about the most.

And now with their most recent news of [stopping work](https://github.com/maybe-finance/maybe/releases/tag/v0.6.0) and them even ending support of their [finance icon service](https://synthfinance.com/), I see even less of a point of returning.

That being said, feel free to take this code and do what you need with it! If you have any questions about it I can try and provide help but know my Ruby skills are still lackluster lol.

PR #2272, opened 2025-05-21
Add new ai chat button, tweak ai navigation

ahatzz11:
I've been playing with the AI chat a bit, and some of the navigation of it all kept tripping me up so I thought I'd suggest a couple of improvements!

1. Add a 'new chat' button to the main page. A new chat being hidden behind the 3 dot menu of an existing chat felt a bit strange - this option hasn't been removed. The default chat on the dashboard load is also still the most recent chat.

  image

2. Tweak the navigation in the AI chat. Currently when you're on a new chat there's a 3 line button that puts you back to the 'main chat page', and the main chat page has a back button which takes you to your most recent. This feels very unintuitive to me - when I'm inside of something it makes sense to go 'back' to something else. If I'm on the main menu I don't expect to really go 'back' to a chat that is within that menu. If I'm alone on this, happy to swap this back to the current behavior! You can see the lack of a back button on the main page above, and here is the new back button on each chat:

  CleanShot-003141 2025-05-20 at 20 59 14@2x

3. Fix the padding around the nav title on a new chat

  CleanShot-003143 2025-05-20 at 21 03 20@2x

### DIFF
--- a/app/views/chats/_chat_nav.html.erb
+++ b/app/views/chats/_chat_nav.html.erb
@@ -8,7 +8,7 @@
       id: "chat-nav-back",
       variant: "icon",
       icon: "menu",
-      href: chats_path,
+      href: path,
       frame: chat_frame,
       text: "All chats"
     ) %>
--- a/app/views/chats/_chat_nav.html.erb
+++ b/app/views/chats/_chat_nav.html.erb
@@ -7,9 +7,10 @@
     <%= render LinkComponent.new(
       id: "chat-nav-back",
       variant: "icon",
-      icon: "arrow-left",
+      icon: "menu",
       href: chats_path,
-      frame: chat_frame
+      frame: chat_frame,
+      text: "All chats"
     ) %>
 
     <div class="grow">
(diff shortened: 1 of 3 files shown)


Shpigford:
bugbot run

ahatzz11:
bugbot run

justinfar:
In terms of:

- Fixing the padding for the nav title
- Adding a "plus" icon button (should be 36x36 container and 20x20 icon) that lets users start a new chat

All good re that

In terms of tweaking the navigation and adding a back button on the recent chat and removing it within the Chats view, I'm also fine with that. The only point I wanted to make here is that I'd retain the menu icon (instead of using the left arrow) as that's meant to signify "All chats", instead of "Go back to Chats".

image

image

Other than that good to go on these changes!

ahatzz11:
@justinfar @zachgoll all fixed! Icon is the proper size - the 'back' button is now the menu button to go back to the list. I personally still find this a bit odd, but we can keep that one!

zachgoll (approved):

PR #2276, opened 2025-05-21
UI Fixes

ahatzz11:
This PR is another round of UI fixes from @justinfar

- Use rounded-full on budget allocation bar to match balance sheet:
  
  image

- Fix backgrounds when balance sheet groups are open, add rulers between accounts/groups

  CleanShot-003162 2025-05-21 at 14 46 07@2x

  Here's another shot of before on the balance sheet when the groups were open, there is a weird slice between the corners (before I messed up the background in #2271):

  image

### DIFF
--- a/app/views/pages/dashboard/_balance_sheet.html.erb
+++ b/app/views/pages/dashboard/_balance_sheet.html.erb
@@ -58,10 +58,13 @@
             </div>
           </div>
 
-          <div class="shadow-border-xs rounded-lg font-medium text-sm bg-container min-w-fit">
-            <% classification_group.account_groups.each do |account_group| %>
-              <details class="group">
-                <summary class="cursor-pointer p-4 group-open:bg-surface bg-container rounded-lg flex items-center justify-between">
+          <div class="shadow-border-xs rounded-lg bg-container font-medium text-sm min-w-fit">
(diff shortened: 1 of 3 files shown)


Shpigford:
bugbot run

app/views/pages/dashboard/_balance_sheet.html.erb
```diff
@@ -147,3 +153,10 @@
     </div>
   <% end %>
 </div>
+
+<%# Custom style for hiding ruler when details are open %>
+<style>
+  details[open] + .group-ruler {
+    display: none;
+  }
+</style>
```
ahatzz11, line 162: If there's a better way to do this let me know! I could add a stimulus but this felt a little cleaner. This is kind of a strange issue - the border being present when the group is open leads to a border on the bottom of the new color. This isn't awful in itself, but it ends up taking up space between two open groups which leads to a white line: 

CleanShot-003164 2025-05-21 at 14 51 05@2x

zachgoll: Yeah that's a tricky one.  I like your solution a lot better than adding Stimulus to the mix.

zachgoll (approved):
Looking good!

PR #2277, opened 2025-05-22
Fix: ui fixes on mobile activity view

LEstradioto:
Ok just going forward on my latest mobile fixes (now adding more PRs, better scope to easy review)

This first one fixes below:

image

### DIFF
--- a/app/views/accounts/show/_activity.html.erb
+++ b/app/views/accounts/show/_activity.html.erb
@@ -70,8 +70,8 @@
                               data: { action: "bulk-select#togglePageSelection" } %>
             <p><%= t(".date") %></p>
           </div>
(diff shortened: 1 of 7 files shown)


ahatzz11:
Some small feedback from my perspective - this title bar is a bit weird now that there are two stacked items on the right 
CleanShot-003170 2025-05-22 at 08 45 44@2x

LEstradioto:
Thanks. I could not find this screen on Figma. But on another screens, the amount/balance column text would be hidden. 

Idk about the $ Values, stacked is understandable but not pretty

justinfar:
@LEstradioto the fixes for how you're showing the metrics works!

Re the table, I think I'd prefer if we had the table be horizontally scrollable rather than stack columns (reason being is that there's some tables that have more than 3 columns and won't stack as neatly as in your example).

PR #2278, opened 2025-05-22
Feat: Mobile Settings menu with preserve scroll + scroll on connect

LEstradioto:
Here is a quick video:

https://github.com/user-attachments/assets/a7075c75-0721-4c6b-aaba-15df8838cc3c

* `scroll_on_connect_controller` are meant to be generic, could be reused to scroll any scrollable content passing a css selector argument
* `preserve_scroll_controller` handles horizontal and vertical, meant to be re-usable too


zachgoll (approved):
Nice touches!

PR #2280, opened 2025-05-22
fix: mobile responsive category color picker

LEstradioto:
- Popup and color picker is responsive at Table/mobile
- Enables click outside to close it


zachgoll (approved):
Looks great.  The close on click is a nice addition!

PR #2321, opened 2025-05-30
fix(mappings_controller) is vulnerable to Code injection

odaysec:
https://github.com/maybe-finance/maybe/blob/ee9fe1b62db55db7616910a9c57149195386c1c0/app/controllers/import/mappings_controller.rb#L17-L17
https://github.com/maybe-finance/maybe/blob/ee9fe1b62db55db7616910a9c57149195386c1c0/app/controllers/import/mappings_controller.rb#L41-L41

Fix the issue need to validate or sanitize the `type` parameter before passing it to `constantize`. A safe approach is to maintain a whitelist of allowed class names and ensure that the user-provided `type` matches one of these names. This prevents arbitrary values from being processed.

**Steps to fix:**
1. Define a whitelist of allowed class names for `type`.
2. Validate the `type` parameter against this whitelist before calling `constantize`.
3. If the `type` parameter is invalid, handle the error gracefully (e.g., return `nil` or raise an exception).

The changes will be made in the `mapping_class` method on line 41.

#### 🎟️ Ticket #2322 
---

### DIFF
--- a/app/controllers/import/mappings_controller.rb
+++ b/app/controllers/import/mappings_controller.rb
@@ -38,6 +38,9 @@ class Import::MappingsController < ApplicationController
     end
 
     def mapping_class
-      mapping_params[:type]&.constantize
+      allowed_classes = %w[AllowedClass1 AllowedClass2 AllowedClass3]
+      type = mapping_params[:type]
+      return nil unless allowed_classes.include?(type)
+      type.constantize
     end
 end


zachgoll (changes requested):
Thanks for the contribution!  In order to get this merged, we'll need the actual classes we're using rather than the placeholder classes.

PR #2323, opened 2025-06-01
fix(ui): mfa backup codes dark mode

tonytkachenko:
This PR updates the styling of the backup MFA codes block to support dark mode.

### DIFF
--- a/app/views/mfa/backup_codes.html.erb
+++ b/app/views/mfa/backup_codes.html.erb
@@ -13,7 +13,7 @@
     <div class="space-y-6">
       <div class="grid grid-cols-2 gap-4">
         <% @backup_codes.each do |code| %>
-          <div class="p-3 bg-gray-100 rounded-lg font-mono text-lg">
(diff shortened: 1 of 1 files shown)


app/views/mfa/backup_codes.html.erb
```diff
@@ -13,7 +13,7 @@
     <div class="space-y-6">
       <div class="grid grid-cols-2 gap-4">
         <% @backup_codes.each do |code| %>
-          <div class="p-3 bg-gray-100 rounded-lg font-mono text-lg">
+          <div class="p-3 bg-gray-100 theme-dark:bg-gray-700 rounded-lg font-mono text-lg">
```
zachgoll, line 16:
```suggestion
          <div class="p-3 bg-surface-inset rounded-lg font-mono text-lg">
```

I think this was a case that we forgot to update our design tokens for, so would prefer to use `bg-surface-inset` here.

PR #2336, opened 2025-06-05
Use UUIDv7 for Primary Key

iuri-gg:
UUIDv7 has better index locality, which makes it more performant for large tables compared to UUIDv4. There is a [nice article explaining the technical parts](https://dev.to/umangsinha12/postgresql-uuid-performance-benchmarking-random-v4-and-time-based-v7-uuids-n9b). 

This PR introduces `gen_uuid_v7` function in PG and updates all PK defaults to use it. I am hoping this will help DB query performance issues since the data will have better index locality. I am not sure if existing data needs to be migrated to fully utilize the benefit of UUIDv7.


zachgoll:
Hey @iuri-gg thanks for the contribution here. Unfortunately I don't think we should merge this primarily due to the readability/complexity tradeoff.  The cost of changing `schema.rb` -> `structure.sql` is pretty high and I don't think this performance improvement warrants that cost given that we're an OSS project and we want our code to be as easy to read and as predictable (for Rails devs) as possible.

Issue #2332, opened 2025-06-03
Improvement: Copy current month's budget from previous month


vaishsriram:
Manually recreating the same budget amounts every month is a pain in the UI. Can we add a feature that lets users copy the previous month’s budget into the current one with a single click?

Nice to have:
* Auto-suggest copying on first visit to a new month.
* Keyboard shortcut?

PR #2385, opened 2025-06-16
Added budget copy feature

sathudeva7:
Fixed issue #2332


zachgoll (commented):
Thanks for the contribution here!

I've got some performance related code that will be pushed shortly related to budgeting features, so I'll have to revisit this once those are in as they will likely affect how we approach this copy/duplicate functionality.

PR #2391, opened 2025-06-18
fix(ui): iOS PWA should take full screen height

### LINKED ISSUE #2342: Bug: PWA iOS Not Fullscreen / Not Theme Aware
### General checklist

- [x] I have removed personal / sensitive data from screenshots and logs
- [x] I have searched [existing issues](https://github.com/maybe-finance/maybe/issues?q=is:issue) and [discussions](https://github.com/maybe-finance/maybe/discussions) to ensure this is not a duplicate issue
    
### How are you using Maybe?

- [ ] I am a paying Maybe customer (hosted version)
  - Paying Maybe users can also open requests in Intercom (if there is sensitive info involved)
- [x] I am a self-hosted user

### Self hoster checklist

_Paying, hosted users should delete this entire section._

If you are a self-hosted user, please complete all of the information below.  Issues with incomplete information will be marked as `Needs Info` to help our small team prioritize bug fixes.

- Self hosted app commit SHA (find in user menu): [d059465](https://github.com/maybe-finance/maybe/commit/d05946596efa3b1c925ac07fccd7d9f7aaa094ce)
  - [x] I have confirmed that my app's commit is the latest version of Maybe
- Where are you hosting?
  - [ ] Render
  - [x] Docker Compose
  - [ ] Umbrel
  - [ ] Other (please specify)

---

## Bug description

When the Maybe Finance web application is added to the Home Screen as a Progressive Web App (PWA) on iOS devices (specifically iPhone), it does not display in full-screen mode. There is persistent "dead space" or white bars at the top (around the signal/battery indicator area, or dynamic island) and at the bottom (around the home indicator bar). This prevents the PWA from fully matching the device's screen and integrating seamlessly with the iOS theme, especially in Dark Mode.

Additionally, the system UI elements (like the signal, battery, and time indicators at the top, and the home indicator bar at the bottom) do not adapt to the dark theme of the Maybe app. They remain with light/white backgrounds, breaking the immersive dark mode experience of the PWA.

### To Reproduce

Be as specific as possible so Maybe maintainers can quickly reproduce the bug you're experiencing.

1. Open Safari browser.
2. Navigate to the self-hosted Maybe app URL
3. Tap the "Share" icon (square with an arrow pointing up).
4. Select "Add to Home Screen".
5. Tap "Add" to install the PWA.
6. Close Safari and launch the Maybe app from the Home Screen.
7. Observe the white bars/dead space at the top and bottom of the screen.
8. (Optional) Ensure iOS is in Dark Mode and observe that the system UI elements (signal, time, battery) do not adapt to the app's dark theme.

### Expected behavior

What is the intended behavior that you would expect?

The PWA should launch and display in a true full-screen, immersive experience, utilizing the entire screen real estate of the iOS device. The top status bar (signal, time, battery) and the bottom home indicator bar should ideally integrate seamlessly with the app's chosen theme (e.g., blend into the dark background of the app when in Dark Mode), rather than showing distracting white bars or non-matching colors.

### Screenshots and/or recordings

We highly recommend providing additional context with screenshots and/or screen recordings.  This will _significantly_ improve the chances of the bug being addressed and fixed quickly.

Image

KenTandrian:
This PR fixes #2342, which comprises these issues:
* The page content was not taking full height of the iOS screen in PWA, resulting in dead space at the bottom.
* The top status bar and the bottom home indicator were not theme-aware, breaking the immersive dark mode experience.

## Changes
* Set `env(safe-area-inset-top)` as top margin instead of top padding for `html`.
* Remove `safe-area-inset-bottom` padding from the bottom navigation component since `html` has it.
* Ensured `body` adapts background color by dynamically setting it to `bg-surface`.
* Add top and bottom padding based on safe area height to dialog component.

## After

Dark mode
Light mode


zachgoll (changes requested):
Thanks for the fix here, but this seems to cause several visual regressions to the desktop version of the app. Below are just a few examples, but I'm sure there are several more:

CleanShot 2025-06-23 at 10 27 27

CleanShot 2025-06-23 at 10 28 08

We need to be super careful editing these files as they apply to many views across the app.

KenTandrian:
@zachgoll Thanks for the feedback. My bad for removing `h-full` from `body`, resulting in the page not taking full height.

I have reverted some of the changes and only focus on 3 things:
1. Set `env(safe-area-inset-top)` as top margin instead of top padding for `html`.
2. Remove `safe-area-inset-bottom` padding from the bottom navigation component since `html` has it.
3. Ensured `body` adapts background color by dynamically setting it to `bg-surface`.

albertorizzi:
@KenTandrian can you fix also modals? I can't close.

image

albertorizzi:
Wow! Thanks a lot!😍

albertorizzi:
Your PR https://github.com/maybe-finance/maybe/pull/2410 fix also the Current Market Price "Unknown" in these screens?

KenTandrian:
> Your PR #2410 fix also the Current Market Price "Unknown" in these screens?

Yes, I think that should be covered since [`@holding.security.current_price`](https://github.com/maybe-finance/maybe/blob/c0617f74cdadd9f96c8701158a35c5274d3f450a/app/views/holdings/show.html.erb#L24) is retrieved from the same [`find_or_fetch_price`](https://github.com/maybe-finance/maybe/blob/c0617f74cdadd9f96c8701158a35c5274d3f450a/app/models/security.rb#L15) method.

albertorizzi:
Is it possible to improve the mobile view of the Holdings tab?

image

KenTandrian:
> Is it possible to improve the mobile view of the Holdings tab?

I don't think it's related to PWA screen height? It should be a separate issue.

zachgoll (commented):
Hey @KenTandrian, sorry this is a bit of a tough one to review just given this will affect every page in our app.

I haven't had a chance to dig deeply into this yet, but my gut tells me we really need to keep that existing "global padding" intact on our `html` element, then apply specific padding/margin adjustments to things like our Dialog to make them look good on mobile.  Changing the global property to margin introduces a lot of weird UI states on "full-bleed" sections as I had screenshotted earlier, so I don't think we should be using that here.

So I think in summary what we should probably do is:

- Revert global CSS changes (keep existing padding safe areas)
- Apply specific safe area padding to fixed elements like dialogs, mobile navs, etc.

PR #2392, opened 2025-06-18
addedstart.sh script

Mpasha17:
#PR Description:
This PR introduces a new start.sh script to the repository, designed to simplify and automate the local development setup for maybe

added the start.sh which helps to start the app with just two commands

**How to use
chmod +x start.sh
./start.sh

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,112 @@
+# ===========================================================================
+# Example Docker Compose file
+# ===========================================================================
+#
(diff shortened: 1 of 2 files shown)


zachgoll:
@Mpasha17 thanks for sharing this script! 

Unfortunately we cannot accept this to the codebase as we try to keep things as "Vanilla Rails" as possible, but I think this could definitely help some devs speed up their workflow and is a good reference!

PR #2399, opened 2025-06-20
Fix: Use PATCH method for onboarding preferences form

### LINKED ISSUE #2398: Bug: [Onboarding form submits with POST, expects PATCH]
### General checklist

- [x] I have removed personal / sensitive data from screenshots and logs
- [x] I have searched [[existing issues](https://github.com/maybe-finance/maybe/issues?q=is:issue)](https://github.com/maybe-finance/maybe/issues?q=is:issue) and [[discussions](https://github.com/maybe-finance/maybe/discussions)](https://github.com/maybe-finance/maybe/discussions) to ensure this is not a duplicate issue

### How are you using Maybe?

- [x ] I am a paying Maybe customer (hosted version)
- [x] I am a self-hosted user (This bug appeared in the self host dev environment)

### Self hoster checklist

-   Self hosted app commit SHA (find in user menu): `Running latest main branch commit from a fresh dev container setup`
-   [x] I have confirmed that my app's commit is the latest version of Maybe
-   Where are you hosting?
    - [ ] Render
    - [x] Docker Compose
    - [ ] Umbrel
    - [ ] Other (please specify)

---

## Bug description

Submitting the user preferences form during the new user onboarding flow results in an `ActionController::RoutingError`. The error occurs because the form incorrectly submits with the `POST` HTTP method to a `/users/:id` path, which is only configured to accept `PATCH` or `PUT` requests for updates.

### To Reproduce

1.  Set up a fresh local development environment using the provided Docker Compose configuration.
2.  Navigate to `http://localhost:3000` and create a new user account.
3.  Proceed to the second step of the onboarding flow, "2. Preferences", at the `/onboarding/preferences` path.
4.  Leave the default preferences or make changes.
5.  Click the "Complete" button.
6.  See the `Routing Error` page with the message: `No route matches [POST] "/users/31b7dee2-7067-4ec9-a8bf-50a0617c07fd"`.

### Expected behavior

After clicking the "Complete" button on the preferences form, the form should submit successfully using the `PATCH` HTTP method. The user's preferences should be saved, and they should be redirected to the next step in the onboarding flow ("3. Goals").

### Screenshots and/or recordings

1.  **The "Configure your preferences" page** just before clicking "Complete" (`

Image

`).
2.  **The Rails `Routing Error` page

Image

** that appears immediately after (`image_201586.jpg`).

---

Contributor's Note

Hi team,

Just wanted to add some extra context. I encountered this issue while setting up a fresh local development environment for the first time. If this behavior is intended or I've missed a specific setup step, my apologies!

I did some initial debugging and confirmed that:

The config/routes.rb file correctly expects a PATCH request for this update action.
The OnboardingsController also appears correct, as it properly loads the existing Current.user.
Interestingly, this seems to be a state-dependent bug. After applying a local fix that allowed the form to submit successfully once, the original code also began to work on subsequent reloads for my user. This suggests the bug only manifests for users in a specific initial onboarding state before their preferences have been saved for the first time.

I have a simple, one-line fix ready on a local branch that explicitly sets method: :patch in the preferences.html.erb view, which resolves the issue for new users.

If you can confirm this is a valid bug, I would be happy to open a pull request right away.

alvaro-rrdt:
The form on the user onboarding preferences page was submitting with the POST HTTP method, which caused an ActionController::RoutingError because the /users/:id route only accepts PATCH/PUT for updates.

This appeared to be a state-related bug where form_with was not correctly inferring the method from the persisted @user object during the initial onboarding flow.

This change explicitly sets `method: :patch` on the form_with helper to ensure the correct verb is always used, resolving the routing error for all new users.

Fixes #2398

### DIFF
--- a/app/views/onboardings/preferences.html.erb
+++ b/app/views/onboardings/preferences.html.erb
@@ -63,7 +63,7 @@
 
     <p class="text-secondary text-xs mb-4"><%= t(".preview") %></p>
 
-    <%= styled_form_with model: @user, data: { turbo: false } do |form| %>
+    <%= styled_form_with model: @user, method: :patch, data: { turbo: false } do |form| %>
       <%= form.hidden_field :set_onboarding_preferences_at, value: Time.current %>
       <%= form.hidden_field :redirect_to, value: "goals" %>
 


app/views/onboardings/preferences.html.erb
```diff
@@ -63,7 +63,7 @@
 
     <p class="text-secondary text-xs mb-4"><%= t(".preview") %></p>
 
-    <%= styled_form_with model: @user, data: { turbo: false } do |form| %>
+    <%= styled_form_with model: @user, method: :patch, data: { turbo: false } do |form| %>
```
zachgoll, line 66: If I'm not mistaken, Rails should already be submitting a PUT request since the `@user` exists already and is not a new record. I'm not actually able to reproduce the original error either, so I'm wondering if there is an unrelated bug going on instead?

zachgoll: Gotcha, I'm going to close out this PR and the corresponding issue for now as I don't think we've got a bug, but happy to reopen in the future if we can reproduce it more reliably!

PR #2404, opened 2025-06-23
perf(sync): Preload and cache exchange rate data when syncing.

Joelute:
When we are performing transaction syncs, we are looking up for the exchange rate of the transaction if the currency of the transaction is different from the account currency. This look up is performed on every transactions that is being synced, which results in a `N+1` query.

This change loads and caches the exchange rates when performing syncs.

### DIFF
--- a/app/models/balance/sync_cache.rb
+++ b/app/models/balance/sync_cache.rb
@@ -15,16 +15,33 @@ class Balance::SyncCache
     converted_entries.select { |e| e.date == date && (e.transaction? || e.trade?) }
   end
 
+  def find_rate_by_cache(amount_money, to_currency, date: Date.current, fallback_rate: 1)
+    raise TypeError unless amount_money.respond_to?(:amount) && amount_money.respond_to?(:currency)
+
+    iso_code = Money::Currency.new(amount_money.currency).iso_code
+    other_iso_code = Money::Currency.new(to_currency).iso_code
+
+    return amount_money if iso_code == other_iso_code
+
+    exchange_rate = exchange_rates(to_currency)[[ iso_code, date ]]&.last&.rate ||
+      ExchangeRate.fetch_rate(from: iso_code, to: other_iso_code, date: date)&.rate ||
+      fallback_rate
+
+    raise Money::ConversionError.new(from_currency: iso_code, to_currency: other_iso_code, date: date) unless exchange_rate
+
+    Money.new(amount_money.amount * exchange_rate, other_iso_code)
+  end
+
   private
     attr_reader :account
 
     def converted_entries
-      @converted_entries ||= account.entries.order(:date).to_a.map do |e|
+      @converted_entries ||= entries.map do |e|
         converted_entry = e.dup
-        converted_entry.amount = converted_entry.amount_money.exchange_to(
+        converted_entry.amount = find_rate_by_cache(
+          converted_entry.amount_money,
           account.currency,
           date: e.date,
-          fallback_rate: 1
         ).amount
         converted_entry.currency = account.currency
         converted_entry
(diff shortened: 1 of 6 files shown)


zachgoll (changes requested):
I agree with your diagnosis here and like the overall direction you're taking this with the `exchange_rates` method in the `sync_cache`.

That said, I don't think `cached_exchange_to` belongs in the `Money` class. That class should not be aware of caching, and if we need these bulk loaded rates, I'd rather just encapsulate the "find rate from cache" logic directly in our `sync_cache.rb` class.

Also, I'm thinking we should remove `QueryAssertions` as this adds complexity and makes our tests brittle by asserting things that aren't directly relevant to the subject under test. Rails already ships with [a helper for this](https://api.rubyonrails.org/v7.2.1/classes/ActiveRecord/Assertions/QueryAssertions.html#method-i-assert_queries_count), but I still think we ought to be very sparse with our usage of it and let our APM tools notify us of these performance issues.

Joelute:
Hi Zach, I've updated the PR to:
- Rename `cached_exchange_to` to `find_rate_by_cache`
- Moved the `find_rate_by_cache` logic over to `sync_cache.rb`
- Removed my custom test case helper in favour for the built in helper that Rails has

Please do take a look and give it another review!

app/models/balance/sync_cache.rb
```diff
@@ -19,28 +19,65 @@ def get_entries(date)
...
+      return entry if iso_code == other_iso_code
+
+      exchange_rate = exchange_rates[[ iso_code, date ]]&.last&.rate ||
+        ExchangeRate.find_or_fetch_rate(from: iso_code, to: other_iso_code, date: date)&.rate ||
+        fallback_rate
+
+      raise Money::ConversionError.new(from_currency: iso_code, to_currency: other_iso_code, date: date) unless exchange_rate
+
+      Money.new(entry.amount * exchange_rate, other_iso_code)
+    end
```
zachgoll, line 62: I think this method needs another look. A few issues I'm seeing at first glance:

- Errors not handled
- Inconsistent return values
- Over-fetching rates (we don't need to `find_or_fetch_rate` if we're preloading to a cache already)

test/models/balance/forward_calculator_test.rb
```diff
@@ -35,7 +35,10 @@ class Balance::ForwardCalculatorTest < ActiveSupport::TestCase
     assert_equal 0, @account.balances.count
 
     expected = [ 0, 0 ]
-    calculated = Balance::ForwardCalculator.new(@account).calculate
+    calculated = nil
+    assert_queries_count(3) do
+      calculated = Balance::ForwardCalculator.new(@account).calculate
+    end
```
zachgoll, line 41: We probably should not be using this `assert_queries_count` at such a broad level. This test is brittle because if we make any change to `ForwardCalculator` (highly possible in the future), the number of queries may change and we'd have to come back and update all these tests.

Joelute: I've just pushed a couple testcases aimed at testing `find_rates_by_cache` and performing more query assertions there instead.

app/models/balance/sync_cache.rb
Joelute: I've updated the param names from:
- `entry` -> `money`
- `other_currency` -> `to_currency`
Hopefully this reduces the confusions around the names and makes the code more readable.

In the event that we don't have the exchange rates in our cache, we want to be able to fetch for that exchange rate and load it for use. Hence, `fetch_rate`

Joelute:
Hi @zachgoll, I've just pushed some code to provide a clearer naming, better error handling, and new test cases for all the changes. Please take a look.

jjmata:
This repo is [no longer maintained](https://github.com/maybe-finance/maybe/blob/main/README.md?plain=1#L15).  Want to submit your PR [here](https://github.com/we-promise/sure) instead, @Joelute? 🙏

PR #2419, opened 2025-06-27
Create plaid.md

Facinorous-420:
Added instructions on how to set up plaid with your self hosted instance

### DIFF
--- a/docs/hosting/plaid.md
+++ b/docs/hosting/plaid.md
@@ -1,0 +1,86 @@
+# Setting Up Plaid
+
+## 1. Create Your Plaid Account
+
+Go to [https://dashboard.plaid.com](https://dashboard.plaid.com) and register for a new account.
+
+---
+
+## 2. Requesting Access Based on Your Bank Type
+
+### For Banks That **Do Not Require OAuth**
+
+1. On the Home page, find the section labeled "Learn how to build with Plaid" and click **Unlock real data**.
+2. Enter your real name and phone number.
+3. In the description box, write a statement such as:  
+   This is for personal use only on a self-hosted version of the Maybe Finance software. I am only using it to manage my finances, sync my bank accounts, track my spending, and create a budget.
+4. Leave the "Additional products" section **unchecked**.
+5. Click **Request Access**.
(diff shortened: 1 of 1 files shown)


Facinorous-420:
> It's probably important to also note that in order for Plaid webhooks to work, your selfhosted server needs to be accessible from external networks. I don't think you need to set anything up for webhooks in the Plaid dashboard.

You are correct, your Maybe instance needs a FQD with SSL. Plaid does not work properly without this. I first tried a local domain with local SSL generated via mkcert, but of course Plaid doesn't have the DNS to see my custom domains so that didn't work. 

It would be nice if we didn't have to expose our instances to the web, such as Actual is about to using Simplefin.

I can add a line regarding the need for external access, FQD, and SSL. :)

zachgoll:
Love it! Thanks for the guide, I think a lot of self hosting folks will find this useful.

We've mentioned several times across our channels that our internal team does not have the capacity to officially support self-hosted Plaid, so we can't accept this into the main codebase. That said, we'll move it over to Community discussions and anyone who's looking to learn how to configure Plaid can reference it there.

Facinorous-420:
> @Facinorous-420 I just remembered that Github does not have a way to convert a Pull Request to a discussion, so I'm going to close this one out. I'd recommend copy/pasting this guide to a discussion.

Thanks. Created it here:
https://github.com/maybe-finance/maybe/discussions/2421

PrplHaz4:
> @Facinorous-420 I just remembered that Github does not have a way to convert a Pull Request to a discussion, so I'm going to close this one out. I'd recommend copy/pasting this guide to a discussion.

Thanks so much Zach for embracing the open and self-hosted-ness while drawing a clear line and providing a place where the community can and should continue support.

jjmata:
Hey @Facinorous-420 ... would you be interested in sending this to a new [community fork](https://github.com/we-promise/sure) we're trying to get going?  Want to re-submit this PR there?

I am sure we could do what you mention above now that we're not worried about the "commercial support" angle:
> I must say, I was looking for a guide like this specifically in the docs of the repo because there was that docs folder.
It would be nice to have all documentation organized in 1 spot, rather than having it all over the place.

